PR #313, opened 2016-09-03
Finetunings in repo_head UI and html_helper

sbcgua:
Better placement of repo_head elements: package displayed most right both for on/offline, url shown just after repo name and without text-overflow cut
html_helper not indents style content in curly brackets


sbcgua:
image

sbcgua:
abaplint seems to dislike `IGNORING CASE` :)

larshp:
yeah, it is full of bugs :) will fix

https://larshp.github.io/abaplint/viz/find.svg

larshp:
thanks :+1:

PR #351, opened 2016-09-20
UI: Main page, hide file list, adjust diff #347

sbcgua:
to #347 

P.S. one commit, please don't squash :)

### DIFF
--- a/src/zabapgit_definitions.prog.abap
+++ b/src/zabapgit_definitions.prog.abap
@@ -53,7 +53,7 @@ TYPES: BEGIN OF ty_repo_file,
          path        TYPE string,
          filename    TYPE string,
          is_changed  TYPE abap_bool,
-         remote_only TYPE abap_bool,
+         new         TYPE char1,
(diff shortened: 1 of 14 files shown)


larshp:
The dropdown triangle looks a bit strange, but it works,
image

sbcgua:
Mmmmm, which version of GUI do you have ? Is it 7.3 ? It might force older IE simulation :((( Which does not support css transform. I'll have a look ... but I'm not sure what to do here ... maybe use square instead of triangle ? (no transform required). By the way, does menu dropdown work fine ? I commented IE7 workaround in css - they might overlap the menu item instead of being below them - is it the case for you ?

P.S. Would you mind if I add some feature to show some tech info about browser in some of next commits ?  It will be either command to popup, or some faint text somewhere at the bottom.

larshp:
yes it is SAP GUI 730 SP 6. Though I normally use 740

yes, the menu dropdown works

browser information: good idea, it will help troubleshooting in the future

PR #419, opened 2016-11-12
Independent diffs, local state per file, refactorings

sbcgua:
To #407 

Sorry for huge PR. The issue was not superficial at all. Had to refactor much internals. Here is come comment for history and as a help for review.

- changelog.txt added
- INCLUDE zabapgit_migrations added - called from run form, contains migration methods and call, probably a good place for future migrations
- Definitions: ty_file_signature added as a shorter vehicle for hash data, gc_state - list of file states
- zabapgit_file_status - completely rewritten to support individual state calculation and performance. Hopefully has better structure now. All files in the result has now remote and local states, path and filename (needed for whole diff in particular).
- zabapgit_git.prog - push returns the list of actually pushed files
- zabapgit_objects.prog - lcl_objects_files collects accessed files which after is used to update only those checksums which were actually pulled
- page_main and other UI classes - styles and code for independent diff display
- zabapgit_repo.prog - build_local_checksums removed, instead rebuild_local_checksums for complete guess-rebuild, and update_local_checksums to update specific filelist (actually pulled/pushed). Update does not recalc sha1 - it relyes on proper param input. 
- service_repo - can refresh_local_checksums
- zabapgit_tadir.prog - got helper method to find object package (without all tadir struc)
- unit test for file_state


larshp:
cool, will take a look tomorrow or on monday :smile:

sbcgua:
- added global `View diff` to the main page, change Pull/Stage showing logic. I think all this dtuff above does indirectly close #391 by the way
- Diff page: style tweaks, paddings, display change-side marker, invert green/red logic for files changed remotely (still to be improved, but later)

P.S. Ufff, tired from abap for some time :) Next will probably have a look at repo favourites thing after a while.

larshp:
Hi,
I get the warning each time I start abapGit, and I click OK each time,
image

looks like all files are skipped in `rebuild_local_checksums`, because the path does not match
```abap
      READ TABLE lt_remote ASSIGNING <ls_remote>
        WITH KEY path = <ls_local>-file-path filename = <ls_local>-file-filename
        BINARY SEARCH.
      CHECK sy-subrc = 0.  " Ignore new ones
```

Also I get the following errors:
image

sbcgua:
Hmmm, very strange. I cannot repeat both of errors ...
Also the "Files for object" error is from the part of the code I didn't touch

Could you send me a screen of `it_results` at beginning of `lcl_sap_package=>check` ?

> Suggest taking the SHA1 from lt_local in REBUILD_LOCAL_CHECKSUMS and store it as the checksum? Guess also the correct path is required for it to work later?

Checksums: not really. It would build checksums based on actual curent state of local files whereas they might be changed from last-sync-state. The whole idea of this code is to have 3 states: actual local, actual remote, and last-sync to compare them versus.

path: that's a confusing part - both local and remote paths should be correct by the call of status ... but I cannot recreate your issue. Probably also `it_local` and `it_remote` screens from `lcl_file_status->calculate_status` could be helpful

P.S. might be that `lcl_sap_package=>check` should be also adjusted to the new status output. I'll have a look, **but "Files for object" error logic looks OK now and should work without change** so the issue is somewhere else.

larshp:
code in method rebuild_local_checksums, must be changed from
```abap
    lt_local        = get_files_local( ).
    lt_remote       = get_files_remote( ).
```
to
```abap
    lt_remote       = get_files_remote( ).
    lt_local        = get_files_local( ).
```
to make sure it can find the .abapgit.xml from remote in get_local_files

larshp:
on another note, we have diff in

- the file/object list

- the staging page

- and now in the main menu

it seems that there are too many places the diff page can be called from?

sbcgua:
local/remote -> done. Yeap, I had this issue also in status. Probably can be refactored a bit ...

diff link: well, yes. Why not? seems to be convenient. Personally I missed the whole diff overview from main. Maybe it can show up a bit less ... I'll think about design a bit more ... but in general I think it look ok.

larshp:
Has something changed in the diff page? The counter shows 2 added(green) lines, but the diff shows 2 red lines. Perhaps it also worked like this earlier?

image

sbcgua:
highlighting: yes, it was changed in https://github.com/larshp/abapGit/pull/419/commits/430b6e1f9bba50974a0ec0b89d1cd1a119454a9d 
The idea of change: now we track not only the fact that files differ but where the difference come from (local or remote). Consequently if it comes from remote - the highlighting logic is reversed.
It probably need some tweaks. My guess is that side of string should be also changed to reflect it correctly. But I would not do it in this PR, it may take some time. So ... how you prefer ... I can either comment new logic for now or just fix it some later separately.

P.S. those errors in your first response - did they go away ?

larshp:
okay, we'll fix the coloring of the diff in next PR, I'll open a separate issue for it

squashing and merging

thanks :smile:

sbcgua:
Seems like there are some issues ... :(
I'll have a look at the evening

larshp:
I've fixed the .abapgit.xml issue, I dont think the other ones are part of the normal flow, they are power user features, which are nice to have. :smile:

PR #451, opened 2016-11-22
favbar

sbcgua:
To #401 

image

### DIFF
--- a/src/zabapgit_html.prog.abap
+++ b/src/zabapgit_html.prog.abap
@@ -204,6 +204,7 @@ CLASS lcl_html_toolbar DEFINITION FINAL.
           iv_txt TYPE string
           io_sub TYPE REF TO lcl_html_toolbar OPTIONAL
           iv_act TYPE string    OPTIONAL
+          iv_ico TYPE string    OPTIONAL
           iv_opt TYPE c         OPTIONAL
           iv_typ TYPE c         DEFAULT gc_action_type-sapevent,
       count
@@ -215,6 +216,7 @@ CLASS lcl_html_toolbar DEFINITION FINAL.
           iv_vertical               TYPE abap_bool OPTIONAL
           iv_sort                   TYPE abap_bool OPTIONAL
           iv_as_angle               TYPE abap_bool OPTIONAL
+          iv_with_icons             TYPE abap_bool OPTIONAL
         RETURNING
           VALUE(ro_html)            TYPE REF TO lcl_html_helper.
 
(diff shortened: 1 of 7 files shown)


sbcgua:
Ah no, hold a bit with this PR pls, I forgot a "new/clones repo to favs" feature

src/zabapgit_html.prog.abap
```diff
@@ -250,15 +253,17 @@ CLASS lcl_html_toolbar IMPLEMENTATION.
...
     ls_item-opt = iv_opt.
     ls_item-typ = iv_typ.
     APPEND ls_item TO mt_items.
   ENDMETHOD.  "add
 
-  METHOD render.
+  METHOD render. "TODO refactor
 
     DATA: lv_class   TYPE string,
+          lv_text    TYPE string,
```
EduardoCopat: Varaible not used?

sbcgua:
added toggle_favorite after clone & newoffline.
removed lv_text, indeed forgot it, thanks. favbar is fine - this is naming for all menus more or less. 

@larshp there is a bug with locks but I don't know locks well so not sure how to fix. 
Actually toggle_fav after clone or new does not work. It raises error `Could not aquire lock`. It does not happen in debugger, seems like the code lock-release-lock too quickly and 'release' does not complete. Could you have a look ?

e.g. `ZABAPGIT_SERVICES_REPO->new_offline`
```
    lcl_app=>user( )->set_repo_show( lo_repo->get_key( ) ). " <<< LOCK 1
    toggle_favorite( lo_repo->get_key( ) ).                  " <<< LOCK 2
```

larshp:
locking: try adding COMMIT WORK AND WAIT between the 2 statements (I think we have 2 open issues on looking into locking)

sbcgua:
didn't work. Can it be because `set_repo_show` issues `COMMIT WORK` without `WAIT` just before ?

larshp:
yes, adding "AND WAIT" in set_repo_show will fix the issue

larshp:
can we extend the icon element into the red area? I typically move my mouse in a downward left direction with dropdowns, and when I move the mouse into the red area, the dropdown disappears. It looks like there is enough space to extend the element,

image

sbcgua:
moving `wait` to `set_repo_show ` fixed the issue, thanks (not commited yet).

icon: so the question is about mouse movement, not about icon size, right ? I think I'll fix it another way, design will remain the same. It might not work for IE7 ... but I guess it is OK. Have a look in the evening.

larshp:
when the WAIT/locking issue is fixed I guess we can merge? The mouse movement is a minor issue which we can always fix in a follow up commit

sbcgua:
OK, agree. Then can be merged. I amended the previous commit so it includes `wait`.

PR #467, opened 2016-11-27
External files for css and js

sbcgua:
to #374 Main part of code
See also (but not a prerequisite for the moment): https://github.com/larshp/abapmerge/pull/16

- css/js removed from code, created as w3mi objects
- new `lcl_asset_manager` class created to extract the w3mi or fall back to inline resource which should be assembled by abapmerge (checked - works)
- `lcl_path` got back that path remover method (used now) + a small fix to w3mi serialization to hide original path to the file (for the privacy reasons)
- refactor of JS code - moved stage logic to a "class"

Seems this is it. There is some more work on css refactoring but I'll do it later - now the apprach works already so should be tested a bit and then finetuned.


sbcgua:
P.S. assembled abapgit is 1.1Mb - wow :)

sbcgua:
The last commit also centralizes images to `asset_manager`. So pages now are clean code.

larshp:
time to write a ABAP minifier? :stuck_out_tongue_closed_eyes:

larshp:
I get the following diff after pulling the code. Can we CONDENSE the filesize value, and perhaps CLEAR the version, so it is not in the XML?

image

sbcgua:
Yeap, good idea.

sbcgua:
Ah, forgot the version. Moment

sbcgua:
SHould be ok now

larshp:
lets do it :+1:

larshp:
compiled version does not work:

image

larshp:
fixed, updated package.json to use latest abapmerge version

PR #504, opened 2016-12-15
Removed invalid mix between CHANGING and RETURNING

nununo:
+ parameter names renamed to conform to the project's style
+ now the lines are inserted in the table inside the auxiliary methods (since, without RETURNING, we can no longer call the methods directly from the INSERT command.


nununo:
Cool. I just upgraded it in my 702 system and it works flawlessly ;)

PR #534, opened 2017-01-12
HEAD in select branch dialog & highlighting css bug fix 

sbcgua:
to #531 + a small bug fix after previous `page_diff` commits.

2017-01-12_16h51_13

Removed `refs/heads` prefix. Seems like look better. But will need changes if we will support switching to tags once. Also in general `branch_list_popup` is a bit clunky now :( Not sure how to improve it immediately. But it works :)

plus added normalize_branch_name call in new_branch_popup to substitute space to '-' (space are not supported in branch names if I understand well, guthub also does similarly)

### DIFF
--- a/src/zabapgit_css_common.w3mi.data.css
+++ b/src/zabapgit_css_common.w3mi.data.css
@@ -519,19 +519,17 @@ table.diff_tab code {
   white-space: pre;
 }
 table.diff_tab td.code {
-  /* font-family: inherit; */
-  /* white-space: pre; */
   word-wrap: break-word;
   white-space: pre-wrap;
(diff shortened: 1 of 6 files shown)


sbcgua:
added also some small fixes for html intentation - page_main were not indeting well due to tricky div structure in menu.
Adjusten menu_end logic (hiding the separator at the last item in menu), nwo it is done by CSS

sbcgua:
BTW, created a conveient tool for CSS adjusting - https://github.com/sbcgua/abap_w3mi_poller
Saves time on CSS upload in `SMW0` ;)

PR #575, opened 2017-01-23
Correct commit message formatting

filak-sap:
Title message and full commit message must be delimited with an empty line,
otherwise CLI tools take full commit message as a part of title message.

Signed-off-by: Jakub Filak <jakub.filak@sap.com>

### DIFF
--- a/src/zabapgit_services_git.prog.abap
+++ b/src/zabapgit_services_git.prog.abap
@@ -201,7 +201,7 @@ CLASS lcl_services_git IMPLEMENTATION.
     ls_comment-comment  = is_commit-comment.
 
     IF NOT is_commit-body IS INITIAL.
-      CONCATENATE ls_comment-comment is_commit-body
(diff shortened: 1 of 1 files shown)


larshp:
Okay, do you have an example where it does not work?

filak-sap:
Here you are:

```
$ git show --pretty=short 1c4769be135ebe92df26c9c6dabe9586f75848d8
commit 1c4769be135ebe92df26c9c6dabe9586f75848d8
Author: XXX <jakub.filak@sap.com>

    Short message
    Body

$ git log --oneline
1c4769b Short message Body
```

However, the expected output is the following:

```
$ git log --oneline
1c4769b Short message 
```

I'll write more verbose commit messages next time ;)

larshp:
did some testing, will need to look through some more code to check that this fix will not cause problems somewhere else

image

PR #580, opened 2017-01-25
Removed original language validation

christian1020-94:
According to #545. Please validate this change doesn't affect anything else.

### DIFF
--- a/src/zabapgit_repo_impl.prog.abap
+++ b/src/zabapgit_repo_impl.prog.abap
@@ -473,9 +473,6 @@ CLASS lcl_repo IMPLEMENTATION.
     IF mo_dot_abapgit IS INITIAL.
       mo_dot_abapgit = lcl_dot_abapgit=>build_default( ms_data-master_language ).
     ENDIF.
-    IF mo_dot_abapgit->get_master_language( ) <> sy-langu.
-      lcx_exception=>raise( 'Current login language does not match master language' ).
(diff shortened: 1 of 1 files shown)


larshp:
the check mentioned in #545 can be removed, but not the general check.

Eg. in the example from #545 only the master language text is saved in the xml, so we want to make sure this text is installed in the target system using the same language

christian1020-94:
Mmm then I have to log on in Englisb in order to update my abapGit version? Or is there something else that can be done?

larshp:
perhaps implement a feature allowing star in MASTER_LANGUAGE, `<MASTER_LANGUAGE>*</MASTER_LANGUAGE>`

However I am not sure if you install abapGit with login language = Spanish, and then modify it when logged in in English and submit a PR with english? Perhaps this is why the description was lost in PR https://github.com/larshp/abapGit/pull/581

image

christian1020-94:
Yes,  I think that is exactly what happened. Should i uninstall abapgit for a clean setup?

larshp:
would be nice, or alternatively just dont commit/stage the xml file

christian1020-94:
Right,  but this should be fixed in your master?

larshp:
fixed in https://github.com/larshp/abapGit/commit/a0ad8dcec9467ba630bdcc5110dd4a713e9ad11f

larshp:
closing

PR #584, opened 2017-01-27
Fix handling of extenstion objects

pfaller:
I just cloned a repository which contained also a transportable Code Inspector variant. These variants are some kind of extension objects, which are stored as 'program' (object type _PROG_ in _TADIR_), but which are named _...=====VC_ (where the 'VC' is at position 31 and 32 of the program name), e.g. 
image
image

This caused an error ("error from INSERT REPORT") since _INSERT REPORT_ by default can not handle such extensions, e.g. programs with names longer than 30 characters. However, it works using the internal addition _EXTENSION TYPE_, see http://help.sap.com/abapdocu_751/en/abapinsert_report_internal.htm#!ABAP_ADDITION_1@1@. (There, type group SREXT is referenced for allowed extension types, but VC is not in it. Seems that the suffix of the program name corresponds to the required value for _exttype_ of the addition _EXTENSION TYPE_.

This pull request contains a fix for this issue, thus I was able to clone the repository, also the code inspector variant was successfully imported.

### DIFF
--- a/src/zabapgit_objects.prog.abap
+++ b/src/zabapgit_objects.prog.abap
@@ -1104,12 +1104,28 @@ CLASS lcl_objects_program IMPLEMENTATION.
     ELSE.
 * function module RPY_PROGRAM_INSERT cannot handle function group includes
 
-      INSERT REPORT is_progdir-name
-        FROM it_source
-        STATE 'I'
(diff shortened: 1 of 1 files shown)


mbtools:
FYI, not really a program but some data... Here's an example

image

PR #588, opened 2017-01-30
Separate CLAS and INTF serialization

EduardoCopat:
Hi,

Sorry for the long PR, but serialization had many dependencies which had to be isolated in order to safely refactor.

Still open on Issue #21:

- delete
- remove INTF inheritance from CLAS


EduardoCopat:
I tested locally and also with the abapOpenChecks repo.

christian1020-94:
I'm getting an error in ABAP 702:

> Method SERIALIZE_ABAP may not have a RETURNING parameter and an EXPORTING or CHANGING parameter at the same time.

Not looked into the details, but I guess this is not supported in 702.

EduardoCopat:
@christian1020-94 Looking into it

PR #608, opened 2017-02-09
Prog serializing texts

TheWirtschaftsmann:
Hello @larshp , 

During last couple of days I was working on serializing / deserializing of program texts maintained as translations. With a lot of support from @sbcgua and @sshlapak, I've finally arrived at a version that is working:)
 
Please note, there was one problem involved with deserializing: translations were correctly retrieved from XML and populated as translations, however we couldn't activate them. After a lot of experiments, @sshlapak suggested to remove parameter state = 'I' in method deserialize_textpool (screenshot below). Currently if works fine - both for objects PROG and FUGR, which use this method. Do you possibly know if that might involve possible issue here?

2017-02-09_12h51_46

### DIFF
--- a/src/zabapgit_object_prog.prog.abap
+++ b/src/zabapgit_object_prog.prog.abap
@@ -15,7 +15,7 @@ CLASS lcl_object_prog DEFINITION INHERITING FROM lcl_objects_program FINAL.
 
   PRIVATE SECTION.
     TYPES:  BEGIN OF ty_tpool_i18n,
-              langu    TYPE langu,
+              language TYPE langu,
               textpool TYPE textpool_table,
             END OF ty_tpool_i18n,
             tt_tpool_i18n TYPE STANDARD TABLE OF ty_tpool_i18n.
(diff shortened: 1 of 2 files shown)


src/zabapgit_object_prog.prog.abap
```diff
@@ -13,6 +13,21 @@ CLASS lcl_object_prog DEFINITION INHERITING FROM lcl_objects_program FINAL.
...
+            END OF ty_tpool_i18n,
+            tt_tpool_i18n TYPE STANDARD TABLE OF ty_tpool_i18n.
+
+    METHODS:
+    serialize_texts
+      IMPORTING io_xml TYPE REF TO lcl_xml_output
+      RAISING lcx_exception,
+    deserialize_texts
+      IMPORTING io_xml TYPE REF TO lcl_xml_input
+      RAISING lcx_exception.
```
EduardoCopat: Suggestion: What if `serialize_texts` is renamed to serialize_i18n_texts ? It makes more explicit that this is serializing of translated texts. Same applies for `deserialize`

TheWirtschaftsmann: Hi @EduardoCopat,  use of method names with "i18n" is also an option. But then again existing names are also descriptive and quite precise. Besides, I've followed existing approach to naming convensions for (de)serializing for objects DOMA, DTEL.

P.S. One more reason is that not everyone might know what i18n means;) At least I didn't know about it a week ago.

Issue #656, opened 2017-03-10
UI: lcl_toolbar refactoring


sbcgua:
connected to #655 (as prepreq) and ready.

Mainly inspired by this: https://codepen.io/philhoyt/pen/ujHzd
So implemented, html and CSS is cleaner now + a couple of new features/sideeffects available

1) Dropdowns look like this now

![image](https://cloud.githubusercontent.com/assets/15635498/23812293/96d630a0-05e2-11e7-855a-f37bd6d3a33e.png)

2) Theoreticaly they support 3 levels menu. I didn't try. Might be usefull for future.

3) Items can have separator type (e.g. Filter menu in Diff + may be useful in future).

![image](https://cloud.githubusercontent.com/assets/15635498/23812333/ccf97d36-05e2-11e7-96cb-3e8425c4a2c3.png)

4) Lists may be "check box" lists. In fact just list with predefined icons but with cleaner API. E.g.

```
      ro_toolbar->add(  " Show/Hide files
        iv_txt = 'Show files'
        iv_chk = boolc( NOT mv_hide_files = abap_true )
        iv_act = c_actions-toggle_hide_files ).
```

Check box list can work with sapevents (grid options at main page) or with JS (filter at diff page). For the latter case there is a wrapper in common.js - `CheckListWrapper` - which simplifies hooks and action handling.

5) Repo grid menu move to repo toolbar (indeed looks cleaner)

![image](https://cloud.githubusercontent.com/assets/15635498/23812489/6c39aba0-05e3-11e7-9f1e-29610ee355fb.png)

6) minor tweaks around (save at DB edit page move upwards, css tweaks, etc, etc )

sbcgua:
Let it hang here for a while. Unless there are bug and wishes I'll close in a couple of days.

EduardoCopat:
This new UI style looks great 🥇 .

sbcgua:
@EduardoCopat Thanks :)

sbcgua:
I guess can be closed

Issue #655, opened 2017-03-10
UI: page_diff new features (author, filter, stage)


sbcgua:
A couple of new friday features for diff page ready

1) Each diff shows author of change (if available)

![image](https://cloud.githubusercontent.com/assets/15635498/23811616/63a4d8d8-05df-11e7-9ad7-4e431a1a79fa.png)

2) Toolbar has Filter command. Files in diff can filtered by type (abap, xml, other) and by author (useful for teams). Active Filter is highlighted. Set of filters detected automatically

![image](https://cloud.githubusercontent.com/assets/15635498/23811688/b2343ab6-05df-11e7-9d7c-d725d2437bc8.png)

3) "other" files also studied if they are completely binary and then not displayed as diff. The detection is rather straight forward - check if first 100 bytes are out of [typical text range](https://stackoverflow.com/questions/277521/how-to-identify-the-file-content-as-ascii-or-binary) . But works ok for the moment. This is needed mainly for W3MI binary content (which I have ;) )

![2017-03-10_22h22_26](https://cloud.githubusercontent.com/assets/15635498/23811866/79065a52-05e0-11e7-88f4-9a70675c8736.png)

4) Last but not least ! Stage command at diff page ! Why: imho main->stage->diff->stage->commit flow is unnatural. I always check diff before staging. So main->diff->stage->commit. Less action = good. Also! If diff is filtered (or it is diff for a specific file) and you switch to stage - then the filtered files are already marked as "A" (nice to commit one object only OR for the mentioned team work - filter and commit files of certain person)

![image](https://cloud.githubusercontent.com/assets/15635498/23811977/09d48356-05e1-11e7-9987-e97c36afd68b.png)

![image](https://cloud.githubusercontent.com/assets/15635498/23811986/1809a42e-05e1-11e7-8857-febaeed7330e.png)

Feature is ready for PR. So in case of any comment/wished - it is a good time :) Code to test is [here](https://github.com/sbcgua/abapGit/tree/diff-ux-improvements)

sbcgua:
Let it hang here for a while. Unless there are bug and wishes I'll close in a couple of days.

sbcgua:
I guess also can be closed

PR #657, opened 2017-03-10
Diff UX improvements and lcl_html_toolbar refactoring

sbcgua:
to #655 , #656 (described there)
Sorry for huge PR. I started filter feature one evening and virtually done it but then last 10% required some refactoring and it went on ... could not split it for several parts because the result as it is now was not obvious from the beginning - code lives :) .

### DIFF
--- a/src/zabapgit_gui_router.prog.abap
+++ b/src/zabapgit_gui_router.prog.abap
@@ -26,6 +26,7 @@ CLASS lcl_gui_router DEFINITION FINAL.
 
     METHODS get_page_diff
       IMPORTING iv_getdata     TYPE clike
+                iv_prev_page   TYPE clike
       RETURNING VALUE(ri_page) TYPE REF TO lif_gui_page
       RAISING   lcx_exception.
 
@@ -35,8 +36,8 @@ CLASS lcl_gui_router DEFINITION FINAL.
       RAISING   lcx_exception.
 
     METHODS get_page_stage
-      IMPORTING iv_key          TYPE lcl_persistence_repo=>ty_repo-key
-      RETURNING VALUE(ri_page)  TYPE REF TO lif_gui_page
+      IMPORTING iv_getdata     TYPE clike
+      RETURNING VALUE(ri_page) TYPE REF TO lif_gui_page
       RAISING   lcx_exception.
 
     METHODS get_page_db_by_name
@@ -72,7 +73,7 @@ CLASS lcl_gui_router IMPLEMENTATION.
     lv_url = iv_getdata. " TODO refactor
 
     CASE iv_action.
-        " General routing
+        " General PAGE routing
       WHEN gc_action-go_main                          " Go Main page
           OR gc_action-go_explore                     " Go Explore page
           OR gc_action-go_db                          " Go DB util page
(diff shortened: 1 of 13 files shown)


src/zabapgit_page_stage.prog.abap
```diff
@@ -25,8 +26,7 @@ CLASS lcl_gui_page_stage DEFINITION FINAL INHERITING FROM lcl_gui_page.
   PRIVATE SECTION.
     DATA: mo_repo  TYPE REF TO lcl_repo_online,
           ms_files TYPE ty_stage_files,
-          mo_stage TYPE REF TO lcl_stage,
-          mv_ts    TYPE timestamp.
+          mv_seed  TYPE string.
```
larshp: can you add a comment somewhere regarding what the `seed` variable is used for?

src/zabapgit_gui_router.prog.abap
```diff
@@ -85,11 +86,17 @@ CLASS lcl_gui_router IMPLEMENTATION.
...
-        ev_state = gc_event_state-new_page.
-      WHEN gc_action-go_stage.                        " Go Staging page
-        ei_page  = get_page_stage( lv_key ).
+        ei_page  = get_page_diff(
+          iv_getdata   = iv_getdata
+          iv_prev_page = iv_prev_page ).
         ev_state = gc_event_state-new_page_w_bookmark.
+      WHEN gc_action-go_stage.                        " Go Staging page
+        ei_page  = get_page_stage( iv_getdata ).
+        IF iv_prev_page = 'PAGE_DIFF'.
```
larshp: I think we need to look into if this special case can be avoided, but we can do that later, after this PR is merged

src/zabapgit_gui_router.prog.abap
```diff
@@ -289,9 +296,10 @@ CLASS lcl_gui_router IMPLEMENTATION.
 
     CREATE OBJECT lo_page
       EXPORTING
-        iv_key    = lv_key
-        is_file   = ls_file
-        is_object = ls_object.
+        iv_key           = lv_key
+        is_file          = ls_file
+        is_object        = ls_object
+        iv_supress_stage = boolc( iv_prev_page = 'PAGE_STAGE' ).
```
larshp: I think we need to look into if this special case can be avoided, but we can do that later, after this PR is merged

larshp:
A: it looks like the font size for the menu has been decreased? Plus there is more vertical whitespace?

image

src/zabapgit_page_stage.prog.abap
sbcgua: added at variable: `" Unique page id to bind JS sessionStorage`
at constructor: `" Generate based on time unless obtained from diff page`
and also some comment to diff_page (where it is also used).

Does it look OK ?

src/zabapgit_gui_router.prog.abap
sbcgua: Generally agree with both points. But it is not that clear how to do it. 
There is also one use case for DB_DISPLAY/DB_EDIT.

Thoughts:
- for DB pages EDIT and DISPLAY can be merged in one class - I think it would be reasonable. Then the issue go completely.
- for DIFF/STAGE ... I don't know. I guess there should be some common context accessible by any page. But this raises completely new challenges.

src/zabapgit_page_stage.prog.abap
larshp: yes, thanks 👍

src/zabapgit_gui_router.prog.abap
larshp: yeah, I am also not sure, but lets look at this in another issue/PR, it is not critical to fix it right now

sbcgua:
More vertical space: fixed

image

font size: yes, I descreased it on 1 pt. Imho looks better, no ? On the other hand now there are different font sizes on one page which is not ideal ... maybe reduce all the fonts ? Don't know ... arguable ... opinion needed.

This is how same font size looks (chrome) - better click to see 1:1 size

image

larshp:
I like old stuff 😄 and also things that are symmetric. So can we change it back to the old font size?

sbcgua:
No prob :) done. 

One consideration: I added padding around menu items and now items are nice and fat and easy to point - which is good. However they also take more space. And on my system the list of repos (burger) just does not fit the page :) Maybe padding should be reduced a bit ... let's see how it feels

larshp:
yeah, merging

larshp:
burger list: perhaps we can change it so it shows the list in multiple columns, eg. 10 per column

sbcgua:
burger list: I actually have some ideas towards burger icon, I'll create an issue for this

PR #663, opened 2017-03-15
Transport to branch

EduardoCopat:
References #263

You select a transport request, a new branch is created and the objects from the transport are pushed.

Important remarks:

- Core logic is at class `lcl_transport_to_branch` --> Needs a good review
- If, something fails after branch creation. The branch still remains in the repository.
- Using standard SAP popups, instead of HTML...
- Maybe show the stage list?


src/zabapgit_definitions.prog.abap
```diff
@@ -183,60 +189,61 @@ CONSTANTS: gc_newline TYPE abap_char1 VALUE cl_abap_char_utilities=>newline.
 
 CONSTANTS: gc_english TYPE spras VALUE 'E'.
 
-CONSTANTS: gc_root_dir    TYPE string VALUE '/',
-           gc_dot_abapgit TYPE string VALUE '.abapgit.xml' ##NO_TEXT,
-           gc_author_regex TYPE string VALUE '^([\w\s\.@\-_1-9]+) <(.*)> (\d{10}) .\d{4}$' ##NO_TEXT.
+CONSTANTS: gc_root_dir     TYPE string VALUE '/',
+           gc_dot_abapgit  TYPE string VALUE '.abapgit.xml' ##NO_TEXT,
+           gc_author_regex TYPE string VALUE '^([\w\s\.@\-_1-9]+) <(.*)> (\d{10}).\d{4}$' ##NO_TEXT.
```
larshp: looks like the space in the regex disappeared?

src/zabapgit_forms.prog.abap
```diff
@@ -137,6 +137,14 @@ FORM package_popup TABLES   tt_fields TYPE ty_sval_tt
...
 ENDFORM.                    "package_popup
 
+FORM branch_to_transport_popup TABLES   tt_fields TYPE ty_sval_tt
+                        USING    pv_code TYPE clike
+                        CHANGING cs_error TYPE svale
+                                 cv_show_popup TYPE c
+                        RAISING  lcx_exception ##called ##needed.
+* called dynamically from function module POPUP_GET_VALUES_USER_BUTTONS
+ENDFORM.
+
```
larshp: is this form needed? it is empty

src/zabapgit_gui_router.prog.abap
```diff
@@ -171,6 +174,21 @@ CLASS lcl_gui_router IMPLEMENTATION.
...
+        IF lt_tadir IS INITIAL.
+          lcx_exception=>raise( 'Canceled or List of objects is empty ' ).
+        ENDIF.
+        ls_transport_to_branch = lcl_popups=>popup_to_create_transp_branch(
+          it_transport_headers = lt_transport_headers
+          it_transport_objects = lt_tadir ).
+        lcl_services_repo=>transport_to_branch(
+          iv_repository_key      = lv_key
+          is_transport_to_branch = ls_transport_to_branch
+          it_transport_objects   = lt_tadir ).
```
larshp: can we move these 20 lines to lcl_services_repo=>transport_to_branch? would like to keep the router as clean as possible

src/zabapgit_services_trans_branch.prog.abap
```diff
@@ -0,0 +1,125 @@
...
+    ENDTRY.
+  ENDMETHOD.
+
+
+  METHOD add_new_and_changes_to_stage.
+    DATA ls_transport_object TYPE tadir.
+    DATA ls_local_file TYPE ty_file_item.
+
+    LOOP AT it_transport_objects INTO ls_transport_object.
+      IF ls_transport_object-delflag = abap_false.
```
larshp: looks like this condition can be moved to the LOOP, ie. `WHERE delflag = abap_false`

src/zabapgit_transport.prog.abap
```diff
@@ -205,3 +192,38 @@ CLASS lcl_transport IMPLEMENTATION.
   ENDMETHOD.
 
 ENDCLASS.
+CLASS lcl_transport_popup IMPLEMENTATION.
+  METHOD show.
```
larshp: yes, +1

src/zabapgit_services.prog.abap
```diff
@@ -2,8 +2,9 @@
 *&  Include           ZABAPGIT_SERVICES
 *&---------------------------------------------------------------------*
 
+INCLUDE zabapgit_services_git.
+INCLUDE zabapgit_services_trans_branch.
```
larshp: I would like to keep it out of the git include, as this is not a raw git command

I think it is okay to have it in its own include, perhaps rename it to something else?

src/zabapgit_definitions.prog.abap
EduardoCopat: Wow. Nice eye. But that is really bizarre. In my system it is correct! I'll need to adjust it manually in the branch I think

src/zabapgit_forms.prog.abap
EduardoCopat: If this form does not exist, popup_get_values_user_buttons short dumps. I'll add a comment

larshp: but there are no buttons in the popup? perhaps change it to a different function module?

src/zabapgit_gui_router.prog.abap
EduardoCopat: Agreed

src/zabapgit_forms.prog.abap
EduardoCopat: It has buttons... I'll look for other function modules then 👍

larshp: try `POPUP_GET_VALUES`

EduardoCopat: That works, at the cost of not having a custom button with a fancy icon

src/zabapgit_services_trans_branch.prog.abap
```diff
@@ -0,0 +1,125 @@
...
+    ls_stage_objects = lcl_stage_logic=>get( io_repository ).
+
+    add_new_and_changes_to_stage(
+      it_transport_objects = it_transport_objects
+      io_stage             = lo_stage
+      is_stage_objects     = ls_stage_objects ).
+
+    "Stage remote files as removal, for they could be deleted
+    LOOP AT ls_stage_objects-remote INTO ls_remote_file.
+      lo_stage->rm(
```
EduardoCopat: I agree, it is not clear and the comment is bad. I need help with this one.

So, when I call `lcl_stage_logic=>get( io_repository )`, I get access to these "local" and "remote" files.

Local files are the new or modified objects. Remote files are objects who are in the remote, but not locally. Thus, they are commited as "removed".

The transport list contains deleted objects, but I couldn't map them to these "remote files".

src/zabapgit_services.prog.abap
sbcgua: How about `zabapgit_trans_utils` ? But any would agree to anything without 'services' in the name :) And also it should be moved to main prog.

src/zabapgit_services_trans_branch.prog.abap
sbcgua: Ah ! Hmmm, interesting problem. Well, that is certainly not `rm` for all remotes. RM will actually delete files from repo.

I think it is better to use lcl_repo->status() instead of stage_logic. It has all the data in a convenient and unified form. Have a look into `lcl_repo_content_browser->list(), build_repo_items_online(), filter_changes()` - it prepares the data for main page 

I think the logic should be:
- loop through transport list
- if `delflag = false` -> find the file, ensure local status is M or A and  `stage->add`
- if `delflag = true` -> find the file, ensure local status is D and `stage-rm`
- I would also put a lot of assertions around just to catch a bug if there is one ... like we assert to find the file with status XXX, if not - exception.

conseqently all can be done in what is currently called `add_new_and_changes_to_stage`.

P.S. if I understand the `delflag` well. I'm not an expert in internals :)

sbcgua: image

sbcgua: And `status` has the object mapping always

sbcgua:
@EduardoCopat 'review changes' - I guess I pressed `Comment only` at some stage and cannot redo it after ... yet many features to learn :)

src/zabapgit_services_trans_branch.prog.abap
larshp: I daily have problems with wrong state in lcl_repo->status(), however stage_logic always shows the correct result

sbcgua: hmmm ... could you log an issue with example ? :)
Any the obstacle for the case is that stage_logic->remote does not have object.

larshp: my test system is messed up, part of the problem is wrong time settings

anyhow, in the stage_logic it already calculates the actions for the files, so it can be reused

src/zabapgit_services.prog.abap
EduardoCopat: What if I move the class `lcl_transport_to_branch` to already existing include `zabapgit_transport`

larshp: sounds good to me

src/zabapgit_services_trans_branch.prog.abap
EduardoCopat: Thanks for the inputs @sbcgua. `lcl_repo->status` worked pretty well to me.

src/zabapgit_transport.prog.abap
```diff
@@ -203,5 +183,140 @@ CLASS lcl_transport IMPLEMENTATION.
...
+  ENDMETHOD.
+
+  METHOD generate_commit_message.
+    rs_comment-committer-name  = sy-uname.
+    rs_comment-committer-email = |{ rs_comment-committer-name }@localhost|.
+    rs_comment-comment         = is_transport_to_branch-commit_text.
+  ENDMETHOD.
+
+
+  METHOD stage_transport_objects.
```
EduardoCopat: In a next pull request, I'll extract this method to a class and put under test. This is the core logic...

EduardoCopat:
@sbcgua @larshp ready for re-review

src/zabapgit_gui_router.prog.abap
```diff
@@ -64,10 +64,10 @@ CLASS lcl_gui_router IMPLEMENTATION.
...
   METHOD on_event.
 
-    DATA: lv_url     TYPE string,
-          lv_key     TYPE lcl_persistence_repo=>ty_repo-key,
-          ls_db      TYPE lcl_persistence_db=>ty_content,
-          ls_item    TYPE ty_item.
+    DATA: lv_url                 TYPE string,
+          lv_key                 TYPE lcl_persistence_repo=>ty_repo-key,
+          ls_db                  TYPE lcl_persistence_db=>ty_content,
+          ls_item                TYPE ty_item.
```
sbcgua: Could you remove the spaces back pls ? :)

src/zabapgit_transport.prog.abap
```diff
@@ -203,5 +183,140 @@ CLASS lcl_transport IMPLEMENTATION.
...
+
+          io_stage->add(
+            iv_path     = ls_local_file-file-path
+            iv_filename = ls_local_file-file-filename
+            iv_data     = ls_local_file-file-data ).
+        WHEN gc_state-deleted.
+          ASSERT ls_transport_object-delflag = abap_true.
+          io_stage->rm(
+            iv_path     = ls_object_status-path
+            iv_filename = ls_object_status-filename ).
```
sbcgua: I'd add `WHEN OTHERS. raise ...`, just to catch potential bugs.

EduardoCopat: Yeah, good one. What is the `gc_state-mixed` (`*`) ? Should I account for it?

sbcgua: No. It should not appear in bare status. This is the status which means that underlying object contain multiple statuses (e.g. some FM has several files: one include changed, another added - so the agregated status of FM is mixed). It can appear only in aggregations in the main page (see `_reduce_state` macro) and cannot be in `status()` results as it contain one file per line. I hope it clarifies ... So I guess just put `when others` - in case of exception the case will be more specific to study :)

sbcgua:
From my side look fine :) Thanks @EduardoCopat !

PR #815, opened 2017-07-06
#811 progress indicator for encoding objects

bigld:


src/zabapgit_git.prog.abap
```diff
@@ -1043,7 +1043,18 @@ CLASS lcl_git_pack IMPLEMENTATION.
...
+
     LOOP AT it_objects ASSIGNING <ls_object>.
+
+      lv_objects_processed = sy-tabix.
+
+      cl_progress_indicator=>progress_indicate(
+        EXPORTING
+          i_text               = |encoding objects &1% ( &2 of &3 )|
+          i_processed          = lv_objects_processed
+          i_total              = lv_objects_total ).
```
larshp: hi, was thinking perhaps to add code that only updates the progress indicator each second, as this loop is quite fast and showing the progress indicator is expensive.

bigld: FM SAPGUI_PROGRESS_INDICATOR is called only every 10 seconds - this should be ok...

larshp: oh, yeah, sorry

I'm not sure if CL_PROGRESS_INDICATOR exists in 702, but lets try merging it

bigld: Just checked...
image

PR #820, opened 2017-07-08
Issue 795

christianguenter2:
#795 
Replaced call to cl_http_utility=>if_http_utility~string_to_fields with custom implementation

### DIFF
--- a/src/zabapgit_unit_test.prog.abap
+++ b/src/zabapgit_unit_test.prog.abap
@@ -1832,13 +1832,12 @@ CLASS ltcl_html_action_utils IMPLEMENTATION.
 
     _when_fields_are_parsed( ).
 
-    _then_fields_should_be(:
-      index = 1 name = `COMMITTER_NAME`   value = `Albert Schweitzer` 															 ),
-      index = 2 name = `COMMITTER_EMAIL`  value = `albert.schweitzer@googlemail.com`								 ),
-      index = 3 name = `COMMENT`          value = `dummy comment` 																	 ),
-      index = 4 name = `BODY`             value = `Message body<<new>><<new>>with line break<<new>>` ),
-      index = 5 name = `AUTHOR_NAME`      value = `Karl Klammer`																		 ),
-      index = 6 name = `AUTHOR_EMAIL`     value = `karl@klammer.com`																 ).
+    _then_fields_should_be( index = 1 name = `COMMITTER_NAME`   value = `Albert Schweitzer` ).
+    _then_fields_should_be( index = 2 name = `COMMITTER_EMAIL`  value = `albert.schweitzer@googlemail.com` ).
+    _then_fields_should_be( index = 3 name = `COMMENT`          value = `dummy comment` ).
+    _then_fields_should_be( index = 4 name = `BODY`             value = `Message body<<new>><<new>>with line break<<new>>` ).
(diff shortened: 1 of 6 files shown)


src/zabapgit_unit_test.prog.abap
```diff
@@ -1793,6 +1811,90 @@ CLASS ltcl_html_action_utils IMPLEMENTATION.
...
+
+    _when_fields_are_parsed( ).
+
+    _then_fields_should_be(:
+      index = 1 name = `COMMITTER_NAME`   value = `Albert Schweitzer` 															 ),
+      index = 2 name = `COMMITTER_EMAIL`  value = `albert.schweitzer@googlemail.com`								 ),
+      index = 3 name = `COMMENT`          value = `dummy comment` 																	 ),
+      index = 4 name = `BODY`             value = `Message body<<new>><<new>>with line break<<new>>` ),
+      index = 5 name = `AUTHOR_NAME`      value = `Karl Klammer`																		 ),
+      index = 6 name = `AUTHOR_EMAIL`     value = `karl@klammer.com`																 ).
```
larshp: hmm, does this compile? Anyhow, can we unchain this to 6 method calls? I easily get confused when the colon operator is used

also see https://help.sap.com/http.svc/rc/abapdocu_751_index_htm/7.51/en-US/abenchained_statements_guidl.htm

christianguenter2: yep, it does compile. In my opinion in this case the colon operator leads to cleaner code. Anyway I unchain the method calls. What worries me more is that I can't use non 7 bit ASCII characters in my unit tests to test the german umlauts. Any suggestions about this?

larshp:
A: we should at some point replace all uses of the standard method,

image

B: non-7 bit, perhaps use hex values instead? would like to keep the code to simple characters

christianguenter2:
A: Oh, didn't recognize the others uses. I try to replace them.
B: Ok, I try it the hex way ;-)

src/zabapgit_unit_test.prog.abap
larshp: ahh, okay, the `),` are hiding in column 200+

christianguenter2: strange, I use a eclipse plugin which makes indentation easier but somehow it screws up whitespaces and tabs...

larshp: tip: run https://github.com/larshp/abapOpenChecks/wiki/ZCL_AOC_CHECK_09 to check for tabs

christianguenter2:
Not quite sure whether the hex handling for umlauts in the test class makes any sense. Can someone check?

src/zabapgit_unit_test.prog.abap
christianguenter2: I did, but somehow missed the findings...

larshp:
hex handling for umlauts: looks good to me 👍

larshp:
All looks good to me, thanks, will merge tomorrow

larshp:
When editing XML via Advanced -> Database Util, it produces invalid xml

Guess the string has to be URL decoded

%3D = "="
%3F = "?"

image

christianguenter2:
I added some URL decoding logic. Not quite sure whether it's the right place to do it. Further review highly appreciated.

larshp:
thanks, LGTM, merging, squashing

PR #883, opened 2017-08-04
#879 DIAL Serialization

christianguenter2:
#879


larshp:
hi, the DIAL is not inserted into TADIR, add `tadir_insert( iv_package ).` somewhere in the `deserialize` method

test repository https://github.com/larshp/SHI3

christianguenter2:
I added tadir_insert. 

DIAL objects are unfortunately always marked with diffs, but I can't see why.
image

larshp:
diff probably due to https://github.com/larshp/abapGit/issues/799

PR #915, opened 2017-09-05
ENHS: disabled dark mode to allow popup for transport

apex8:


larshp:
#913

apex8:
OK yeah that would be great, but I´m not sure if the standard code can
support this. If you don´t work with local objects you always need a
transport request. I haven´t had a look, but the standard infrastructure
would need to provide a parameter for the transport request for each kind
of object, to get rid of the transport popup.

PR #934, opened 2017-09-27
Replacing per object 'overwrite local Object' dialog with checkbox list

apex8:
When Pulling local objects are checked if they have been modified. If so there was a confirmation popup for each objects. This has been replaced with a popup table allowing to have a look at all modified objects at once and selecting a checkbox to confirm that all selected objects should be overwritten.

### DIFF
--- a/src/zabapgit_popups.prog.abap
+++ b/src/zabapgit_popups.prog.abap
@@ -85,8 +85,8 @@ CLASS lcl_popups DEFINITION FINAL.
   PRIVATE SECTION.
     TYPES: ty_sval_tt TYPE STANDARD TABLE OF sval WITH DEFAULT KEY.
 
-    TYPES BEGIN OF t_popup_select_list.
-      TYPES selected TYPE flag.
+    TYPES: BEGIN OF t_popup_select_list,
+        selected TYPE flag.
       INCLUDE TYPE lif_defs=>ty_result.
     TYPES END OF t_popup_select_list.
 
--- a/src/zabapgit_popups.prog.abap
+++ b/src/zabapgit_popups.prog.abap
@@ -76,18 +76,46 @@ CLASS lcl_popups DEFINITION FINAL.
         RAISING   lcx_exception
                   lcx_cancel,
       popup_to_select_transports
-        RETURNING VALUE(rt_trkorr) TYPE trwbo_request_headers.
+        RETURNING VALUE(rt_trkorr) TYPE trwbo_request_headers,
+      popup_select_obj_overwrite
+        IMPORTING it_list TYPE lif_defs=>ty_results_tt
+        RETURNING VALUE(rt_list) TYPE lif_defs=>ty_results_tt
+        RAISING   lcx_exception.
 
   PRIVATE SECTION.
     TYPES: ty_sval_tt TYPE STANDARD TABLE OF sval WITH DEFAULT KEY.
 
-    CLASS-METHODS: add_field
-      IMPORTING iv_tabname    TYPE sval-tabname
-                iv_fieldname  TYPE sval-fieldname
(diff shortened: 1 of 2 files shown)


EduardoCopat:
Related to #490

src/zabapgit_popups.prog.abap
```diff
@@ -699,4 +727,114 @@ CLASS lcl_popups IMPLEMENTATION.
...
+  ENDMETHOD.
+
+  METHOD on_select_list_link_click.
+    DATA:
+          lsr_line TYPE REF TO t_popup_select_list,
+          lv_line  TYPE sytabix.
+
+    lv_line = row.
+
+    READ TABLE mtr_select_list->* REFERENCE INTO lsr_line INDEX lv_line.
```
EduardoCopat: The conventions on the project are mt_ and lr_

EduardoCopat:
I just did some exploratory testing and it looks great.
Two comments from my side:

1.  Is it possible to have a button to mark all?
2. There is one failing check on Travis from abapGit

christianguenter2:
I think the popup should only appear when there are any objects

image

larshp:
regarding the error on Travis CI, I will look into that tonight, it might be an error in abaplint

src/zabapgit_popups.prog.abap
```diff
@@ -76,18 +76,46 @@ CLASS lcl_popups DEFINITION FINAL.
...
-      IMPORTING iv_tabname    TYPE sval-tabname
-                iv_fieldname  TYPE sval-fieldname
-                iv_fieldtext  TYPE sval-fieldtext
-                iv_value      TYPE clike DEFAULT ''
-                iv_field_attr TYPE sval-field_attr DEFAULT ''
-      CHANGING  ct_fields     TYPE ty_sval_tt.
+    TYPES BEGIN OF t_popup_select_list.
+      TYPES selected TYPE flag.
+      INCLUDE TYPE lif_defs=>ty_result.
+    TYPES END OF t_popup_select_list.
```
larshp: Hi, can you change this part to

```abap
TYPES: BEGIN OF t_popup_select_list,
         selected TYPE flag.
    INCLUDE TYPE lif_defs=>ty_result.
TYPES END OF t_popup_select_list.
```

it will fix the abaplint error

apex8:
Issues should be solved now.
A select all button seems to require more investigation. I tried to enable the button bar, which isn´t a problem in fullscreen mode. But since it´s shown as popup, the bar doesn´t appear. I will have deeper look into this once I find some time.

larshp:
Looks good to me, the select-all feature can be added in a follow up PR, the current solution is better than what we currently have in the code 😄

christianguenter2:
@apex8 
Select all should be quite straightforward. Just copy the PF-Status, add the fcodes for select and deselect all and handle the fcodes appropriately.

You can have a look here.
https://github.com/christianguenter2/ABAP_DEMO/blob/c1f1c010822ef76fff9aa46d4a2f34b55f1a5844/z_test_alv_popup.prog.abap

image

larshp:
It wont work with a PF-Status, it should be possible to easily install ZABAPGIT, eg. via copy pasting

christianguenter2:
You're right I forgot about that

christianguenter2:
Yet another approach would be to use a modal dialogbox container popup. Then you can use the add_function method of cl_salv_table. Something like this.

image

https://gitlab.com/christianguenter/ABAP_Demo_ALV_Popup/blob/master/zcl_test_alv_popup.clas.abap

larshp:
Will merge this, @christianguenter2(or someone else) feel free to submit a follow up PR with the select-all feature

PR #979, opened 2017-10-09
Support object type SRFC - RFC Service #977

christianguenter2:
#977

### DIFF
--- a/src/zabapgit_object_serializing.prog.abap
+++ b/src/zabapgit_object_serializing.prog.abap
@@ -48,6 +48,7 @@ INCLUDE zabapgit_object_shma.
 INCLUDE zabapgit_object_sicf.
 INCLUDE zabapgit_object_smim.
 INCLUDE zabapgit_object_splo.
+INCLUDE zabapgit_object_srfc.
 INCLUDE zabapgit_object_ssfo.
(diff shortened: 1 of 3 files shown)


larshp:
When installing a remote enabled RFC function module, I've previously experienced the SRFC object to be generated automatically, any thoughts on this?

christianguenter2:
I don't have a clue.

larshp:
another approach is to skip all SRFC objects, hmm, not sure what is best

christianguenter2:
I just did some experiments with RFC FMs, but there were no SRFC objects been created. Are there any special steps required?

larshp:
usually I've seen them being generated automatically when pulling the RFC FM, but not sure if it depends on version/system settings

christianguenter2:
It seems that it can be activated via profile parameter ucon/rfc/active.
Class CL_UCON_HELPER method IS_UCON_RFC_ACTIVE checks that.

There's also the report RS_UCON_PHASE_CUSTOM.
image

I'm not yet sure how this fits in the big picture. I'll try to figure it out.

larshp:
okay, with the possibility of it not being generated automatically, I think we should add the feature

PR #981, opened 2017-10-10
Fix Issue #980 and add test code

christianguenter2:
#980

### DIFF
--- a/src/zabapgit_unit_test.prog.abap
+++ b/src/zabapgit_unit_test.prog.abap
@@ -2652,8 +2652,6 @@ CLASS ltcl_oo_serialize IMPLEMENTATION.
 
   METHOD one_line_include.
 
-    FIELD-SYMBOLS: <source> LIKE LINE OF mt_source.
-
     _given_source_is( `*"* use this source file for your ABAP unit test classes` ).
 
     _when_skip_is_calculated( ).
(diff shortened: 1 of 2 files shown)


src/zabapgit_unit_test.prog.abap
```diff
@@ -38,6 +38,11 @@ DEFINE _append_result.
   <result>-filename = &8.
 END-OF-DEFINITION.
 
+DEFINE _given_source_is.
```
larshp: can we move this macro to a local method instead? The code will not look as pretty, but in this case it is easy to get rid of the macro, also chaining should mainly be used for declarations, https://help.sap.com/http.svc/rc/abapdocu_751_index_htm/7.51/en-US/abenchained_statements_guidl.htm

christianguenter2:
I changed the macro to a method. Now we get this build error.
image

This is line 2609
https://github.com/larshp/abapGit/blob/1c7e4c9fe055635af5a24defc11a57fb6d569e7c/src/zabapgit_unit_test.prog.abap#L2609

Not quite sure what the issue is.

larshp:
I'll take a look tomorrow

larshp:
opened https://github.com/larshp/abaplint/issues/187

Issue #982, opened 2017-10-10
Translations of program text pools are deleted upon activation


thomas-erdoesi:
Translations should be inserted as active text pools, only the text pool of the master language can be inactive. Otherwise translations will be deleted upon activation.

thomas-erdoesi:
Proposed fix: Pull request #983

PR #983, opened 2017-10-10
Issue 982: Translations deleted upon activation

thomas-erdoesi:
Translations should be inserted as active text pools, only the text pool of the master language can be inactive. Otherwise translations will be deleted upon activation.


christianguenter2:
#982

PR #987, opened 2017-10-10
Fixes method reorder when deserializing class

### LINKED ISSUE #756: Text in source code based editor is reorganized on activation
This is not directly an abapGit problem, but something that you stumble upon using abapGit and it kind of ruins `git diff`: When you activate a global class in eclipse or in SE24 using the source code-based editor ~20% of the time the source code will get completely reorganized. This includes:
- Method implementations are reordered alphabetically (which is very annoying if you wrote them in order of visibility and static/instance-bound)
- Method implementations are separated by exactly two empty lines (this includes two lines before the first method)
- The class name in the implementation part is suddenly in all uppercase

I assume this is a remainder of the form based editor and splitting a class in X includes. Currently I have this problem for every file in a project, because I exported the whole package and imported it on another system which I assume triggered a _fresh_ ABAP load generation (which otherwise is only done sometimes).

Any idea how to prevent this other than copying the whole class just in case it gets reorganized?

image

EduardoCopat:
Fixes #756 

Pretty simple, but took many hours of debugging. Thanks @insidy for the help.

@flaiker Could you please retest in your environment?

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,11 @@ Legend
 + : added
 - : removed
 
+2017-10-06 v1.43.1
+------------------
+* Clear some DEVC fields
+* Downport
+
 2017-10-05 v1.43.0
 ------------------
 + DEVC support
--- a/src/zabapgit_definitions.prog.abap
+++ b/src/zabapgit_definitions.prog.abap
@@ -7,7 +7,7 @@ TYPE-POOLS seop.
 INTERFACE lif_defs.
 
   CONSTANTS: gc_xml_version  TYPE string VALUE 'v1.0.0',    "#EC NOTEXT
-             gc_abap_version TYPE string VALUE 'v1.43.0'.   "#EC NOTEXT
+             gc_abap_version TYPE string VALUE 'v1.43.1'.   "#EC NOTEXT
 
   TYPES: ty_type    TYPE c LENGTH 6,
          ty_bitbyte TYPE c LENGTH 8,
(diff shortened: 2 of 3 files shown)


flaiker:
Sure, I'll try it out. Thanks for the detective work!

What's the test case here though? #756 is kind of independent of abapGit. I assume this patch fixes the following:
- New remote version of a class is pulled
- Class gets deserialized and activated
- Because of some black magic buffer shenanigans the method order gets changed to alphabetic and even though the class was just pulled it shows up as a diff.

And it does not fix eclipse or SE24 reorganizing the method order _in seemingly random cases_ on activation. Correct?

EduardoCopat:
Correct.

The test case is the following:

Change the method order, either directly in github or in ADT, to not alphabetically. Then, check if the method order remains the same.

Also, check if the class name does not change. Open both in ADT and SE24

EduardoCopat:
@larshp Now it is upported to clas_new

flaiker:
Sorry for the late reply. I was unable to reproduce the error in the test case you mentioned.

Do you think one could add this fix somewhere in the SAP code so this doesn't happen outside of abapGit as well?

insidy:
You mean in SE80/SE24 to avoid reordering methods?

The problem is that the handling of class source code via SE24 is considerably different to ADT, even when not in form mode. 
Also, if the issue is fixed then it would only be available in the next netweaver release. If downport is possible then it can be requested, but that takes a while.

So I'd suggest to stick to ADT when possible (and create a plugin/ADT-compatible version of abapgit).

flaiker:
@insidy I mean in ADT, it's really annoying at times: https://streamable.com/ee369

insidy:
@flaiker oh.. that is annoying. Do you know in which netweaver version that happened or any idea on how to reproduce?

I've tried the same steps but it didn't reordered in NW 7.51

flaiker:
I definitely have it on 7.40 SP6 and 7.50 SP1. I'll have to try on 7.51. It doesn't happen all the time, just sometimes. In the video I recorded for ~2 minutes, adding / removing parts of the class and activating in between until it happened.

Issue #988, opened 2017-10-10
Support UCSA - communication assemblies


christianguenter2:
https://help.sap.com/viewer/1ca554ffe75a4d44a7bb882b5454236f/7.51.1/en-US/69f70acfd3424c74809634c85b3c02e5.html

PR #989, opened 2017-10-10
UCSA support

christianguenter2:
#988

### DIFF
--- a/src/zabapgit_object_ucsa.prog.abap
+++ b/src/zabapgit_object_ucsa.prog.abap
@@ -1,0 +1,168 @@
+*&---------------------------------------------------------------------*
+*& Include zabapgit_object_ucsa
+*&---------------------------------------------------------------------*
+
+CLASS lcl_object_ucsa DEFINITION INHERITING FROM lcl_objects_super FINAL.
+
+  PUBLIC SECTION.
+    INTERFACES lif_object.
+
+ENDCLASS.
+
+CLASS lcl_object_ucsa IMPLEMENTATION.
+
+  METHOD lif_object~has_changed_since.
+
+    rv_changed = abap_true.
+
+  ENDMETHOD.
+
+  METHOD lif_object~changed_by.
+
+    rv_user = c_user_unknown.
+
+  ENDMETHOD.
+
+  METHOD lif_object~get_metadata.
+
+    rs_metadata = get_metadata( ).
+    rs_metadata-delete_tadir = abap_true.
+
+  ENDMETHOD.
+
+  METHOD lif_object~exists.
+
+    DATA: lv_id TYPE uconservid.
+
+    lv_id = ms_item-obj_name.
+
+    TRY.
+        DATA(lo_persistence) = cl_ucon_sa_db_persist=>if_ucon_sa_persist~get_instance(  lv_id ).
+
+        lo_persistence->load( version  = 'A'
+                              language = sy-langu ).
+
+      CATCH cx_ucon_base.
+        rv_bool = abap_false.
+    ENDTRY.
+
+    rv_bool = abap_true.
+
+  ENDMETHOD.
+
+  METHOD lif_object~serialize.
+
+    DATA: lv_id   TYPE uconservid,
+          lx_root TYPE REF TO cx_root,
+          lv_text TYPE string,
+          sa      TYPE uconservascomplete.
+
+    lv_id = ms_item-obj_name.
+
+    TRY.
+        DATA(lo_persistence) = cl_ucon_sa_db_persist=>if_ucon_sa_persist~get_instance( lv_id ).
+
+        lo_persistence->load(
+          EXPORTING
+            version  = 'A'
+            language = sy-langu
+          IMPORTING
+            sa       = sa ).
+
+        CLEAR: sa-header-createdby,
+               sa-header-createdon,
+               sa-header-createdat,
+               sa-header-changedby,
+               sa-header-changedon,
+               sa-header-changedat.
+
+        io_xml->add( iv_name = 'UCSA'
+                     ig_data = sa ).
+
+      CATCH cx_ucon_base INTO lx_root.
+        lv_text = lx_root->get_text( ).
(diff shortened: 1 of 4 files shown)


src/zabapgit_object_ucsa.prog.abap
```diff
@@ -0,0 +1,240 @@
...
+
+    DATA: lv_id          TYPE ty_id,
+          lo_persistence TYPE REF TO object.
+
+    lv_id = ms_item-obj_name.
+
+    TRY.
+        lo_persistence = get_persistence( lv_id ).
+
+        CALL METHOD lo_persistence->('IF_UCON_SA_PERSIST~LOAD')
```
EduardoCopat: Why calling the method in this way and not directly lo_persistence->if_ucon_sa_persist~load ?

src/zabapgit_object_ucsa.prog.abap
```diff
@@ -0,0 +1,240 @@
...
+          EXPORTING
+            version  = 'A'
```
EduardoCopat: What does 'A' stand for? you could use a constant

src/zabapgit_object_ucsa.prog.abap
```diff
@@ -0,0 +1,240 @@
...
+
+  METHOD lif_object~serialize.
+
+    DATA: lv_id          TYPE ty_id,
+          lx_root        TYPE REF TO cx_root,
+          lv_text        TYPE string,
+          lo_persistence TYPE REF TO object,
+          lr_sa          TYPE REF TO data.
+
+    FIELD-SYMBOLS: <sa>     TYPE any,
```
EduardoCopat: I think you can rename all instances of `sa` to a more meaningful name. What does SA stand for?

src/zabapgit_object_ucsa.prog.abap
```diff
@@ -0,0 +1,240 @@
...
+          EXPORTING
+            version  = 'A'
+            language = sy-langu
+          IMPORTING
+            sa       = <sa>.
+
+        ASSIGN COMPONENT 'HEADER' OF STRUCTURE <sa>
+               TO <header>.
+        ASSERT sy-subrc = 0.
+
```
EduardoCopat: Why is field not cleared here?

src/zabapgit_object_ucsa.prog.abap
```diff
@@ -0,0 +1,240 @@
...
+        lo_persistence = get_persistence( lv_id ).
+
+        CALL METHOD lo_persistence->('IF_UCON_SA_PERSIST~LOAD')
+          EXPORTING
+            version  = 'A'
+            language = sy-langu
+          IMPORTING
+            sa       = <sa>.
+
+        ASSIGN COMPONENT 'HEADER' OF STRUCTURE <sa>
```
EduardoCopat: You could extract all these `ASSIGN COMPONENT` repetitions to a private method

src/zabapgit_object_ucsa.prog.abap
EduardoCopat: Same question to other places where you do this

christianguenter2: Because of downward compatibility. Interface IF_UCON_SA_PERSIST isn't present e.g. in NW 7.31.

src/zabapgit_object_ucsa.prog.abap
christianguenter2: Field? This is the whole header. There are other components in it which shouldn't be cleared.

src/zabapgit_object_ucsa.prog.abap
christianguenter2: Introduced global constant zif_abapgit_definitions=>gc_version-active

src/zabapgit_object_ucsa.prog.abap
christianguenter2: changed to complete_comm_assembly

src/zabapgit_object_ucsa.prog.abap
christianguenter2: extracted to method clear_dynamic_fields

src/zabapgit_object_ucsa.prog.abap
EduardoCopat: Could you add a comment with this statement 😃 ?

src/zabapgit_object_ucsa.prog.abap
```diff
@@ -0,0 +1,251 @@
...
+  METHOD clear_dynamic_fields.
+
+    FIELD-SYMBOLS: <header> TYPE any,
+                   <field>  TYPE any.
+
+    ASSIGN COMPONENT 'HEADER' OF STRUCTURE cs_complete_comm_assembly
+           TO <header>.
+    ASSERT sy-subrc = 0.
+
+    ASSIGN COMPONENT 'CREATEDBY' OF STRUCTURE <header>
```
EduardoCopat: Hi, @christianguenter2, looking better now.

I wonder if you could extract this into a private method like this:

METHODS clear_dynamic_field
	importing
		iv_field_name type string.

METHOD clear_dynamic_field.
FIELD-Symbols: ...

ASSIGN COMPONENT iv_fieldname OF STRUCTURE <header>
           TO <field>.
    ASSERT sy-subrc = 0.
    CLEAR: <field>.
ENDMETHOD.


Still, header should not be called is it not clean. (why?)

EduardoCopat:
Added two small comments

src/zabapgit_object_ucsa.prog.abap
christianguenter2: Sure, did this.

src/zabapgit_object_ucsa.prog.abap
christianguenter2: Good point. I extracted a new private method.

But I don't get this one.
"Still, header should not be called is it not clean. (why?)"
What do you mean by that?

christianguenter2:
Did some changes, please see comments.

PR #1007, opened 2017-10-16
Move objects across folders and packages

### LINKED ISSUE #842: Moving objects does not work, when switching branches
When the package assignment of an object has been changed, a branch switch does not work anymore.

Example with current version 1.37.5:

Preparation:
- Create package `zassign`
- Create package `zassign_one` (sub package of `zassign`)
- Create package `zassign_two` (sub package of `zassign`)
- Create Data Element `zassign_element` in package `zassign_one`

In abapGit do:
- Commit the package `zassign` to a git repo with abapGit (branch `master`)
- Create and checkout a new branch named `reassignment`

Now, do the following
- In SE80 right click `zassign_element` -> 'Other Functions' -> 'Change Package Assignment'
- Set `zassign_two` as new package
- Commit this change to branch `reassignment` (It will remove `zassign_element` from `zassign_one` and add it to `zassign_two` in the repository)

Now the weird stuff happens (at least for my understanding):
- Switch branch from `reassignment` to the original `master` branch. This will show you a popup, that `zassign_element` will be replaced. Click `OK`.
- Check SE80 and you will see, that `zassign_element` has been moved back from `zassign_two` to `zassign_one`

The other way round does not work anymore:
- Switch branch from `master` back to `reassignment`
- Instead of automatically "move" `zassign_element` from `zassign_one` to `zassign_two` it stays in `zassign_one` and abapGit interprets this as a local change compared to the repository.

Is this considered a bug or is there something I do not understand?
Best regards,
Tobias

EduardoCopat:
Fixes #842

See Issue for a detailed test case. It should also work when moving files in the remote (i.e. directly on github).

This is a change that is small, but could have big side effects. It needs some testing from other people :)

`run_checks` still raises warning messages regarding duplication

### DIFF
--- a/src/zabapgit_file_status.prog.abap
+++ b/src/zabapgit_file_status.prog.abap
@@ -52,6 +52,10 @@ CLASS lcl_file_status DEFINITION FINAL
                   it_state         TYPE zif_abapgit_definitions=>ty_file_signatures_ts
         RETURNING VALUE(rs_result) TYPE zif_abapgit_definitions=>ty_result
         RAISING   zcx_abapgit_exception,
+      build_old_folder_deletion
+        IMPORTING is_local         TYPE zif_abapgit_definitions=>ty_file_item
+                  is_remote        TYPE zif_abapgit_definitions=>ty_file
+        RETURNING VALUE(rs_result) TYPE zif_abapgit_definitions=>ty_result,
       identify_object
         IMPORTING iv_filename TYPE string
                   iv_path     TYPE string
@@ -151,8 +155,8 @@ CLASS lcl_file_status IMPLEMENTATION.
 
   METHOD status.
 
-    DATA: lv_index        LIKE sy-tabix,
-          lo_dot_abapgit  TYPE REF TO lcl_dot_abapgit.
+    DATA: lv_index       LIKE sy-tabix,
+          lo_dot_abapgit TYPE REF TO lcl_dot_abapgit.
 
     FIELD-SYMBOLS <ls_result> LIKE LINE OF rt_results.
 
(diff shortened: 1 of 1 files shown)


EduardoCopat:
@larshp, @WurstMitDurst Could you please review and test?

EduardoCopat:
There is a bug where reseting to local using this changes does not reset to the original state (It marks the moved object as Deletion)

larshp:
Hi,
Can you add a bit more details about the problem and implementation? I imagined that the fix would be to move some object to the correct package when deserializing, but it looks like the implementation does something else? Perhaps this happens somewhere else in the callstack, but triggered by your changes?

EduardoCopat:
Lars, yes, that's the idea.

Suppose you have `PackageA/foo.abap` and you move it  to `PackageB/foo.abap`. abapGit currently only detects this as an _addition_ of `foo.abap` to `PackageB`.

This code, in theory, detects that foo has been moved. The logic: it exists locally, but in a different folder compared to the remote.

As I mentioned in the PR, there may be unpredictable side effects. No problems to close it if it does or you think it does.

larshp:
Clone https://github.com/larshp/SHI3.git, switch to branch "moved", click pull, result:

image

The warnings and the status of files is okay, but from my point of view, the error is that the ZSHI3 report was not moved to the correct package during pull

EduardoCopat:
Hi, Lars, I will look into it. Thanks

EduardoCopat:
Lars, just had the time to test it now. 

Are you sure you were using the `EduardoCopat:move-objects-across-folders` branch? I just tested here with your example and it worked

image

larshp:
I tested it without

I think the status and warnings currently in abapGit are okay, we should fix the package assignment after pull

EduardoCopat:
Unfortunately, I will not have the time to continue developing this PR so I'm closing it 😞.

PR #1015, opened 2017-10-20
Callbacks

### LINKED ISSUE #686: Callbacks for installation / update / removal
I think it would be a nice feature to have some hooks for abapGit "events" that can be subscribed to by a project.
For example a library that can be installed using abapGit contains customizing tables. Afaik you cannot check in table contents to abapGit (?) so for the default values you would probably write a report that inserts those values into the db table that has to be executed manually. If there was an installation event this could be triggered automatically after activation. Same for updating a repository (data migration) or removal (cleanup db / generated source code / file system / ...).

I was thinking of something like this (global interface might be a bit too tricky though):
```abap
INTERFACE zif_abapgit_repo_event_listener PUBLIC.
  METHODS:
    on_installation IMPORTING iv_package TYPE devclass,
    on_update IMPORTING iv_old_tag TYPE string OPTIONAL
                        iv_new_tag TYPE string OPTIONAL
                        iv_old_sha1 TYPE string
                        iv_new_sha1 TYPE string,
    on_removal.
ENDINTERFACE.
```
```xml
 <asx:values>
  <DATA>
   (...)
   <CALLBACK_CLASS>ZCL_MY_LISTENER</CALLBACK_CLASS>
  </DATA>
 </asx:values>
</asx:abap>
```

flaiker:
Fixes #686

I added inline documentation for the implementation details. Some notes on the "workflow" to subscribe to these new events:
- Manually copy the method signatures of `LIF_CALLBACK_LISTENER` into a listener class within an abapGit repository
- Register the listener in the .abapgit.xml attribute `CALLBACK_CLASSNAME`
- Implement the callback methods

Side effects that might need discussion:
- For the old / new version parameters in the pull callback I introduced a version attribute to .abapgit.xml. I think this would be needed at some point in the future anyways and the alternative would be comparing commit hashes since tags aren't there yet.

Example: https://github.com/flaiker/callback-example

Local classes should also work (with the full name), though I haven't tested that yet. I also noticed some edge cases, where the type descriptor cache still has an old version (callback class was just pulled and activated and the cl_abap_typedescr=>admin_tab still contains the old descriptor reference). Probably need to do the system-call oneself, don't know of any way to clear that cache without leaving and reentering abapGit.

Note that these events are _dev-system-only_ (I should add that to the documentation). To do something useful with them in a multi system landscape you need to generate transport requests in the callback method implementations (possibly with XPRA report execution).

### DIFF
--- a/src/zabapgit.prog.abap
+++ b/src/zabapgit.prog.abap
@@ -52,7 +52,6 @@ INCLUDE zabapgit_exit.
 INCLUDE zabapgit_stage.
 INCLUDE zabapgit_git_helpers.
 INCLUDE zabapgit_repo.
-INCLUDE zabapgit_callbacks.
 INCLUDE zabapgit_news.
 INCLUDE zabapgit_stage_logic.
 INCLUDE zabapgit_2fa.
@@ -62,6 +61,7 @@ INCLUDE zabapgit_objects.
 INCLUDE zabapgit_tadir.
 INCLUDE zabapgit_file_status.
 INCLUDE zabapgit_popups.
+INCLUDE zabapgit_callbacks.
 INCLUDE zabapgit_zip.
 INCLUDE zabapgit_objects_impl.
 
--- a/src/zabapgit.prog.abap
+++ b/src/zabapgit.prog.abap
@@ -86,6 +86,20 @@ INCLUDE zabapgit_migrations.          " Data migration routines
 INCLUDE zabapgit_forms.
 
 **********************************************************************
+LOAD-OF-PROGRAM.
+  DATA: gs_submit           TYPE zif_abapgit_definitions=>gty_callback_submit,
+        go_callback_adapter TYPE REF TO lcl_callback_adapter.
+
+  IMPORT submit = gs_submit FROM MEMORY ID 'AGT'.
+
+  IF sy-subrc = 0 AND gs_submit IS NOT INITIAL.
+    go_callback_adapter = lcl_callback_adapter=>get_instance(
+      io_repo = lcl_app=>repo_srv( )->get( gs_submit-repokey )
+      iv_use_submit = abap_false ).
+    go_callback_adapter->submit( iv_methname = gs_submit-callback iv_args = gs_submit-args ).
+    LEAVE PROGRAM.
+  ENDIF.
+
 INITIALIZATION.
   lcl_password_dialog=>on_screen_init( ).
 
--- a/src/zabapgit.prog.abap
+++ b/src/zabapgit.prog.abap
@@ -52,6 +52,7 @@ INCLUDE zabapgit_exit.
 INCLUDE zabapgit_stage.
 INCLUDE zabapgit_git_helpers.
 INCLUDE zabapgit_repo.
+INCLUDE zabapgit_callbacks.
 INCLUDE zabapgit_news.
 INCLUDE zabapgit_stage_logic.
 INCLUDE zabapgit_2fa.
--- a/src/zabapgit_callbacks.prog.abap
+++ b/src/zabapgit_callbacks.prog.abap
@@ -149,7 +149,7 @@ CLASS lcl_callback_adapter IMPLEMENTATION.
     ls_parameter-name = lc_parmname_new_version.
     ls_parameter-kind = cl_abap_objectdescr=>exporting.
     lr_new_version->* = iv_new_version.
-    ls_parameter-value = lr_old_version.
+    ls_parameter-value = lr_new_version.
     INSERT ls_parameter INTO TABLE lt_parameters.
 
     dyn_call_method( io_object     = mo_listener
--- a/src/zabapgit_callbacks.prog.abap
+++ b/src/zabapgit_callbacks.prog.abap
@@ -2,14 +2,44 @@
 *& Include zabapgit_callbacks
 *&---------------------------------------------------------------------*
 
+"! Callback interface for abapGit repository events
+"! <p>
+"! To subscribe to abapGit repository events create a preferrably global class in your repository
+"! and <em>copy</em> the method signatures of this interface. You only need to copy the ones you
+"! want to subscribe to. Specify the name of the class (full name for local classes) in the
+"! .abapgit.xml attribute <em>CALLBACK_CLASSNAME</em>.
+"! </p>
+"! <p>
+"! This interface is intentionally defined locally. If it were global users might be inclined to
+"! instead of copying the interface's methods to just implement it in their callback class. Since
+"! abapGit most of the time only exists in the development system this would cause a compiler error
+"! on import in QA / production systems.
+"! </p>
+"! <p>
+"! Version parameters are always taken from the <em>VERSION</em> attribute in .abapgit.xml.
+"! </p>
 INTERFACE lif_callback_listener.
   METHODS:
+    "! Pull event listener
+    "! <p>
+    "! Only works in online repositories. This method gets called after the activation of just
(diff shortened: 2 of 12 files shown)


larshp:
Hi,
A: I think this feature should work the same for online and offline projects, suggest replacing `on_after_install` and `on_after_pull` with one method: `on_after_deserialize`

B: For the popup, it looks like it is triggered each time a project is pulled? Perhaps we can add a method in the callback implementation which can signal if it needs to do some work or not, and display only the popup if it returns true.

C: In abapGit we have had multiple conversions of data without doing the conversion based on version number. In your example repository, it moves data from one table to another, ie. the version number is not required in order to do the conversion. I guess that most of the usages of the callback can be done without knowing a version number, I also guess callbacks will be easier to test and debug without involving the version number, this might leave a bit more work in the callback implementation.

flaiker:
A:
Makes sense to me, I'll change that.

B:
Yes, that's not ideal currently. I am not sure what the correct way would be to handle this. I can add another method for each callback `should_execute_...`. That kind of defeats the purpose of looking at the callback implementation before it's executed though because then that method will always be executed without possible user intervention. I don't know if that's relevant at all, but so far in abapGit I don't think there's a point where repository code will just get executed, so there might be some security concerns here.

C:
Without the version number parameter in the callback I guess the best approach would be for the repository to maintain a database table where migrations are marked as done. Because I cannot think of a way to find out the previous state of the repository in the callback otherwise. Note that branch switching might cause a repository to revert to a previous state even though migrations already happened. Also versions in between will be skipped, because there's no incremental pull. Which generally is fine, but something to keep in mind when implementing a callback method.
But I think the main use case for these would be "user installations" and not "development installations", therefore usually master will be just updated to a newer version, where this behavior is fine.
For "development installations" I thought of having an option in the repo settings to disable callbacks altogether and maybe make it possible to execute them manually.

larshp:
B: true, hmm. Perhaps we should add a option "Always run/trust" in the popup?

C: Yeah, it is difficult, especially anticipating what this will be used for, so I think we should leave everything for the application developer to decide, and just always execute the callback. So far I'm considering the following use cases:
* Data conversion, like in your example, the application developer can look in the new table, if it is empty then perform conversion
* Extra activation steps(eg gateway services), check if the service is inactive, if so then activate it
In both of these cases there is no need for information to be stored in abapGit itself, and if needed, the application developer can store information somewhere else.

flaiker:
I suggest adding a repository setting for "callback trust levels":
- Ask each time
- Always execute
- Never execute

I thought of adding it to the repo settings page. So far there are only settings in there that are saved to .abapgit.xml and not _local_ settings stored in the db table though. So I am not sure if it's the right place.

Maybe one can also recommend in the documentation to execute migrations asynchronously or inform / ask the user about what's happening in more dialog windows.

larshp:
repo settings page:
Sometime I would like to add the IGNORE_SUBPACKAGES to the settings page, https://github.com/larshp/abapGit/issues/952#issuecomment-333626062 which is also not stored in .abapgit.xml, perhaps have a different section for the local settings

flaiker:
Like this?
image

larshp:
Yeah, cool 👍

flaiker:
I rebased, added the setting and fixed some stuff for local classes.
One problem that still exists is the following:
- Pull changes
- `ON_AFTER_DESERIALIZE` is checked if it needs to be executed. There is no method `ON_AFTER_DESERIALIZE` in the registered callback class, which is checked using `cl_abap_classdescr=>methods`
- Pull changes again
- `ON_AFTER_DESERIALIZE`is checked if it needs to be executed. The callback class was just pulled as an updated version. It now has the method. `CL_ABAP_TYPEDESCR` has however cached the last classdescriptor object and its method-itab does not contain the new callback method. The callback does not get executed.

I looked for ways to clear the cache (`admin_tab`) but couldn't find anything. I also looked at replicating the `describe_by_...`-methods to circumvent the cache, but the whole API seems to be quite locked down. One solution would be to do the callback stuff in a new internal mode, but that seems very difficult to do with the copy paste installation. Another way might be to read the classpool / program include manually, invoke the compiler manually and analyze that abap load, but that seems very tedious as well. Or `SUBMIT` the abapGit report again to force the caches to be empty, but then all state of the GUI, logon cache etc. will be lost or needs to be ex- / imported.

Any ideas?

larshp:
hmm, no

larshp:
perhaps check the SEO* database tables instead of using `cl_abap_typedescr`

flaiker:
SEO* tables are a good idea. This will however prevent local classes from being used.
I'll experiment a little bit...

flaiker:
I fixed it using the SUBMIT approach. Please let me know if that's something you don't want.
It shouldn't interfere with database commits, as `AND RETURN` starts a new LUW according to the documentation.

larshp:
I think it is a pretty evil workaround, another idea: just call the callback class if it is specified, it is the authors responsibility that it works, implements the right interface, and does the right thing

flaiker:
I agree we could remove all the checks. However, that won't help with the old version being used. In my tests the callback class version being used is always the one that was active when it was first loaded. So if you update your repository twice without exiting abapGit there's no way I know of to use the newly generated ABAP Load instead of the old one without leaving the main program in some way (RFC, SUBMIT, CALL TRANSACTION). Which would be quite the deal breaker for the feature in my opinion, if you cannot ensure that the callback implementation is always the one that belongs to repository state that is active.

I am wondering: Does abapGit itself not have that problem too when you update abapGit using the repository based installation? Do you need to restart it?

I'll see if I can get a small reproducable example report working to test things in a more isolated way.

Another idea would be to generate a subroutine pool at runtime that contains a subclass of the repository callback class, no idea if then the active version will be used...

flaiker:
Subroutine pool doesn't work unfortunately, here's the small example:
https://github.com/flaiker/class-test/blob/master/zclasstest.prog.abap

larshp:
hmm 😕

larshp:
how about using plain old reports? one report to be run in background job after deserialize, and a different one before uninstall?

flaiker:
Plain old reports? With background jobs? That's madness!!!

I also took a look at OO transactions, because they can call global or local classes' methods in a new mode, but I think there are processed to deep in the kernel to use anything of them for this.

larshp:
but it will work? The SUBMIT solution is more or less the same, except it adds the complexity inside the ZABAPGIT report

larshp:
Okay, another workaround, implement callbacks via global classes, this raises the problem of cached code.

So, in the callback code we can keep track of which callbacks have been called and the timestamp for when the code was last updated. If the class has been updated and already executed, then show an error, requesting the user to restart abapGit?

flaiker:
Another idea:
I was looking for RFC enabled function modules that are available on minimal AS ABAP installations that execute methods in a new mode but of course there are none... However, with that I thought of the object test environment where you can click to execute method or instantiate objects etc. I think we may be able to use that.
Function module `SETS_FA_METHOD_INVOKE` is the one that does the actual method call. Around that there is a generated test frame. It will be a bit tricky to get the parameters to work with that (these are other FM or even forms that are called beforehand) but it should do it (could even to batch input on `SETS_CLASS_TEST_ENTRY` if all else fails). Since these callbacks are again dev system only I don't see a problem with using the test environment API.
What do you think?

larshp:
looks promising 😄

larshp:
hi, perhaps we should close this PR, and continue the work/comments in the issue?

flaiker:
Yes sure, don't want to block anything here.

PR #1082, opened 2018-01-02
implement ligthweight tags

christianguenter2:
First basic support for ligthweight tags. #1081 

~~The feature can be enabled with the 'Enable experimental features' switch.~~

It works quite similar to branches, although you cannot checkout tags yet.
image

Overview
image

˙
Create tag
image
image

Delete tag
image
image


larshp:
Looks good, few comments:

A: Can you add a comment in create_tag that it is a "lightweight" tag?

B: Somehow we should make it visible for the user that the tag is created from the current commit

C: I'm okay with not adding this as an experimental feature, it is a new feature, so it should not affect the current usage of abapGot

christianguenter2:
Good points;

A: I added an explaining comment 
B: I added a popup showing the SHA of the current commit
image

C: I removed the activation with the experimental feature flag

larshp:
Looks good to me, tested successfully, have a few comments which I will add in the isssue, lets get this merged 🚢

PR #1151, opened 2018-01-17
Use DD_MASS_ACT_C3 for DDIC activation

christianguenter2:
#1055 
Use FM DD_MASS_ACT_C3 for DDIC activation if available. Must be enabled with experimental feature flag, otherwise the current activation logic is used.

### DIFF
--- a/src/zabapgit_objects.prog.abap
+++ b/src/zabapgit_objects.prog.abap
@@ -45,8 +45,7 @@ CLASS lcl_objects_activation DEFINITION FINAL.
     CLASS-METHODS activate_ddic
       RAISING zcx_abapgit_exception.
 
-    CLASS-METHODS activate_non_ddic
-      RAISING zcx_abapgit_exception.
+
 
     CLASS-DATA: gt_objects TYPE TABLE OF dwinactiv.
 
@@ -208,7 +207,7 @@ CLASS lcl_objects_activation IMPLEMENTATION.
                                   iv_total   = '100'
                                   iv_text    = '...' ).
 
-      activate_non_ddic( ).
+      activate_old( ).
 
     ENDIF.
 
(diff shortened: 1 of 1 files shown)


src/zabapgit_objects.prog.abap
```diff
@@ -155,6 +171,198 @@ CLASS lcl_objects_activation IMPLEMENTATION.
...
+  METHOD activate_new.
+
+    DATA: lt_ddic     TYPE act_work_area,
+          lt_non_ddic TYPE act_work_area.
+
+    IF gt_objects IS INITIAL.
+      RETURN.
+    ENDIF.
+
+    zcl_abapgit_progress=>show( iv_key     = 'Activate'
```
larshp: "Activate" -> "Activating DDIC"

src/zabapgit_objects.prog.abap
```diff
@@ -155,6 +171,198 @@ CLASS lcl_objects_activation IMPLEMENTATION.
...
+      IF sy-subrc <> 0.
+        zcx_abapgit_exception=>raise( 'error from RS_WORKING_OBJECTS_ACTIVATE' ).
+      ENDIF.
+
+    ENDIF.
+
+  ENDMETHOD.
+
+
+  METHOD filter_ddic_objects.
```
larshp: do we have some objects in abapGit that identifies themselves as DDIC, but are not really DDIC? Or is there a difference between the function modules?

christianguenter2: Oh, I see. We have already the distinction between DDIC and non-DDIC, so filter_ddic_objects is superfluous. Will change the code.

src/zabapgit_objects.prog.abap
christianguenter2: changed

christianguenter2:
Changed the logic, so that the distinction between DDIC and non-DDIC of abapGit ist used

src/zabapgit_objects.prog.abap
```diff
@@ -155,6 +165,161 @@ CLASS lcl_objects_activation IMPLEMENTATION.
...
+      activate_ddic( ).
+
+    ELSE.
+
+      zcl_abapgit_progress=>show( iv_key     = 'Activating non DDIC'
+                                  iv_current = '98'
+                                  iv_total   = '100'
+                                  iv_text    = '...' ).
+
+      activate_non_ddic( ).
```
larshp: perhaps just call `activate_old( abap_false )`, then method `activate_non_ddic` can be deleted

christianguenter2: good point, changed...

christianguenter2:
first try displaying errors:
image

PR #1213, opened 2018-02-18
Improve error messages

flaiker:
Hi. I thought we might be able to improve the error messages that are generated to be more user friendly.

Before:
image

After:
image

I kind of backported [IF_T100_DYN_MSG](https://help.sap.com/http.svc/rc/abapdocu_752_index_htm/7.52/en-US/index.htm?file=abenif_t100_dyn_msg.htm) for this and replaced the errors in DEVC with the new approach as a start.

Not sure if this is something you categorically rule out, because I edited ZCX_ABAPGIT_EXCEPTION "by hand" in eclipse. SE24 still does not allow to edit the constructor of exception classes afaik. If so just close the PR.


larshp:
Thanks, will take a look later today.

I'm thinking it might be possible to do a few changes, and have the default constructor in the exception class

larshp:
Hi,
Can you try to check if this is possible to implement without changing the constructor manually? We just need some attributes, which are automatically added as input to the constructor when adding via SE24

Plus do you know if `if_t100_message` exists on 702?

flaiker:
Should be possible without changing the constructor, just need to add some more logic to the raise methods. Naming conventions will get broken though :P

I am not sure which release it was added. Do you know the releases by date?
image

larshp:
naming conventions: yes, they are also broken in the existing exception classes

release: I guess its okay, lets just try

flaiker:
Sorry for the delay...

Constructor is now the generated one. Though I wasn't able to generate it using SE24 (I assume because of the pragma, but that cannot be removed in either eclipse nor SE24 o0).
So ADT regenerated the constructor with the quick fix thing.

larshp:
no problem, looks good, merging

PR #1287, opened 2018-03-30
ECTD: ECATT test data container serialization

christianguenter2:
#105 
First try implementing ECATT test data container. 

Here's an example.
https://gitlab.com/christianguenter/test_ecatt/blob/master/z_ecatt_test_data.ectd.xml

Not quite sure whether its a good approach how I handle the versions. Maybe it needs some refinements. For sure we need more testing of real life objects.

Feedback welcome.

### DIFF
--- a/src/objects/zcl_abapgit_object_ectd.clas.abap
+++ b/src/objects/zcl_abapgit_object_ectd.clas.abap
@@ -1,0 +1,355 @@
+CLASS zcl_abapgit_object_ectd DEFINITION PUBLIC INHERITING FROM zcl_abapgit_objects_super FINAL.
+
+  PUBLIC SECTION.
+    INTERFACES zif_abapgit_object.
+    ALIASES mo_files FOR zif_abapgit_object~mo_files.
+    METHODS:
+      constructor
+        IMPORTING
+          is_item     TYPE zif_abapgit_definitions=>ty_item
+          iv_language TYPE spras.
+
+  PRIVATE SECTION.
+    CONSTANTS:
+      co_default_version TYPE etobj_ver VALUE '1' ##NO_TEXT,
+      BEGIN OF co_name,
+        version        TYPE string VALUE 'VERSION' ##NO_TEXT,
+        versions       TYPE string VALUE 'VERSIONS' ##NO_TEXT,
+      END OF co_name.
+
+    DATA:
+      mv_object_name TYPE etobj_name.
+
+    METHODS:
+      clear_attributes
+        CHANGING
+          ci_document TYPE REF TO if_ixml_document,
+
+      clear_elements
+        CHANGING
+          ci_document TYPE REF TO if_ixml_document,
+
+      get_version_from_node
+        IMPORTING
+          io_node           TYPE REF TO if_ixml_node
+        RETURNING
+          VALUE(rv_version) TYPE string,
+
+      deserialize_version
+        IMPORTING
+          io_version_node TYPE REF TO if_ixml_node
+          iv_package      TYPE devclass
+        RAISING
+          zcx_abapgit_exception,
+
+      serialize_version
+        IMPORTING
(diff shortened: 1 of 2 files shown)


larshp:
Hi,
I got the following diff after pulling the ECTD the first time:
https://github.com/larshp/DOMA/commit/b2ebe1f192eb81ca878c5c9ed9acf7c4952b711d
for variant BLAH,
image

It might be a bug in SAP standard, as ECATTDEFAULT should be SORTLNR = 1?

src/objects/zcl_abapgit_object_ectd.clas.abap
```diff
@@ -0,0 +1,355 @@
...
+    lo_node = li_document->create_element( co_name-version ).
+    lo_node->append_child( li_document->get_root_element( ) ).
+
+    co_versions_node->append_child( lo_node ).
+
+  ENDMETHOD.
+
+
+  METHOD zif_abapgit_object~changed_by.
+    rv_user = c_user_unknown.
```
larshp: use ETOBJ_GNDT-LUSER for largest ETOBJ_GNDT-LDATE + ETOBJ_GNDT-LTIME ?

src/objects/zcl_abapgit_object_ectd.clas.abap
```diff
@@ -209,7 +246,51 @@ CLASS zcl_abapgit_object_ectd IMPLEMENTATION.
...
+            im_obj_type      = ms_item-obj_type
+          IMPORTING
+            ex_version_info  = lt_version_info  ).
+
+        LOOP AT lt_version_info ASSIGNING <ls_version_info>.
+
+          ls_current_changed = get_change_information( <ls_version_info> ).
+
+          IF    ls_current_changed-ldate > ls_last_changed-ldate
+            AND ls_current_changed-ltime > ls_last_changed-ltime.
```
larshp: OR ldate = ldate AND ltime > ltime

christianguenter2:
> It might be a bug in SAP standard, as ECATTDEFAULT should be SORTLNR = 1?

Not quite sure how we should handle that. For me it seems like an idionsyncrasy in SAP standard too.

src/objects/zcl_abapgit_object_ectd.clas.abap
```diff
@@ -267,8 +267,8 @@ CLASS zcl_abapgit_object_ectd IMPLEMENTATION.
 
           ls_current_changed = get_change_information( <ls_version_info> ).
 
-          IF    ls_current_changed-ldate > ls_last_changed-ldate
-            AND ls_current_changed-ltime > ls_last_changed-ltime.
+          IF     ls_current_changed-ldate >= ls_last_changed-ldate
+             AND ls_current_changed-ltime >  ls_last_changed-ltime.
```
larshp: I think this wont work, in case `3 > 2 and 10 > 20` it will return false. Need something like 
`IF ldate > ldate OR ( ldate = ldate and ltime > ltime )`

christianguenter2: Oh, I see. I'll fix it tomorrow.

christianguenter2:
@larshp do you know why the CI checks fail here?
https://github.com/christianguenter2/abapGit/blob/5deb78bd4437c350acb7d6145285401bcddb0732/src/objects/zcl_abapgit_object_ectd.clas.abap#L374-L379

larshp:
the statement

```abap
          IF is_change_more_recent_than( is_currently_changed = ls_currently_changed
                                         is_last_changed      = ls_last_changed ).
```

does not work on 702, as it is not a comparison. Hence the parser error

christianguenter2:
I'd like to add the serialization for other ECATT objects ECAT and ECTC. Should I add it to this PR or open a new after we merged this one?

PR #1325, opened 2018-04-23
User specific abapGit settings

christianguenter2:
#1316 

Rearrange the settings page. There are now two separated sections "Global settings" and "User specific settings"

image

User specific settings are stored in the user specific config.
image

And removed from gloabl storage.
image

There's also a migration mechanism, migrating existing list size and ADT jump settings to personnel settings of all abapGit users.


larshp:
code looks good, I'll test it sometime during the weekend

larshp:
sorry, I have not had much time, does anyone else have time to do a short test?

larshp:
Hi,
In case there are no global settings, ie. the default are always used, abapGit cannot be started after applying the change,
image
image

christianguenter2:
fixed

PR #1389, opened 2018-05-13
Log: show ALV popup

christianguenter2:
#703 
Use ALV popup instead of POPUP_TO_INFORM

old:
image

new:
image


christianguenter2:
fixed and sqashed.

PR #1407, opened 2018-05-20
Possible solution to zabapgit merge functionality to resolve conflicts

WolfTreak:
Hello together,

I found this open project and was amazed. As abap developer I gave up the idea of good version management. With this project this could be changed greatly.
But to use it to full it would be cool to have a merge mechanism in zabapgit as I have no opportunity to use other git clients where I would like to use this report.

So I would like to provide my first attempt to the merge functionality if it would be a possible solution.
Please look at it and tell me your opinions.

Resolving conflicts would be solved by using a Diff-Page like ResolveConflicts-Page where every conflict of the merge result would be iterated and the user can see the source file, the differences to the target file and on the right side he gets an textarea with default the target file content and can then manually migrate the content of the source part. With apply this content will be converted to xstring and then put to the stage list and a sha1-key will be generated. After all conflicts user will be see the merge page again and now every file has an sha1-key and no conflicts anymore so he can merge now the branches.

Open points:
- css way to resize the textarea
- unified div view for source/target so more space to see the merged textarea
- sha1-key generation of new content should be adjusted to normal sha1-generation of the content -> type was unknown
- Change Diff determination from local/remote to correct source/target-branch, currently the same diff determination will be used as local and remote, so you get an deleted or added row where just updated files would be

#851 Fix nothing to merge bug, because time sort in zcl_abapgit_merge by ancestors was incorrect
#797 Possible solution with textarea

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_merge_res.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_merge_res.clas.abap
@@ -133,11 +133,14 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_MERGE_RES IMPLEMENTATION.
     LOOP AT it_postdata ASSIGNING <postdata_line>.
       lv_string = |{ lv_string }{ <postdata_line> }|.
     ENDLOOP.
-    REPLACE ALL OCCURRENCES OF zif_abapgit_definitions=>gc_crlf IN lv_string WITH lc_replace.
+    REPLACE ALL OCCURRENCES OF zif_abapgit_definitions=>gc_crlf    IN lv_string WITH lc_replace.
+    REPLACE ALL OCCURRENCES OF zif_abapgit_definitions=>gc_newline IN lv_string WITH lc_replace.
+
     lt_fields = zcl_abapgit_html_action_utils=>parse_fields_upper_case_name( lv_string ).
     zcl_abapgit_html_action_utils=>get_field( EXPORTING name = 'MERGE_CONTENT'
                                                         it = lt_fields
                                               CHANGING cv = filedata ).
+    filedata-merge_content = cl_http_utility=>unescape_url( escaped = filedata-merge_content ).
     REPLACE ALL OCCURRENCES OF lc_replace IN filedata-merge_content WITH zif_abapgit_definitions=>gc_newline.
 
     lv_new_file_content = zcl_abapgit_convert=>string_to_xstring_utf8( iv_string = filedata-merge_content ).
@@ -314,8 +317,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_MERGE_RES IMPLEMENTATION.
         ro_html->add( |<form id="merge_form" class="aligned-form" accept-charset="UTF-8"| ).
         ro_html->add( |method="post" action="sapevent:apply_merge">| ).
         ro_html->add( |<textarea id="merge_content" name="merge_content" | ).
-        ro_html->add( |htmlEsacpe="false" rows="{ lines( is_diff-o_diff->get( ) ) }">| ).
-        ro_html->add( |{ lv_target_content }</textarea>| ).
+        ro_html->add( |rows="{ lines( is_diff-o_diff->get( ) ) }">{ lv_target_content }</textarea>| ).
         ro_html->add( '<input type="submit" class="hidden-submit">' ).
         ro_html->add( '</form>' ).                          "#EC NOTEXT
         ro_html->add( '</div>' ).                           "#EC NOTEXT


larshp:
Hi,
Welcome to the abapGit community, check http://dotabap.org for abapgit based projects, if you are located near Germany there is a meetup late next month https://wiki.scn.sap.com/wiki/display/events/abapGit+Community+Meetup, also feel free to join the slack channel, https://abapgit-slackinviter.herokuapp.com

Thanks for the PR, will take a look, also see the errors in the travis build: https://travis-ci.org/larshp/abapGit/builds/381402253

```
src/ui/zcl_abapgit_gui_page_merge.clas.abap[7, 1]        - Bad indentation
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[152, 1]  - Reduce line length
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[315, 1]  - Reduce line length
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[320, 1]  - Reduce line length
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[321, 1]  - Reduce line length
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[205, 5]  - EXIT or CHECK outside of loop
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[479, 5]  - EXIT or CHECK outside of loop
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[484, 5]  - EXIT or CHECK outside of loop
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[499, 5]  - EXIT or CHECK outside of loop
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[135, 5]  - Reorder definitions to top of routine
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[314, 7]  - Reorder definitions to top of routine
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[486, 5]  - Reorder definitions to top of routine
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[542, 11] - Reorder definitions to top of routine
src/ui/zcl_abapgit_gui_page_merge_res.clas.abap[7, 1]    - Bad indentation
```

src/zcl_abapgit_merge.clas.abap
```diff
@@ -231,9 +231,11 @@ CLASS ZCL_ABAPGIT_MERGE IMPLEMENTATION.
       <ls_ancestor>-commit = lv_commit.
       <ls_ancestor>-tree = ls_commit-tree.
       <ls_ancestor>-body = ls_commit-body.
-      FIND REGEX zif_abapgit_definitions=>gc_author_regex IN ls_commit-author
-        SUBMATCHES <ls_ancestor>-time ##NO_TEXT.
-      ASSERT sy-subrc = 0.
+      <ls_ancestor>-time = ls_commit-author.
+
+      "Authorkommentar um alles entfernen, bis auf die Zeitangabe
+      REPLACE ALL OCCURRENCES OF REGEX '[a-zA-Z<>@.-]*' IN <ls_ancestor>-time WITH ''.
```
larshp: English comments please

jfilak:
Good job!

I cannot comment the coding, so here are my 2cents for UX.

I propose to have all 3 panels (Source, Target, Merge) looking the same and place the Merge panel into the middle. So, you either take the version from the left side or from the right side.

I would also swap the Source (<branch name>) panel and the Target (HEAD or <branch name>) panel because the left side is usually understood as history (you know timelines) and the right side is future. Even in programming you usually assign a right side value to a left side variable.

Also a button saying "use this version completely" for both the Source and the Target might be useful.

christianguenter2:
Small bug. If I resolve a conflict and then press again the 'Resolve Conflicts' button I get a shortdump.

image

image

christianguenter2:
When I do the following conflict resolution, the source is corrupted. It seems that the static method invocation operator => isn't escaped properly.

image


image

WolfTreak:
Thx for your comments, I will revisit the code and fix the points, change a little the layout and so on
I will make a further commit if I have a new version :-)

WolfTreak:
Hello together,

here is now the new version of this pull request.

If you click resolve conflicts you get at first the possible choice to select one of the branch versions(click apply next to branch name). If you dont want to select one and want to merge the solutions then you can click above on toggle merge mode and get to the next picture.

selectionmode

Here you can then merge the content in the textarea and click on apply. After that you get back to the merge page and with click on merge you can commit your merging.

mergemode

Now should also be the escaping working correct.

Please test it if it works like desired.

Best regards

WolfTreak

christianguenter2:
Given the following scenario. I have two branches master and feature. I'm on branch master and merge feature into master.

image

I apply the source of the feature branch.
image

Merge, commit and refresh. The merge was successful in the remote repository. But the local master branch is not in sync with its remote. And I cannot pull.
image

image

Shouldn't be the local checked out branch be in sync with its remote tracking branch after another branch was merged into?

WolfTreak:
So I understood the current merge functionality, is that if you merge something into another branch you make a new commit for that branch with the changes and there is no connections to the local source code. 

So you have the opportunity to be in branch bugfix maybe and merge branch feature with master without the dependency to actual be in master or in feature.

But that implies that you have an online repo where you can commit your merge otherwise this wouldn't work.

I have also seen that "bug" that you can't pull if you merged something in master and after that you will be in master. There you can only see the difference and click "Reset local" if you want to reset your local coding.

But that is in my opinion already a bug in the current "productive" code and should be correct sometimes. I think somewhere the application thinks because of the merge that the current local commit is the merge commit and doesn't allow you to pull. But in my opinion that is currently also so if you change local the source code you can't pull, only stage and if you want to reset you must push "Reset local".

larshp:
Hi,
I agree with the comments above, but we might need to take a look at this in the future.

I'm actually okay with merging this PR, I've found some bugs/comments but they are all minor. We can take these in follow up PRs, this PR is getting large, and #1416 is also on the way in.

🚢

PR #1534, opened 2018-06-28
delete empty packages

christianguenter2:
#1533

### DIFF
--- a/src/objects/zcl_abapgit_object_devc.clas.abap
+++ b/src/objects/zcl_abapgit_object_devc.clas.abap
@@ -20,7 +20,16 @@ CLASS zcl_abapgit_object_devc DEFINITION PUBLIC
                          RAISING   zcx_abapgit_exception,
       set_lock IMPORTING ii_package TYPE REF TO if_package
                          iv_lock    TYPE abap_bool
-               RAISING   zcx_abapgit_exception.
+               RAISING   zcx_abapgit_exception,
+      is_empty
+        IMPORTING iv_package_name    TYPE devclass
+        RETURNING VALUE(rv_is_empty) TYPE abap_bool
+        RAISING   zcx_abapgit_exception,
+      load_package
+        IMPORTING iv_package_name   TYPE devclass
+        RETURNING VALUE(ri_package) TYPE REF TO if_package
+        RAISING   zcx_abapgit_exception.
+
     DATA:
       mv_local_devclass TYPE devclass.
 ENDCLASS.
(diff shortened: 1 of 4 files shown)


larshp:
I am worried about the performance of GET_ELEMENTS

image

It does instantiation of each object.

Suggest replacing with an existence check on TADIR,
```abap
SELECT SINGLE * FROM tadir INTO ls_tadir WHERE devclass = package_name.
rv_empty = boolc( ... ).
```

Perhaps we should add a note somewhere regarding the SAP standard behavior for transported packages, they can only be deleted if no rows exists in TADIR. I.e. deletions of objects must be transported before the package can be deleted.

src/objects/zcl_abapgit_object_devc.clas.abap
```diff
@@ -558,4 +621,42 @@ CLASS zcl_abapgit_object_devc IMPLEMENTATION.
...
+
+  METHOD is_empty.
+
+    DATA: lv_object_name TYPE tadir-obj_name.
+
+    SELECT SINGLE obj_name
+           FROM tadir
+           INTO lv_object_name
+           WHERE obj_name <> iv_package_name
+           AND   devclass = iv_package_name.
```
larshp: I guess this will break if objects in the package are named the same as the package?

christianguenter2: you're right. Fixed it.

src/objects/zcl_abapgit_object_devc.clas.abap
```diff
@@ -629,8 +629,10 @@ CLASS zcl_abapgit_object_devc IMPLEMENTATION.
     SELECT SINGLE obj_name
            FROM tadir
            INTO lv_object_name
-           WHERE obj_name <> iv_package_name
-           AND   devclass = iv_package_name.
+           WHERE pgmid    =  'R3TR'
+           AND   object   <> 'DEVC'
+           AND   obj_name <> iv_package_name
+           AND   devclass =  iv_package_name.
```
larshp: If package name = ZNAME and it contains DOMA ZNAME, this will return true, as if the package is empty? Or perhaps I'm just too tired today

christianguenter2: hm, you are probably right. I really shouldnt't watch soccer and code... I'll fix it sometime later.

larshp: I dont remember how subpackages are stored, this is another test case

christianguenter2:
@larshp next try

PR #1568, opened 2018-07-06
Performance Improvement in X_TO_BITBYTE Conversion

mkaesemann:
Since X_TO_BITBYTE only converts a single byte to a bitstring, it is faster to use direct addressing instead of a loop/concatenate pair, since concatenate on characters has to process the whole char array in each run.


christianguenter2:
@mkaesemann this is quite an improvement! Thank you

PR #1619, opened 2018-07-11
 zcl_abapgit_factory naming convention #1132

JohannesKonings:
https://github.com/larshp/abapGit/issues/1132


christianguenter2:
why do we use g prefix instead of m?

JohannesKonings:
It was a finding in http://docs.abapopenchecks.org/checks/69/
It differentiates between instance and static attribute.

PR #1620, opened 2018-07-11
implement .gitignore

christianguenter2:
#1549 

- .gitignore is stored in the repo xml like this
image

- maintenance via database util or local git client

- currently supported:
-- root /
-- subfolders /sub/
-- star *

- .gitignore has to be removed from .abapgit.xml

- more to implement see https://www.atlassian.com/git/tutorials/saving-changes/gitignore

- demo [repo](https://github.com/abapGit-tests/gitignore.git)


larshp:
Is it needed in the repo xml? It will always be fetched to memory anyhow?

christianguenter2:
Currently it's possible to change and commit .gitignore from abapGit. Therefore it is needed to store it somewhere, isn't it?

larshp:
oh, okay, any thoughts on .abapgit.xml vs .gitignore? perhaps we should add some documentation on this

christianguenter2:
I think in the long run we should move the whole ignore stuff from .abapgit.xml to .gitignore. This would be more intuitive for new users as it mirrors the standard way of git. I'm sure we can provide migration mechanism for that if we decide that's the way we want to go. 
But first I'd like to ensure that the new .gitignore feature works as expected and then migrate.

christianguenter2:
Maybe we should also add a proper interface to maintain .gitignore as maintaining via database util is a bit tedious and error prone.

JohannesKonings:
I've done a little test:
 
Number 2 is ignored
image
image

after deletion of  number 2 in .gitignore and a pull, number 2 is shown

https://github.com/abapGit-tests/gitignore/commit/5abb21f1d406165965bbc0efb362a13061493010
image
image

christianguenter2:
Usecases would be e.g. ignore whole subpackage or not supported object types so that they don't show up always as not supported. 
In one project we had also errorneous objects (they were somehow corrupted) which lead to failures during serialization. They could be added to .gitignore as well.

christianguenter2:
> Any thoughts on the staging page, it says "ignore" but this is only for .abapgit.xml

hm, not quite sure. Never used this feature.

larshp:
I have not read the code, but the contents of .gitignore I assume is applied after serialization? So non-supported objects will show in the warnings list no matter how the .gitignore file looks

I assume files specified in .gitignore is not part of the git repository.

christianguenter2:
> I have not read the code, but the contents of .gitignore I assume is applied after serialization? So non-supported objects will show in the warnings list no matter how the .gitignore file looks

You're right, so that's no valid use case for .gitignore. We can still use the userexit to handle this.

larshp:
I'm still not sure about the use case for this, and it adds complexity to some of the central objects

PR #1652, opened 2018-07-18
AVAS: report detailed errors

filak-sap:
Unfortunately, we are versioning AVAS objects that can be instantiated
only in SAP systems but we use the repo also in sandbox Customer
systems.

Whenever I open the repository I get an exception of the type
cx_pak_invalid_data.

Without this commit I see the error message:
"AVAS error".

With this commit I see the error message:
"AVAS $GUID: invalid data: The functionality is not supported in
customer systems"

### DIFF
--- a/src/objects/zcl_abapgit_object_avas.clas.abap
+++ b/src/objects/zcl_abapgit_object_avas.clas.abap
@@ -88,8 +88,8 @@ CLASS ZCL_ABAPGIT_OBJECT_AVAS IMPLEMENTATION.
 
   METHOD instantiate.
 
-    DATA: lv_id TYPE guid_32.
-
+    DATA: lv_id TYPE guid_32,
+          lo_err TYPE REF TO cx_root.
 
     lv_id = ms_item-obj_name.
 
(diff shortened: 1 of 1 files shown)


jfilak:
Now, the question is how to work around the problem that I cannot use the repository with SAP AVAS attributes in our customer sandbox systems :)

christianguenter2:
Does this work? Instantiate doesn't propagate the exception zcx_abapgit_exception in its interface.
https://github.com/larshp/abapGit/blob/b98db9bd7c91f57b618e98b1ca30c8d2a89d0ce3/src/objects/zcl_abapgit_object_avas.clas.abap#L32-L34

jfilak:
> Does this work?

Yes, you get a short dump and the exception message is displayed in the detailed information. At least on our machines.

christianguenter2:
Hm, ok I see.
Wouldn't it be better if we propagate the exception to the caller? I get warnings for this in the code inspector and ATC. 

image

jfilak:
The ATC warning is bad, however, the actual behavior (dump) is ok. What about to use an ancestor of CX_DYNAMIC_CHECK?

One can propose to handle the exception, report the corresponding object as unsupported and keep abapGit running. It might be the solution, but I am not sure if abapGit is ready for such a logic. I would also want to mute the warning - perhaps, by putting the object to ignore list.

jfilak:
I should stop talking and start coding like @larshp -> #1672

larshp:
🚢

PR #1724, opened 2018-07-30
Added Buffering to ZCL_ABAPGIT_SAP_PACKAGE

mkaesemann:
Added Buffering to ZCL_ABAPGIT_SAP_PACKAGE

- Added a buffering option to ZCL_ABAPGIT_SAP_PACKAGE=>LIST_SUBPACKAGES, which gives a performance increase in repos with a deep tree structure, because all relevant packages are read from the DB only once and then the determination of subpackages is performed on an internal buffer table.
- Used buffered SAP package information determination in file status processing

### DIFF
--- a/src/zcl_abapgit_file_status.clas.abap
+++ b/src/zcl_abapgit_file_status.clas.abap
@@ -243,7 +243,7 @@ CLASS ZCL_ABAPGIT_FILE_STATUS IMPLEMENTATION.
 
       IF NOT ls_item-devclass IS INITIAL AND iv_devclass <> ls_item-devclass.
         IF lv_sub_fetched = abap_false.
-          lt_sub_packages = zcl_abapgit_factory=>get_sap_package( iv_devclass )->list_subpackages( ).
+          lt_sub_packages = zcl_abapgit_factory=>get_sap_package( iv_devclass )->list_subpackages( abap_true ).
           lv_sub_fetched = abap_true.
           SORT lt_sub_packages BY table_line. "Optimize Read Access
         ENDIF.
--- a/src/zcl_abapgit_file_status.clas.abap
+++ b/src/zcl_abapgit_file_status.clas.abap
@@ -243,7 +243,7 @@ CLASS ZCL_ABAPGIT_FILE_STATUS IMPLEMENTATION.
 
       IF NOT ls_item-devclass IS INITIAL AND iv_devclass <> ls_item-devclass.
         IF lv_sub_fetched = abap_false.
-          lt_sub_packages = zcl_abapgit_factory=>get_sap_package( iv_devclass )->list_subpackages( abap_true ).
+          lt_sub_packages = zcl_abapgit_factory=>get_sap_package( iv_devclass )->list_subpackages( ).
           lv_sub_fetched = abap_true.
           SORT lt_sub_packages BY table_line. "Optimize Read Access
         ENDIF.
--- a/src/zcl_abapgit_sap_package.clas.abap
+++ b/src/zcl_abapgit_sap_package.clas.abap
@@ -11,11 +11,21 @@ CLASS zcl_abapgit_sap_package DEFINITION
 
   PRIVATE SECTION.
     DATA: mv_package TYPE devclass.
+    DATA mv_buffered TYPE abap_bool VALUE abap_false ##NO_TEXT.
+    DATA mt_buffer TYPE zif_abapgit_sap_package=>ty_devclass_info_tt .
 
     ALIASES:
       create FOR zif_abapgit_sap_package~create,
       create_local FOR zif_abapgit_sap_package~create_local.
 
+    METHODS get_package_info
+      IMPORTING
+        !i_no_sap_packages           TYPE abap_bool DEFAULT abap_true
+      RETURNING
+        VALUE(rt_packages_in_system) TYPE zif_abapgit_sap_package=>ty_devclass_info_tt .
+    METHODS set_buffer
+      IMPORTING
+        !it_buffer TYPE zif_abapgit_sap_package=>ty_devclass_info_tt OPTIONAL .
 ENDCLASS.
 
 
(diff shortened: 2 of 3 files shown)


larshp:
Hi,
A: SAP packages are allowed in abapGit if the system is a SAP system, see https://github.com/larshp/abapGit/blob/7ac8dc15b6d9e6770db62a744881d547b2590fd8/src/zcl_abapgit_repo_srv.clas.abap#L89, it looks like `i_no_sap_packages` is not supplied anywhere

B: Types defined in the interface can be moved to protected/private in the class instead

C: If doing development in Z packages, and a repository covers most of the Z packages in the system, this development will increase performance. However I am worried that too much unnecessary information is loaded into memory for small repositories, any performance measurements for small repos?

D: In the current implementation there is a SELECT for each subpackage, another approach could be to do a SELECT FOR ALL ENTRIES for each sub-level in the package hierarchy, perhaps this will scale linear and not have the constant penalty for small repos.

larshp:
Try the following,

```abap
REPORT ztdevc_performance.

PARAMETERS: p_devc TYPE devclass OBLIGATORY.

START-OF-SELECTION.
  PERFORM run.

FORM run.

  DATA: t1 TYPE i,
        t2 TYPE i.

  DO 2 TIMES.
    GET RUN TIME FIELD t1.
    PERFORM single.
    GET RUN TIME FIELD t2.
    t1 = t2 - t1.
    WRITE: / 'single,', t1.

    GET RUN TIME FIELD t1.
    PERFORM level.
    GET RUN TIME FIELD t2.
    t1 = t2 - t1.
    WRITE: / 'level,', t1.

    GET RUN TIME FIELD t1.
    PERFORM all.
    GET RUN TIME FIELD t2.
    t1 = t2 - t1.
    WRITE: / 'all,', t1.

    WRITE: /.
  ENDDO.

ENDFORM.

FORM single.

  SELECT devclass FROM tdevc INTO TABLE @DATA(lt_sub) WHERE parentcl = @p_devc.

  LOOP AT lt_sub INTO DATA(ls_sub).
    SELECT devclass FROM tdevc APPENDING TABLE @lt_sub WHERE parentcl = @ls_sub-devclass.
  ENDLOOP.

  WRITE: / lines( lt_sub ).

ENDFORM.

FORM level.

  SELECT devclass FROM tdevc INTO TABLE @DATA(lt_level) WHERE parentcl = @p_devc.

  DATA(lt_result) = lt_level.

  WHILE lines( lt_level ) > 0.

    SELECT devclass FROM tdevc INTO TABLE @lt_level FOR ALL ENTRIES IN @lt_level WHERE parentcl = @lt_level-devclass.
    APPEND LINES OF lt_level TO lt_result.

  ENDWHILE.

  WRITE: / lines( lt_result ).

ENDFORM.

FORM all.

  TYPES:
    BEGIN OF ty_devclass_info,
      devclass  TYPE devclass,
      namespace TYPE namespace,
      parentcl  TYPE parentcl,
    END OF ty_devclass_info .
  TYPES:
    ty_devclass_info_tt TYPE SORTED TABLE OF ty_devclass_info
      WITH UNIQUE KEY devclass
      WITH NON-UNIQUE SORTED KEY parent COMPONENTS parentcl .

  DATA: lt_tdevc TYPE ty_devclass_info_tt,
        lt_sub   TYPE STANDARD TABLE OF devclass WITH EMPTY KEY.


  SELECT devclass, namespace, parentcl
    FROM tdevc
    INTO CORRESPONDING FIELDS OF TABLE @lt_tdevc
*       WHERE ( devclass LIKE '$%'
*               OR devclass LIKE 'Y%'
*               OR devclass LIKE 'Z%'
*               OR devclass LIKE '/%' )
*         AND as4user <> 'SAP'
         .

  LOOP AT lt_tdevc ASSIGNING FIELD-SYMBOL(<ls_devc>) USING KEY parent WHERE parentcl = p_devc.
    APPEND <ls_devc>-devclass TO lt_sub.
  ENDLOOP.

  LOOP AT lt_sub INTO DATA(lv_sub).

    LOOP AT lt_tdevc ASSIGNING <ls_devc> USING KEY parent WHERE parentcl = lv_sub.
      APPEND <ls_devc>-devclass TO lt_sub.
    ENDLOOP.

  ENDLOOP.

  WRITE: / lines( lt_sub ).

ENDFORM.

*FORM find.
*
*  SELECT devclass FROM tdevc INTO TABLE @DATA(lt_top) WHERE parentcl = ''.
*
*  LOOP AT lt_top INTO DATA(lv_top).
*    p_devc = lv_top.
*    PERFORM level.
*  ENDLOOP.
*
*ENDFORM.
```

larshp:
For the $ABAPGIT package, it is around 40 times slower to load all rows from TDEVC into memory,
image

The sweet spot seems to be around 2000 packages in a package hierarchy(EA-HR),
image

I.e. if there is more than 2000 packages performance will be better, while everything less will be slower in the worst case scenario.

mkaesemann:
Regarding your fist comment:
**A**: `i_no_sap_packages` is not supplied, because I wanted to evaluate first if the restrictions are okay or not, so the information about allowing abapGit in a SAP system that you gave is actually exactly the kind of info needed to tune this. However, I did forget to remove the default value for the method that I have currently set to `ABAP_TRUE` for our repo for testing. 
**B**: Originally I re-used thy type in the `ZCL_ABAPGIT_FOLDER_LOGIC` class, but while splitting the PRs, I no longer saw the advantage in this and instead created a local type in the `ZCL_ABAPGIT_FOLDER_LOGIC` class. So you are correct that this can be removed from the interface as well. Which also has the advantage that the number of impacted classes/interfaces is reduced.
**C**: Even if all packages are loaded in an S/4 system (about 30000 in our development system; less for pure standard SAP), the amount of memory needed is only 4,2 MB, which does not pose a problem for a modern system. Especially since the performance gained from less DB accesses is very much noticeable.
It is true that the select is slower for repos with few packages, but it is still reasonably low, while the runtime for the single selects increases with the number of packages in the repo.
When measuring the pure DB select overhead, the sweet spot for the unoptimized access (i.e. all packages in the system is about 75-85 packages, in non-SAP systems once the selection restriction can be used it goes down to 2-3 packages (in a system with 28110 packages total). The block select takes a constant ~85 milliseconds and occupies 4,2MB worst case. So I thought the tradeoff was woth it 😇 
**D**: Doing the selection per level is also a valid option, but I would see it more as an option for the unbuffered case, because it will definetly be faster than the single select, but not as fast as the fully buffered version - and in my repo I need every millisecond I can get 😉 

I'm not sure how you got to the 2000 packages as a sweet spot. I'll try to have a look at your measurement program after my next meeting session 😒

mkaesemann:
I've run your test and the single selection is (as expected) no match for neither the level-based nor the block selection. For our repo (220 packages) the single selection is already slower than the block select.
The level-based selection is a good optimization.
Selecting with restriction to custom/partner packages still beats the level-based approach though, but doesn't work in SAP systems.
So I guess, a good solution could be to detect if the system is a SAP system and if so use the level-based approach, while in using the restricted block select in customer systems, where generally only a couple hundred to a thousand packages are likely to exist.
Should I try to incorporate both options into this PR or would it be better to create a PR for the level-based subpackage selection? What do you think?

mkaesemann:
On second thought, I think I will remove the buffering completely and change the PR into "level-based selection only". The runtime differences are minimal and we gain the advantage of a) not having to change anything in `ZIF_ABAPGIT_SAP_PACKAGE` nor any of the calling methods and b) never run a risk of encountering a system constellation, where packages are not found due to the restriction in the selected packages (this shouldn't happen, but its better to be safe than sorry).
I'll rework the PR and get back to you.

larshp:
Cool, that was also my next observation, the following package scenario delivered by SAP:
image

mkaesemann:
Ready for the next review 😅

PR #1835, opened 2018-08-17
Fix close hotkey overview popup #1833

christianguenter2:
#1833


larshp:
Also suggest having a comment in the popup that '?' can be used to close the popup, took me a while to figure it out 🐵

christianguenter2:
done
image

PR #1838, opened 2018-08-17
Delete BOM in XML #1044

JohannesKonings:
https://github.com/larshp/abapGit/issues/1044

After the XML has been created, the BOM will be deleted.
After this PR every repo has to push all XML again. Otherwise the XML will be displayed as modified.

### DIFF
--- a/src/objects/zcl_abapgit_object_msag.clas.abap
+++ b/src/objects/zcl_abapgit_object_msag.clas.abap
@@ -10,7 +10,16 @@ CLASS zcl_abapgit_object_msag DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
              msgnr TYPE t100-msgnr,
              text  TYPE t100-text,
            END OF ty_t100_texts,
-           tt_t100_texts TYPE STANDARD TABLE OF ty_t100_texts.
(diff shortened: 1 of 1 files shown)


larshp:
hmm, thought it was the other way around, that only files from non-unicode systems were modified

JohannesKonings:
XML creation adds the BOM on a unicode system, so all repos which are pushed from a unicode system are affected.
I tried to add a BOM on a non unicode system. But the BOM get lost in the conversion from xstring to string.

Maybe we can delete the BOM only for the diff check? Unfortunately I've found no good position to do that so far.

Any other suggestions?

JohannesKonings:
If I have not misunderstood the approach in https://github.com/larshp/abapGit/issues/799#issuecomment-335928278, the encoding utf-8 can not be set in a unicode system.
image

In a non unicode system it works
image

larshp:
okay, hack it is then 😄 

but I suggest collecting the logic https://github.com/larshp/abapGit/blob/d25bfaf725a9b9b8e256b010cb0eb7f9c8f589c7/src/objects/zcl_abapgit_objects_files.clas.abap#L155-L158 and the BOM in a single place

JohannesKonings:
Unintentionally closed through reorganisation of commits.I'll reopen if I had tested the changes.

PR #1915, opened 2018-09-10
Syntax error in ZCL_ABAPGIT_BRANCH_OVERVIEW (only on 731 SP7) #1914

bigld:
fix RETURNING with EXPORT/CHANGING <740 SP2


bigld:
#1914

Issue #1799, opened 2018-08-07
Dump CX_SY_RANGE_OUT_OF_BOUNDS in  ZCL_ABAPGIT_SYNTAX_HIGHLIGHTER


christianguenter2:
Steps to reproduce:

1) clone [demo](https://gitlab.com/christianguenter/test_diff_rendering.git)
2) click on view diff near file /src/z_test.wapa.fragment.xml

![image](https://user-images.githubusercontent.com/17437789/43795337-54b98660-9a81-11e8-90b7-b3c4290195fd.png)

![image](https://user-images.githubusercontent.com/17437789/43795205-ef37f7cc-9a80-11e8-980a-4a07c1eef204.png)
![image](https://user-images.githubusercontent.com/17437789/43795219-f7898ce2-9a80-11e8-805a-265b3f3cf160.png)
![image](https://user-images.githubusercontent.com/17437789/43795227-005dd9ae-9a81-11e8-89da-7be5dcd64a7c.png)

![image](https://user-images.githubusercontent.com/17437789/43795260-18589d5a-9a81-11e8-9e87-e19a52fd720d.png)

christianguenter2:
Line contains 
`text="{i18n>Menu}"  `
it seems that the > is the root cause.

christianguenter2:
It seems that this regex for xml tags isn't sufficient.
https://github.com/larshp/abapGit/blob/26c6f4b4dc9db279c905146bb781cc9f9be89740/src/syntax/zcl_abapgit_syntax_xml.clas.abap#L22

marcellourbani:
had a quick look but I get an empty sintax check page.
Perhaps I need to set up a default code inspector variant or something?

christianguenter2:
It has nothing to do with syntax check page or code inspector. Just open the diff of z_test_wapa.fragment.xml.

marcellourbani:
Dumps. I will pick this up if you're not working on it already

christianguenter2:
Cool, I'm not working on it.

marcellourbani:
It's not as simple as changing a regex

First of all, our regexes are without submatches, so if I try to escape away 
(?:"(?:(?:[^"]|(?:\\.))*)")?([<>])
would actually work, and return the first submatch as < or >

I will extend the engine to support submatches, which will remove the dump

If we also want to properly highlight the tags we would need to remember we left one opened the line above.

Will look into it and add that too, otherwise we'll end up with something like this:

[...]

PR #1923, opened 2018-09-14
fixed issue #1799 (dump for  < or > in attributes)

marcellourbani:
on top of a new regular expression I added support for submatches.

If relevant_submatch is not 0, a match is only reported if the relevant submatch has length > 0

Also, I added code to highlight XML tags which don't close on the line they open
I noticed a potential problem in attributes starting on a new line (regex starts with \s, so if not indented they're not recognised) but decided to leave those alone, worst case they won't be highlighted 

image


marcellourbani:
Didn't realize there were any, will do

PR #1932, opened 2018-09-18
Issue #1866 #790 hack for java sapgui

marcellourbani:
After a lot of troubleshooting I found out the issue is a bug in the embedded browser.

With either a dynamically added form or an existing one, after we do a 

`form.setAttribute("action", "sapevent:" + action);`

post breaks (get would still work, up to 2048 char of payload).

So the proposed solution is:

- embed a form with a known ID and dummy action in the page
- look it up with a getElementById (fall back to create if not found)
- add an extra hidden field with the actual action requested
- when the GUI recognizes the dummy action, it extracts it from postdata 
- passes down a cleaned postdata with the action requested in javascript 

PS: just realised I don't handle gets properly, will fix it ASAP, don't bother reviewing before that but please let me know if you think it's too cumbersome for the few java gui users

### DIFF
--- a/src/ui/zcl_abapgit_gui.clas.abap
+++ b/src/ui/zcl_abapgit_gui.clas.abap
@@ -412,12 +412,12 @@ CLASS zcl_abapgit_gui IMPLEMENTATION.
         ENDIF.
 
         "split rest in postdata format
-        WHILE rest <> ''.
+        WHILE pdata <> ''.
           APPEND INITIAL LINE TO postdata ASSIGNING <postdata>.
-          <postdata> = rest.
+          <postdata> = pdata.
           DESCRIBE FIELD <postdata> LENGTH pdatalen IN CHARACTER MODE.
-          IF strlen( rest ) >= pdatalen.
-            rest = rest+pdatalen.
+          IF strlen( pdata ) >= pdatalen.
+            pdata = pdata+pdatalen.
           ELSE.
             EXIT.
           ENDIF.
--- a/src/ui/zcl_abapgit_gui.clas.abap
+++ b/src/ui/zcl_abapgit_gui.clas.abap
@@ -362,7 +362,7 @@ CLASS zcl_abapgit_gui IMPLEMENTATION.
 
 
   METHOD transform_dummy_form_params.
-    " to support SAPGUI java we handle some postcalls a hardcoded action
+    " to support SAPGUI java we handle some post calls a hardcoded action
     " passing the actual action in a parameter
     " this code simulates the original call using it
     DATA: pdata           TYPE string,
@@ -378,7 +378,12 @@ CLASS zcl_abapgit_gui IMPLEMENTATION.
     FIELD-SYMBOLS: <postdata> LIKE LINE OF postdata.
 
     IF action = '__abapgit_dummy__'.
+      "convert postdata in a string, as the bit we look for might be split between lines
       CONCATENATE LINES OF postdata INTO pdata RESPECTING BLANKS.
+      "I expect a & separated list of varname=value, perhaps starting with a ?
+      " want to remove my own variable and preserve the rest
+      " my variable is called __ABAPGIT_ACTION, and might include URL-encoded parameters
+      "   which will look like %3Fsomething %3F is an encoded ?
       FIND REGEX '([&?]?)__ABAPGIT_ACTION=([^\s&\%]+)(?:(?:%3[fF])([^\s&]*))?(&?)' IN pdata
         MATCH OFFSET moff MATCH LENGTH mlen
         SUBMATCHES prefix newaction queryparameters suffix.
@@ -386,12 +391,15 @@ CLASS zcl_abapgit_gui IMPLEMENTATION.
       IF sy-subrc = 0.
         "match: I did use the static form
         action = newaction.
+        "the encoded form has a fixed action, getdata will always be blank.
+        "I want to pass down what was included in my action above
         getdata = cl_http_utility=>unescape_url( queryparameters ).
 
-        "for postdata, remove the query result (except perhaps the initial or final ?/&) and split the rest
+        "for postdata, remove the query result (except the initial or final ?/& if I have both)
         CLEAR postdata.
         rest = pdata+moff.
         rest = rest+mlen.
+        "if what is left is blanks, ignore them
         FIND REGEX '^\s*$' IN rest.
         IF sy-subrc = 0.
           pdata = pdata(moff).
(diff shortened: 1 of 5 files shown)


marcellourbani:
I did fix the get stuff. Would like someone to test it, perhaps on a mac or other non-linux box (could test myself on windows but didn't yet)

larshp:
Hmm, seems a bit dangerous and hack, but something like this is probably also needed if Java GUI has to be fixed.

You mention a restriction of 2048 bytes payload, where does this arise? We have pages with textareas for editing code, this data will be more than 2048 bytes.

marcellourbani:
The current implementation has no size limit AFAIK.
The first hack I tried was to detect java sapgui in JS and force a get (then put the getdata in postdata): worked but had this limitation of 2048 char which is the actual size of getdata.

AFAIK the current solution supports everything and is transparent, but the ABAP code to achieve transparency is an ugly hack

I personally like reusing an existing form more than adding one every time one presses a button (not a big deal anyway, I still do it for gets, as otherwise I would need either a 2nd static form or an even uglier hack for gets).

I think I can get this a bit cleaner (i.e. adding another static form for GETs and refactoring the ABAP code a bit to isolate the hack from the rest of the code).
But wanted to hear from you before: no reason to clean this up if it won't be merged anyway.

Was keen on doing it because I often work on a linux box and this issue was really bothering me.

larshp:
@christianguenter2 any thoughts on this PR?

christianguenter2:
Hm, I have mixed feelings about that. On the one hand we should fix things for SAPGui Java on the other hand the hack makes the code more complex. But if it works I would be ok with that. 

But it seems that the shortcuts aren't compatible with the hack. E.g. if I press `R` to refresh.
image

This should be fixed definitely.

marcellourbani:
indeed, will look into it

src/ui/zcl_abapgit_gui_page_repo_sett.clas.abap
```diff
@@ -242,11 +243,19 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_SETT IMPLEMENTATION.
...
 
+    lt_sorted_fields = it_post_fields.
+    sort lt_sorted_fields.
+
+    LOOP AT lt_sorted_fields INTO ls_post_field WHERE name CP 'req_*'.
+      IF ls_post_field-name+4(3) = 'com'.
+        INSERT INITIAL LINE INTO TABLE lt_requirements ASSIGNING <ls_requirement>.
+        <ls_requirement>-component = ls_post_field-value.
+      ENDIF.
+    ENDLOOP.
```
jfilak: Well, I am sorry but I have absolutely no idea why this commit exists :(

src/ui/zcl_abapgit_gui.clas.abap
```diff
@@ -278,12 +278,39 @@ CLASS ZCL_ABAPGIT_GUI IMPLEMENTATION.
...
+          lv_len TYPE i,
+          lv_temp TYPE string.
+
+    FIELD-SYMBOLS: <postdata> LIKE LINE OF lt_postdata.
+
+    lv_action = action.
+    lt_postdata = postdata.
+    "hack for java webgui support
+    IF action = '__abapgit_dummy__'.
+      LOOP AT lt_postdata ASSIGNING <postdata>.
```
jfilak: Perhaps, it might help some example - what is on input and what is the expected output. Such a documentation can be stored in commit long message or in a comment.

src/ui/zcl_abapgit_gui.clas.abap
```diff
@@ -345,4 +373,26 @@ CLASS ZCL_ABAPGIT_GUI IMPLEMENTATION.
     SET HANDLER me->on_event FOR mo_html_viewer.
 
   ENDMETHOD.                    "startup
+
+  METHOD remove_dummy_action.
```
jfilak: This the right place for documenting what the method does and why it even exists.

src/ui/zcl_abapgit_gui_page_repo_sett.clas.abap
marcellourbani: The previous code assumes it_post_fields comes in a certain order, in the java GUI with the JS fix comes down in a different one.
Might try to hack the form creation but instead I decided to copy and sort, then use 2 loops for creation of the table entries and setting the other fields.

I  don't particularly like this code but the alternatives I could think of were worse

src/ui/zcl_abapgit_gui.clas.abap
marcellourbani: yep good point, will do

marcellourbani:
@jfilak what breaks the form is setting its action.
Either we add a different form for every possible action or we work around it

jfilak:
What about this? https://github.com/filak-sap/abapGit/commit/067e1df5f98ef2ad6488416549456db1b2dd7eed

marcellourbani:
@christianguenter2 
I got to the bottom of this.
Current hotkey code uses a hack, sending an action like "repo_refresh?000000000011" as a post action, which ends up with a getdata of 000000000011
my hack sends this as a post parameter so it loses the getdata.

The easy way out would be to replace 

`submitSapeventForm({}, sUiSapEvent, "post");`
with 
`submitSapeventForm({}, sUiSapEvent, "get");`

As the parameters are hardcoded to empty, this should be harmless.
The alternative is an even uglier hack in the event handler, messing with getdata too

PS: the whole get query is restricted to ~2000 chars, if we want hotkeys for commit or similar we should review the solution

PS2: hotkeys don't work in webgui

christianguenter2:
Yes, everything is a hack :)

> The easy way out would be to replace
>
> submitSapeventForm({}, sUiSapEvent, "post");
> with
> submitSapeventForm({}, sUiSapEvent, "get");

I'm ok with that if it works.

PS: Regarding hotkeys I think we can ignore this at the moment.

PS2: Webgui? You mean SAPGui Java?

marcellourbani:
Yes sorry, sapgui Java.
Another issue I guess, unless I find a trivial fix

src/ui/zcl_abapgit_gui.clas.abap
marcellourbani: Ok, I refactored this heavily.
this method doesn't exist anymore, and the new transform_dummy_form_params is documented much better

marcellourbani:
@christianguenter2 get didn't work, I sorted it on both sides:

- in javascript, as this hack was required to handle parameters in post calls, I use the old method of creating a form then I have none
- in abap. I added code to set getdata based on the query attached to the action

I tested both, but until we add parameters in a hotkey call only the js fix will be in use

jfilak:
Please, have a look at the commit I added in my last comment. It fixes the java gui problem while it does not break other stuff.

marcellourbani:
edit: ZCL_ABAPGIT_GUI_PAGE_REPO_SETT already uses a static form anyway.
The dump was unrelated to my fix, and happens in the current code too

marcellourbani:
Ok

PR #1934, opened 2018-09-19
FORM: record transport request #1933

christianguenter2:
Before this commit is applied no transport request entry is recorded for SAPscript
form (object type FORM).

#1933


christianguenter2:
Test repo https://github.com/abapGit-tests/FORM

PR #1965, opened 2018-09-27
Object Type UDMO Data Model #1905

d023604:
Object Type UDMO Data Model #1905


d023604:
Reviewing again for the cause of the bad indentation error.

PR #2040, opened 2018-11-03
Code Inspector: Fix background processing

christianguenter2:
This commit enables the code inspector for background processing.
Before this commit is applied code inspection in background resulted in lock errors.

### DIFF
--- a/src/zcl_abapgit_code_inspector.clas.abap
+++ b/src/zcl_abapgit_code_inspector.clas.abap
@@ -44,10 +44,21 @@ CLASS zcl_abapgit_code_inspector DEFINITION
           zcx_abapgit_exception.
 
   PRIVATE SECTION.
+    CONSTANTS:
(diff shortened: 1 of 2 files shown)


larshp:
hmm, think I've used the old approach before, without lock errors

what kind of lock errors occured?

christianguenter2:
Also running in a background job? I get this error
image

created here
image

christianguenter2:
You can reproduce it with this code



```
PARAMETERS: package TYPE devclass.

TRY.
    DATA(li_syntax_check) = zcl_abapgit_factory=>get_syntax_check( package ).

    DATA(lt_list) = li_syntax_check->run( ).

    cl_demo_output=>display( lt_list ).

  CATCH zcx_abapgit_exception INTO DATA(lx_error).
    MESSAGE lx_error TYPE 'S' DISPLAY LIKE 'E'.
ENDTRY.
```

Issue #2036, opened 2018-11-02
SUSC: Deletion calls UI


christianguenter2:
![image](https://user-images.githubusercontent.com/17437789/47927414-8b307d80-dec3-11e8-8e01-799df0fe34c3.png)

demo repo
https://github.com/abapGit-tests/SUSC_and_SUSO.git

PR #2079, opened 2018-11-09
SUSC: delete in more direct way - no confirm popup #2036

bigld:
delete needs to be done similar to FM SUSR_DELETE_OBJECT_CLASS
I really hate this but there is no other way

There are not relevant changes in FM SUSR_DELETE_OBJECT_CLASS from 702SP14 to 740SP08

#2036

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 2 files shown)


bigld:
adjusted to more direct way and cleaner code

PR #2096, opened 2018-11-11
Refactor repo class, part1

sbcgua:
epigraph: beware entering the sacred lands ;)

**Why**: working on #1950 and respecting comments from #1953 I faced that current internal logic of repo classes is somewhat confusing and not transparent. Online and offline logic are a bit overlapping, many actions are triggered indirectly and planning changes is very error prone and require to hold a lot in the head. The idea here is to straighten the logic a bit (took some time ...).

Design ideology (IMHO):
1) repo_base should be an abstract container, though quite intelligent to understand that it may have remote files and able to handle them irrespective to the source (online/offline)
2) repo_base must know minimum about online logic. At the same time repo_online must not manage the state (mt_local/remote/state) itself, but just be responsible for fetching remote files (plus some meta stuff like branches)
3) I'd also claim that repo_base should be also abstracted from external logic like: UI (and progress indication), persistence, code_inspector, repo deletion, and repo_srv - they are now quite intermixed. But this is **not in this PR so more for discussion**.

Pls have a look at the code. does it make sense ? I'm ready to answer questions :)
P.S. the 2 step zip import PR will follow after this one is agreed

Summary of new method structure (not all, just the ones which were the scope to refactor):

```
REPO BASE
  # common repo management methods
  get_name - gets repo title [ABSTRACT]
  has_remote_source - check if the repo has some source to sync with (online or zip) [ABSTRACT]
  refresh - unified in base
  status - unified in base
  reset_status - unified in base
  reset_remote - unified in base
  deserialize - unified in base
  get_files_remote
  set_files_remote

REPO OFFLINE
  # refinitions
  get_name - always URL
  has_remote_source - true if has remote lines
  reset_remote - dirty hack to preserve remote in offline pkg, didn't find how to make better :)

REPO ONLINE
  # refinitions
  get_name - escaped URL
  get_files_remote - need to fetch remote files
  rebuild_local_checksums - logic differs but should be unified in base later (also see #2086)
  has_remote_source - always true
  # new methods
  fetch_remote - instead of initialize, just for better naming and to separate from refresh
```


src/zcl_abapgit_repo.clas.abap
```diff
@@ -7,6 +7,7 @@ CLASS zcl_abapgit_repo DEFINITION
 
   PUBLIC SECTION.
 
+    INTERFACE zif_abapgit_definitions LOAD .
```
larshp: dont think this statement is required

src/zcl_abapgit_repo.clas.abap
```diff
@@ -66,6 +68,7 @@ CLASS zcl_abapgit_repo DEFINITION
         !is_checks TYPE zif_abapgit_definitions=>ty_deserialize_checks
       RAISING
         zcx_abapgit_exception .
+    TYPE-POOLS abap .
```
larshp: dont think this statement is required

src/zcl_abapgit_repo.clas.abap
```diff
@@ -107,37 +108,55 @@ CLASS zcl_abapgit_repo DEFINITION
...
+        !iv_url             TYPE zif_abapgit_persistence=>ty_repo-url OPTIONAL
+        !iv_branch_name     TYPE zif_abapgit_persistence=>ty_repo-branch_name OPTIONAL
+        !iv_head_branch     TYPE zif_abapgit_persistence=>ty_repo-head_branch OPTIONAL
+        !iv_offline         TYPE zif_abapgit_persistence=>ty_repo-offline OPTIONAL
+        !is_dot_abapgit     TYPE zif_abapgit_persistence=>ty_repo-dot_abapgit OPTIONAL
+        !is_local_settings  TYPE zif_abapgit_persistence=>ty_repo-local_settings OPTIONAL
+        !iv_deserialized_at TYPE zif_abapgit_persistence=>ty_repo-deserialized_at OPTIONAL
+        !iv_deserialized_by TYPE zif_abapgit_persistence=>ty_repo-deserialized_by OPTIONAL
+      RAISING
+        zcx_abapgit_exception .
```
larshp: suggest when making changes to core functionallity, to split the formatting into a separate commit and/or PR, it would have made it easier to review

src/zcl_abapgit_repo.clas.abap
```diff
@@ -403,24 +413,24 @@ CLASS zcl_abapgit_repo IMPLEMENTATION.
   ENDMETHOD.
 
 
-  METHOD rebuild_local_checksums. "LOCAL (BASE)
+  METHOD rebuild_local_checksums.
 
-    DATA: lt_local     TYPE zif_abapgit_definitions=>ty_files_item_tt,
+    DATA:
+          lt_local     TYPE zif_abapgit_definitions=>ty_files_item_tt,
```
larshp: no need to change this part, code is the same, formatting is a bit different, more diffs makes it more difficult to review

src/zcl_abapgit_repo.clas.abap
```diff
@@ -403,24 +413,24 @@ CLASS zcl_abapgit_repo IMPLEMENTATION.
...
           ls_last_item TYPE zif_abapgit_definitions=>ty_item,
           lt_checksums TYPE zif_abapgit_persistence=>ty_local_checksum_tt.
 
-    FIELD-SYMBOLS: <ls_checksum> LIKE LINE OF lt_checksums,
+    FIELD-SYMBOLS:
+                   <ls_checksum> LIKE LINE OF lt_checksums,
```
larshp: same here

src/zcl_abapgit_repo.clas.xml
```diff
@@ -14,6 +14,14 @@
     <FIXPT>X</FIXPT>
     <UNICODE>X</UNICODE>
    </VSEOCLASS>
+   <DESCRIPTIONS>
+    <SEOCOMPOTX>
+     <CLSNAME>ZCL_ABAPGIT_REPO</CLSNAME>
+     <CMPNAME>MT_STATUS</CMPNAME>
+     <LANGU>E</LANGU>
+     <DESCRIPT>TY_RESULTS_TT</DESCRIPT>
```
larshp: please remove this description

src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -229,34 +224,36 @@ CLASS ZCL_ABAPGIT_REPO_ONLINE IMPLEMENTATION.
...
-    DATA: lt_remote    TYPE zif_abapgit_definitions=>ty_files_tt,
+    DATA:
+          lt_remote    TYPE zif_abapgit_definitions=>ty_files_tt,
           lt_local     TYPE zif_abapgit_definitions=>ty_files_item_tt,
           ls_last_item TYPE zif_abapgit_definitions=>ty_item,
           lt_checksums TYPE zif_abapgit_persistence=>ty_local_checksum_tt.
 
-    FIELD-SYMBOLS: <ls_checksum> LIKE LINE OF lt_checksums,
+    FIELD-SYMBOLS:
+                   <ls_checksum> LIKE LINE OF lt_checksums,
```
larshp: the DATA and FIELD-SYMBOLS in this method looks identical to the previous version, but there is a diff

src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -86,31 +69,40 @@ CLASS zcl_abapgit_repo_online DEFINITION
...
+                       iv_text    = 'Fetch remote files' ) ##NO_TEXT.
 
-    reset_status( ).
+    ls_pull = zcl_abapgit_git_porcelain=>pull(
+      iv_url         = get_url( )
+      iv_branch_name = get_branch_name( ) ).
 
-    COMMIT WORK AND WAIT.
+    set_files_remote( ls_pull-files ).
+    set_objects( ls_pull-objects ).
```
larshp: set_objects was previously only called during push?

larshp:
Hi, this is very central functionality, so lets take it slow and make sure we end up with a better solution than before. I agree that currently it is a mess, I've also been spending lots of time to get it cleaned up to where it is now

src/zcl_abapgit_repo.clas.abap
sbcgua: Yeap, inserted by SE80 automatically... removed in review fix

src/zcl_abapgit_repo.clas.abap
sbcgua: same. Removed

src/zcl_abapgit_repo.clas.abap
sbcgua: Agree. reverted to the original.
Although this was a standard pretty print. A bit difficult to isolate them as it is done by se80.

BTW is there any description about recommended editor settings to contribute to AG ? Sometimes I edit code in se80, sometimes in eclipse. And automatic replacements are quite annoying. I think it could be nice to have some "standard recommendation" somewhere in contribution.md ...

src/zcl_abapgit_repo.clas.abap
sbcgua: In this particular case I'd keep it. I did it to compare `rebuild` in `base` and `online`. They were styled a bit differently which made the diff less clear. Now it is visible that the difference is just a couple of lines. I'm going to refactor it later (#2086). Let's keep this particular exception please (and 2 below)

src/zcl_abapgit_repo.clas.abap
sbcgua: see `rebuild` comment above

src/zcl_abapgit_repo.clas.xml
sbcgua: done

src/zcl_abapgit_repo_online.clas.abap
sbcgua: see `rebuild` comment above

src/zcl_abapgit_repo_online.clas.abap
sbcgua: it was `mt_objects = ls_pull-objects` before. So the same, just unified the assignment.

BTW as we touched this: in fact `mt_objects` just called/modified internally (in `fetch` and `push`). I didn't do it but I though to remove `get/set_objects` at all and address just the attribute directly. What do you think ?

src/zcl_abapgit_repo.clas.abap
larshp: I use SE80, but I dont think we can make it the standard, and I'd like not to require Eclipse. Over time, files will have SE80 standard

sbcgua:
Absolutely. :) That's why I made a separate PR not to mix with 2-step-zip-import. Believe me it took some effort and several attempts (4 :) ) to get to the current code proposal too. And ... yes, I didn't check 100%, so let's review, totally agree.

BTW could you please comment the point 3 in the PR text ? Not for this PR for sure but more for future changes.

If it helps, here is a pull flow sequence:
```
- repo->refresh at service_git->pull
  - requests local refresh
  - reset_remote
    - clear mt_remote and mt_status (except offline, redefined method will keep mt_remote)
- repo->deserialize at service_git->pull / gui_deserialize
  - zcl_abapgit_objects=>deserialize
    - repo->get_files_remote
      - fetch_remote (renews the remote data)
...      
- zif_abapgit_gui_page~render
  - browser->list, build_items_online
    - repo->status
      - zcl_abapgit_file_status=>status
        - get_files_local - re-serialize due to mv_request_local_refresh
        - get_files_remote - cached
```

src/zcl_abapgit_repo.clas.abap
```diff
@@ -107,15 +106,27 @@ CLASS zcl_abapgit_repo DEFINITION
...
   PROTECTED SECTION.
 
     DATA mt_local TYPE zif_abapgit_definitions=>ty_files_item_tt .
     DATA mt_remote TYPE zif_abapgit_definitions=>ty_files_tt .
-    DATA mv_do_local_refresh TYPE abap_bool .
+    DATA mv_request_local_refresh TYPE abap_bool .
     DATA mv_last_serialization TYPE timestamp .
     DATA ms_data TYPE zif_abapgit_persistence=>ty_repo .
     DATA mv_code_inspector_successful TYPE abap_bool .
+    DATA mv_remote_is_initialized TYPE abap_bool .
```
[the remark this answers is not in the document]
sbcgua: I see what you mean ... let's discuss, this topic is not that clear imho

1) Ideally (ideally) I would remove both attributes ...
2) `mv_request_local_refresh` is used only in `get_files_local` - do you think checking `mt_local is initial` is a safe replacement ?
3) same for `mv_remote_is_initialized`, in fact just used in `fetch_remote`. Would the `mt_remote is initial` be safe enough ?

What do you think ?

If both "no", I'll rename `mv_remote_is_initialized` to `mv_request_remote_refresh`. Because the mv_request_LOCAL_refresh fits it's role really well now.

src/zcl_abapgit_repo_offline.clas.abap
```diff
@@ -5,11 +5,40 @@ CLASS zcl_abapgit_repo_offline DEFINITION
...
+  ENDMETHOD.
+
+
+  METHOD reset_remote.
+
+    DATA lt_backup LIKE mt_remote.
+
+    lt_backup = mt_remote.
+    super->reset_remote( ).
+    set_files_remote( lt_backup ).
```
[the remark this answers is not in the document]
larshp: Suggest adding a comment explaining why it works this way

src/zcl_abapgit_repo.clas.abap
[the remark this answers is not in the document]
larshp: I would also like to remove both attributes, but I'm not sure it is safe, I think "no" for this PR, we can move the issue to a new issue/PR

Suggest just renaming so both are similar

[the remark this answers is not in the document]
sbcgua: ok, done

larshp:
cool, merging 👍

PR #2126, opened 2018-11-21
2-step zip import

sbcgua:
Second try for 2-step zip import (#1950)
Should work well, but would be nice if someone can double test (@christianguenter2 seems like you were interested in the feature last time? :)) ).

Notes:
- No changes were done to: `jump_display_transport` - it's se80 auto reordering :) Also just  `colspan` fixes in `render_parent_dir`
- completely removed deserialize from zip. So no new actions compared to previous approach in #1953 (which is good)
- renamed methods of `ZCL_ABAPGIT_REPO_CONTENT_LIST` for sematics reasons

### DIFF
--- a/src/ui/zcl_abapgit_gui_router.clas.abap
+++ b/src/ui/zcl_abapgit_gui_router.clas.abap
@@ -47,7 +47,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_gui_router IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_GUI_ROUTER IMPLEMENTATION.
 
 
   METHOD get_page_background.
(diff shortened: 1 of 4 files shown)


larshp:
merge conflicts...

sbcgua:
Conflict after 'Open in master language'. Did a rebase onto latest master

sbcgua:
"Import zip": hmmm. I added "zip" to emphasize that all these operations (Pull/Import/Export) are related to external files and not to remote repo. I think this is clearer. But indeed a bit bulky. Technically "Attach zip" will be more correct. But probably less understandable. Or we can arrows instead of works (+ some hint). Kind of:

image

or (icon based)

image

... or leave as is for the time being

larshp:
just leave as is for now 😄

PR #2148, opened 2018-11-25
Reset sy-langu after open repo in master language

christianguenter2:
Reset language, because otherwise abapGit continues with wrong sy-langu


christianguenter2:
closed accidentially

Issue #2142, opened 2018-11-24
UI: Design and UX improvements of hotkeys


sbcgua:
Proposal
- Cleaned design of hotkeys popup
- unified css and js for news and hotkeys
- `chunk_lib=>render_infopanel` wrapper for popups
- bottom corner hint about hotkeys
- hotkey list shows just shortcuts, relevant to current page
- minor changes in hotkey descriptions

![2018-11-25_00h18_47](https://user-images.githubusercontent.com/15635498/48973564-98680480-f04a-11e8-9b2c-1dca4564e66b.png)

sbcgua:
hmmm, maybe ... it makes sense. But it uses the same logic as news. And news may be longer than screen. Then there will be the logic for news and another one for hotkeys. 
But OK, i'll check it

sbcgua:
Can be closed ?

PR #2155, opened 2018-11-27
hotkeys in fixed position

sbcgua:
to #2142 , @christianguenter2 request

- added `iv_scrollable` param to the `RENDER_INFOPANEL` wrapper. So now it is controllable
- respectively hotkeys render the panel with this flag = false


sbcgua:
Removed conflicts. Rebased on master

PR #2179, opened 2018-12-01
Report name

gregorwolf:


flaiker:
I'd recommend not to name the report ZABAPGIT. Otherwise if you switch to the full online installation afterwards you'll get a conflict.

gregorwolf:
I've made the adjustment to ZABAPGIT_FULL. Good idea.

PR #2185, opened 2018-12-04
 Handle structure "type ref to " circular dependency

EduardoCopat:
This PR solves the circular dependency discussed #2015. The best way to test it is to clone @christguenter2 [example](https://gitlab.com/christianguenter/test_activation.git). I also tested cloning ABAP open checks.

In short:
- I created a multistep deserialization for CLAS and INTF. Basically, you first deserialize the class without any methods or includes. 
- With this, the structure with a type ref to an interface or class works. Then, you can continue with the regular serialization.

Example:
1. First step deserialization - only class/interface "name":
```abap
class ZCL_TEST_ACTIVATION definition
  public
  final
  create public .
  public section.
  PROTECTED SECTION.
  PRIVATE SECTION.
ENDCLASS.
CLASS ZCL_TEST_ACTIVATION IMPLEMENTATION.
  method FOO.
  endmethod.
ENDCLASS.
```
2. Second step - the rest
```abap
class ZCL_TEST_ACTIVATION definition
  public
  final
  create public .

public section.

  interfaces ZIF_TEST_ACTIVATION .

  types INSTANCES type ZTEST_ACTIVATION .

  methods FOO .
  PROTECTED SECTION.
  PRIVATE SECTION.
ENDCLASS.
CLASS ZCL_TEST_ACTIVATION IMPLEMENTATION.


  method FOO.
  endmethod.
ENDCLASS.
```


src/objects/zcl_abapgit_object_clas.clas.abap
```diff
@@ -37,15 +38,20 @@ CLASS zcl_abapgit_object_clas DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
...
+
 CLASS zcl_abapgit_object_clas IMPLEMENTATION.
 
+
   METHOD constructor.
     super->constructor( is_item     = is_item
                         iv_language = iv_language ).
 
     CREATE OBJECT mi_object_oriented_object_fct TYPE zcl_abapgit_oo_class_new.
+    mv_activation_placeholder_step = abap_true.
```
EduardoCopat: The bad thing about this is that it introduces state to the deserializer. Another option would be to have a completely different deserializer only for the class "name"

src/objects/zcl_abapgit_object_clas.clas.abap
```diff
@@ -81,35 +87,197 @@ CLASS zcl_abapgit_object_clas IMPLEMENTATION.
...
+      mi_object_oriented_object_fct->update_descriptions(
+        is_key          = ls_class_key
+        it_descriptions = lt_descriptions ).
+      mv_activation_placeholder_step = abap_true.
+    ENDIF.
+
+  ENDMETHOD.
+
+
+  METHOD deserialize_docu.
```
EduardoCopat: Pretty printer...

larshp:
I tested the suggestion from https://github.com/larshp/abapGit/issues/2015#issuecomment-433053014

Test repo: https://github.com/larshp/type_ref_to

Deserialize changed to the following to force CLAS to be deserialized first

```abap
break-point.
    deserialize_objects( EXPORTING it_objects = lt_rest
                                   iv_descr   = 'Objects'
                         CHANGING ct_files = rt_accessed_files ).

    deserialize_objects( EXPORTING it_objects = lt_ddic
                                   iv_ddic    = abap_true
                                   iv_descr   = 'DDIC'
                         CHANGING ct_files = rt_accessed_files ).
```

And it seems to work without any problems. So I suggest this approach again

larshp:
what is the error?

EduardoCopat:
Syntax error. Because it refers to something that has not been created yet:

image

larshp:
looks like the force activation of INTF did not work

larshp:
I guess `DESERIALIZE_SOURCE` only saves the code as inactive, and then tries to activate. Suggest cleaning up the ZCL_ABAPGIT_OO_* objects, I have a difficult time understanding the setup, it seems there are too many classes. Plus it looks like DESERIALIZE_SOURCE is only called for interfaces, which means it should not be part of BASE. And my SE80 navigation is broken :(

larshp:
will close this and move the discussion to #2015

EduardoCopat:
👍

PR #2231, opened 2018-12-27
Fix ADT link in readme

flaiker:


flaiker:
(apparently also removal of trailing whitespace)

Issue #2275, opened 2019-01-10
Octicons - use class mega-octicon depending on DPI setting


bigld:
The 16px font size from the used CSS class .octicon may be too small for high DPI setting:
![image](https://user-images.githubusercontent.com/4643389/50960511-d4094180-14c5-11e9-96c4-3e37046b5695.png)

Use class .mega-octicon depending on DPI settings

PR #2276, opened 2019-01-10
switch to css class mega-octicon for HiDPI #2275

bigld:
#2275

### DIFF
--- a/src/ui/zcl_abapgit_html.clas.abap
+++ b/src/ui/zcl_abapgit_html.clas.abap
@@ -90,7 +90,7 @@ ENDCLASS.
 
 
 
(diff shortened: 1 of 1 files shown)


larshp:
can you add a screenshot of how it looks after the change?

bigld:
HiDPI after fix
image

sbcgua:
seems like it is not padding but the fixed width of icon cell ...
could you pls try to replace the style in css:

```css
.repo_tab td.icon {
    width: 1px;
    text-align: center;
    padding-left: 8px;
    padding-right: 4px;
}
```

or alternatively merge, i'll PR the fix ... (if i find a HiDPI screen to check it...)

sbcgua:
:+1:

bigld:
HiDPI looks good:
image

Issue #2337, opened 2019-01-27
Serialize view cluster inactive/not exits


GoWale:
When serialize a view cluster that is inactive it will not get exported.

### Step to reproduce
Just create an package with a basic view cluster (se54).
Activate once (The TOBJ will be created on the first activation)
Change the cluster - adjust text and save
![grafik](https://user-images.githubusercontent.com/9630039/51800700-494d7400-2233-11e9-86a4-5a2ca135cb1e.png)

Save the cluster but "forget to activate".

When add the package wit the view cluster that object is not shown at all:
![grafik](https://user-images.githubusercontent.com/9630039/51800738-1952a080-2234-11e9-9f92-1963b7441fc4.png)

But in tadir the object exists:
![grafik](https://user-images.githubusercontent.com/9630039/51800722-af39fb80-2233-11e9-8add-eeaeea950506.png)

### Situation after active the view cluster
![grafik](https://user-images.githubusercontent.com/9630039/51800747-4ef78980-2234-11e9-8d61-f01007aef716.png)

![grafik](https://user-images.githubusercontent.com/9630039/51800749-60d92c80-2234-11e9-9045-9e3ab09dbb37.png)

### Caused by VCLS object
ZCL_ABAPGIT_OBJECT_VCLS->ZIF_ABAPGIT_OBJECT~EXISTS

### suggestion
Adjust the coding so a existing but inactive view cluster will be shown as inactive an not disappear

PR #2339, opened 2019-01-27
fix for serialize view cluster inactive/not exits

GoWale:
fix for issue #2337

In addition add logic to check if the view cluster is opend in change mode (via se54 or sm34)

### DIFF
--- a/src/objects/zcl_abapgit_object_vcls.clas.abap
+++ b/src/objects/zcl_abapgit_object_vcls.clas.abap
@@ -4,18 +4,18 @@ CLASS zcl_abapgit_object_vcls DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
     INTERFACES zif_abapgit_object.
 
   PROTECTED SECTION.
-private section.
+  PRIVATE SECTION.
 
 * See include MTOBJCON:
-  constants C_CLUSTER_TYPE type C value 'C' ##NO_TEXT.
-  constants C_MODE_INSERT type OBJ_PARA-MAINT_MODE value 'I' ##NO_TEXT.
+    CONSTANTS c_cluster_type TYPE c VALUE 'C' ##NO_TEXT.
+    CONSTANTS c_mode_insert TYPE obj_para-maint_mode VALUE 'I' ##NO_TEXT.
 
-  class-methods CHECK_LOCK
-    importing
-      !IV_VIEWNAME type TABNAME
-      !IT_SELLIST type SCPRVIMSELLIST
-    returning
-      value(RV_IS_LOCKED) type ABAP_BOOL .
+    CLASS-METHODS check_lock
+      IMPORTING
+        !iv_viewname        TYPE tabname
+        !it_sellist         TYPE scprvimsellist
+      RETURNING
+        VALUE(rv_is_locked) TYPE abap_bool.
 ENDCLASS.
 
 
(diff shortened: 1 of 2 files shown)


src/objects/zcl_abapgit_object_vcls.clas.abap
```diff
@@ -5,17 +5,52 @@ CLASS zcl_abapgit_object_vcls DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
...
   PROTECTED SECTION.
   PRIVATE SECTION.
+
 * See include MTOBJCON:
-    CONSTANTS: c_cluster_type TYPE c VALUE 'C'.
-    CONSTANTS: c_mode_insert  TYPE obj_para-maint_mode VALUE 'I'.
+    CONSTANTS c_cluster_type TYPE c VALUE 'C' ##NO_TEXT.
+    CONSTANTS c_mode_insert TYPE obj_para-maint_mode VALUE 'I' ##NO_TEXT.
 
+    CLASS-METHODS check_lock
```
christianguenter2: There's a method exists_a_lock_entry_for for reading enques in zcl_abapgit_objects_super.
https://github.com/larshp/abapGit/blob/06616580eb500d94d1d6ca165638c7c4ac01cf61/src/objects/zcl_abapgit_objects_super.clas.abap#L187-L212

Maybe you can use that method instead of locking and releasing?

src/objects/zcl_abapgit_object_vcls.clas.abap
```diff
@@ -141,13 +171,52 @@ CLASS ZCL_ABAPGIT_OBJECT_VCLS IMPLEMENTATION.
...
+    SELECT SINGLE changedate INTO lv_changedate FROM vcldir
+      WHERE vclname = ms_item-obj_name.
+
+    IF lv_changedate IS NOT INITIAL.
+* see logic in function module VIEWCLUSTER_GET_DEFINITION
+      rv_active = abap_true.
+    ELSE.
+      rv_active = abap_false.
+    ENDIF.
+
```
christianguenter2: Hm, how does that work? What's the connection between changedate and is_active?

src/objects/zcl_abapgit_object_vcls.clas.abap
```diff
@@ -141,13 +171,52 @@ CLASS ZCL_ABAPGIT_OBJECT_VCLS IMPLEMENTATION.
...
+        rv_is_locked = check_lock( iv_viewname = 'V_VCLSTDEP'
+                                   it_sellist  = lt_sellist ).
+
+        IF rv_is_locked = abap_false.
+          rv_is_locked = check_lock( iv_viewname = 'V_VCLMF'
+                                     it_sellist  = lt_sellist ).
+
+        ENDIF.
+      ENDIF.
+    ENDIF.
```
christianguenter2: Can you reduce the nesting of the conditionals?

src/objects/zcl_abapgit_object_vcls.clas.abap
GoWale: I think this is possible. I will adjust the section.

src/objects/zcl_abapgit_object_vcls.clas.abap
GoWale: Some thing only SAP will know...
The main reason I think is that view clsuters are more a client indipendet customizing than real DDIC objects.
the logic behind this statement can be found in VIEWCLUSTER_GET_DEFINITION (line 25/26 on my release).
I also don't know a other flag to ident if a view cluster is active or not.

src/objects/zcl_abapgit_object_vcls.clas.abap
GoWale: Will change when I reworke the lock method

PR #2392, opened 2019-02-13
Refactoring of CORR_INSERT and adjustment for object types

### LINKED ISSUE #2378: Transport Object for MSAG
Hi,

when messages classes are de-serialized, object type "LIMU MSAD <msgclass>" is added to the transport request (object_class = 'T100' in combination with global_lock='X' -> LIMU MSAD):
   CALL FUNCTION 'RS_CORR_INSERT'
      EXPORTING
        global_lock         = abap_true
        devclass            = iv_package
        object              = ls_t100a-arbgb
        object_class        = 'T100'
        mode                = 'INSERT'
When the message class is deleted, "R3TR MSAG <msgclass>" is used.

Why is "LIMU MSAD <msgclass>" used for de-seriialization instead of "R3TR MSAG <msgclass>"?

Regards,
Martin

mfsap:
Method `ZCL_ABAPGIT_OBJECTS_SUPER->CORR_INSERT` enhanced to be used for several object types (dictionary and non-dictionary) in dialog and dark mode.
Fixes #2378

### DIFF
--- a/src/objects/zcl_abapgit_objects_super.clas.abap
+++ b/src/objects/zcl_abapgit_objects_super.clas.abap
@@ -142,12 +142,8 @@ CLASS ZCL_ABAPGIT_OBJECTS_SUPER IMPLEMENTATION.
 
   METHOD corr_insert.
 
-    DATA: lv_object           TYPE string,
-          lv_object_class     TYPE string,
-          lv_gui_is_available TYPE c,
-          lv_korrnum_check    TYPE e070-trkorr,
-          lv_ordernum_check   TYPE e070-trkorr,
-          lv_korrnum          TYPE e070-trkorr.
+    DATA: lv_object       TYPE string,
+          lv_object_class TYPE string.
 
     IF iv_object_class IS NOT INITIAL.
       lv_object_class = iv_object_class.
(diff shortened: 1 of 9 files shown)


src/objects/zcl_abapgit_objects_super.clas.abap
```diff
@@ -141,28 +142,97 @@ CLASS ZCL_ABAPGIT_OBJECTS_SUPER IMPLEMENTATION.
...
+          mode                = 'I'
+        EXCEPTIONS
+          cancelled           = 1
+          permission_failure  = 2
+          unknown_objectclass = 3
+          OTHERS              = 4.
+    ELSE.
+
+      "check whether object is currently locked in different transport
+      CALL FUNCTION 'RS_CORR_CHECK'
```
larshp: Looks like this part relates to https://github.com/larshp/abapGit/issues/2345 ? and not issue #2378 directly

Suggest removing this fix from this PR, and submitting it separately. I also wonder why the logic is different for background and foreground mode?

mfsap: In foreground mode, a popup is shown that the object is locked in a different request and the object can only be recorded in that request. In background mode, the object is automatically assigned to the other request w/o showing a popup. I agree, that this is a different issue than #2378 - I've added the complete `CORR_INSERT` method (with the 'non SAPGUI version') as this is required for our project. If this is too much for #2378, I'll reduce the coding in this pull request and deal with the difference in our project or create a different pull request later...

larshp: the transport popup is handled via `zcl_abapgit_default_transport` and `zcl_abapgit_ui_factory=>get_popups( )->popup_transport_request`, in GUI_DESERIALIZE. If this method shows a popup, then something is wrong somewhere else, so this looks like a workaround for a broken feature that is implemented elsewhere

larshp: how can above issue be reproduced?

larshp:
tested ok,
image

larshp:

PR #2463, opened 2019-02-27
Solved: Maximum Number of GUI Session Reached

kuzeydev:
Solved: Maximum Number of GUI Session Reached

### DIFF
--- a/src/zcl_abapgit_serialize.clas.abap
+++ b/src/zcl_abapgit_serialize.clas.abap
@@ -93,6 +93,8 @@ CLASS zcl_abapgit_serialize IMPLEMENTATION.
 
   METHOD determine_max_threads.
 
+    DATA lt_user_info TYPE STANDARD TABLE OF uinfo2.
+
     IF iv_force_sequential = abap_true.
       rv_threads = 1.
       RETURN.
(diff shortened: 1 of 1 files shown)


kuzeydev:
> see #2187

I have already review #2187.
The problem is you may not use all threads the user has. Sometimes users have other GUI session active.  You may use only ( MaxThread - ActiveThread ) threads.

larshp:
error "Maximum Number of GUI Session Reached" is not related to the general server setup, but the number of gui sessions allowed for the user. As mentioned in #2187, parallel processing should not take up any gui sessions for the user, if it does there is a bug in the serialization code.

Have you installed SAP note 2759002?

What SAP version are you on?

kuzeydev:
Please check: 
https://help.sap.com/doc/abapdocu_751_index_htm/7.51/en-US/abapcall_function_starting.htm

It says: "... If the destination is not specified and also not defined in a callback routine using the addition KEEPING TASK of the statement RECEIVE, the destination "NONE" is used implicitly. ..."

image

if destination 'NONE' is used then a new user session is opened and It is same as GUI Session..

kuzeydev:
> error "Maximum Number of GUI Session Reached" is not related to the general server setup, but the number of gui sessions allowed for the user. As mentioned in #2187, parallel processing should not take up any gui sessions for the user, if it does there is a bug in the serialization code, not the parallelization.
> 
> Have you installed SAP note 2759002?
> 
> What SAP version are you on?

S4HANA 1809.
After Implementing 2759002, It seems problem is solved. thanks

PR #2488, opened 2019-03-07
docs css improvements

sbcgua:
inline code in docs look ugly

2019-03-07_18h24_36

image

propose to improve

2019-03-07_18h28_15

2019-03-07_18h21_38


sbcgua:
removed italic, no problem. looks fine too for me.

PR #2508, opened 2019-03-13
fix filter_files_to_deserialize

mfsap:
in method zcl_abap_objects->filter_files_to_deserialize the object condense is done too early, need to be done at the end

### DIFF
--- a/src/zcl_abapgit_objects.clas.abap
+++ b/src/zcl_abapgit_objects.clas.abap
@@ -727,14 +727,14 @@ CLASS ZCL_ABAPGIT_OBJECTS IMPLEMENTATION.
     rt_results = it_results.
 
     DELETE rt_results WHERE match = abap_true.     " Full match
+    DELETE rt_results WHERE obj_type IS INITIAL.
+    DELETE rt_results WHERE lstate = zif_abapgit_definitions=>c_state-added AND rstate IS INITIAL.
+
     SORT rt_results
       BY obj_type ASCENDING
          obj_name ASCENDING
          rstate   DESCENDING. " ensures that non-empty rstate is kept
     DELETE ADJACENT DUPLICATES FROM rt_results COMPARING obj_type obj_name.
-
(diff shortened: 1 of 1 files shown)


jfilak:
Thank you for taking the time open this PR. Could please add a short description on how to test the patch fixes the problem outlined in the commit message?

mfsap:
Old coding looks like this:
```
    DELETE rt_results WHERE match = abap_true.     " Full match
    SORT rt_results
      BY obj_type ASCENDING
         obj_name ASCENDING
         rstate   DESCENDING. " ensures that non-empty rstate is kept
    DELETE ADJACENT DUPLICATES FROM rt_results COMPARING obj_type obj_name.
    DELETE rt_results WHERE obj_type IS INITIAL.
    DELETE rt_results WHERE lstate = zif_abapgit_definitions=>c_state-added AND rstate IS INITIAL.
```
The `SORT` and `DELETE ADJACENT DUPLICATES` reduces the file list to a single file to get an object list at the end - but the remaining file for the object is more less "random". After the SORT, there is a `DELETE ... WHERE lstate = ... AND rstate IS INITIAL`. This may to a deletion of the file (and the deletion of the object) when there are other files that indicates a change are not taken into account.
In the UI, this object - if it has changed files - is not shown in the list for changed objects and during the pull operation, an error/abort occurs that indicates that the "**Object xyz is undecided**".
So I moved the `SORT` and `DELETE ADJACENT DUPLICATES` to the end of the method:
```
    DELETE rt_results WHERE match = abap_true.     " Full match
    DELETE rt_results WHERE obj_type IS INITIAL.
    DELETE rt_results WHERE lstate = zif_abapgit_definitions=>c_state-added AND rstate IS INITIAL.
    SORT rt_results
      BY obj_type ASCENDING
         obj_name ASCENDING
         rstate   DESCENDING. " ensures that non-empty rstate is kept
    DELETE ADJACENT DUPLICATES FROM rt_results COMPARING obj_type obj_name.
```

filak-sap:
Thank you!

PR #2644, opened 2019-04-22
Transport overview page

flaiker:
Draft pull request for discussion.

I would like to streamline the process to work with feature branches based on transports in abapGit. Currently you either use the transport-to-branch feature or manually fiddle with branches and committing the correct objects.

I added a page that groups objects by the transport they are currently locked in. If there is no branch on the remote that has the same name as the transport you can click a button to create that branch. If that branch already exists you can commit changes to it. The stage screen is filtered so only objects of that transport show up. The diff is calculated against that branch. If the branch does not exist yet master is used for comparison (this should lead to all uncommitted changes being displayed).

image

image

In my opinion this could provide a somewhat missing link between CTS and Git. I however am not sure if this belongs in core abapGit (as it kind of forces a specific branching model) and if it does to what extend. Building this outside of abapGit also seems weird and difficult though.

Opinions?

(Proof of concept, for merging performance needs to be optimized, syntax backported, CSS adjusted etc.)


larshp:
perhaps add it in the view dropdown instead of the menubar?

image

src/zif_abapgit_definitions.intf.abap
```diff
@@ -314,16 +314,22 @@ INTERFACE zif_abapgit_definitions
...
+      obj_name       TYPE tadir-obj_name,
+      inactive       TYPE abap_bool,
+      sortkey        TYPE i,
+      path           TYPE string,
+      is_dir         TYPE abap_bool,
+      changes        TYPE i,
+      lstate         TYPE char1,
+      rstate         TYPE char1,
+      files          TYPE tt_repo_files,
+      is_transport   TYPE abap_bool,
```
larshp: add this part as sub structure?

flaiker:
> perhaps add it in the view dropdown instead of the menubar?
> 
> image

You mean as a toggle to switch the main repo page into this layout or just move the link?
At a previous attempt I had in the main page but noticed that I basically had to remove all other UI elements as they don't make sense for this.

flaiker:
Update:
image

I added a local repo setting "CTS target branch" which indicates the branch transport-based branches get merged into (in most cases master). If it is set it controls which branch is taken as a base for the "create branch" button. If it is not set the actions for commit and create branch are not shown and the currently checked out branch is used for diffs.
I figured this way you have to first enable the workflow and get a chance to read some documentation. If you don't enable it you can still see all changes grouped by transports that are not yet in the current branch.

larshp:
I'd probably suggest defaulting it to HEAD or master, or let the user specify in a popup, but this is a minor issue

larshp:
this is getting quite old, can we close the PR for now, and reopen when its ready?

flaiker:
Sure, no problem

ed-holland:
I'm looking into a similar integration with CTS for streamlining our change process with AbapGit. Where you able to finish it?

Our process requires a TL to create a Transport Request and Change requests for the developers.
To enforce, it could be done when creating a branch (creating TR and linking it with Branch).
Commits to the branch would also release the Change Request.
Pull Request would release also the Transport Request.
This way the abapGit would enforce the process.

larshp:
this pull request is closed, feel free to open new issues with questions/suggestions/comments/feature requests 😄

flaiker:
@ed-holland 

One of the main problems with this was the relation between objects in transports and files in abapGit. For example abapGit does currently not determine the correct transport for LIMU-objects in function groups (#2961) and for other things it is simply impossible to create the correct relation (for example if two developers change two methods of the same class in two different transports).

I worked around some of the problems by implementing the CTS BAdIs and syncing transport state with GitLab merge requests (https://github.com/flaiker/ci-lib) but in the end it is a lot of work with lots of edge cases and technical limitations.

Issue #2673, opened 2019-05-16
Performance issue when patching large files


FreHu:
Try adding all lines of a file with >500 lines. This can easily freeze the UI for over 10 seconds.

FreHu:
I'm attaching some profiling results. You can import them in the IE debugger (don't use the one within sapgui, this failed for me).

It seems like all it's doing is toggling classes, which adds up:
![image](https://user-images.githubusercontent.com/5097067/57834255-2e8c1480-77bc-11e9-9539-5b45e1014b85.png)

I also feel that the performance gets worse with the size of the diff page, no matter the size of individual files.

[patch.zip](https://github.com/larshp/abapGit/files/3185821/patch.zip)

christianguenter2:
Patch UI relies heavily on CSS classes and the rendering engine of SAPGUI browser control is quite slow.
I'm not sure whether there's an easy way to improve the performance without reengineering the UI. 

An idea is mentioned here, but I'm not sure whether the suggested techniques are supported by our archaic rendering engine.
https://ultimatecourses.com/blog/stop-toggling-classes-with-js-use-behaviour-driven-dom-manipulation-with-data-states

FreHu:
Looks like it's toggling two things - the add link and the remove link - for each line. The second link isn't necessary, you can just toggle one and style it differently based on whether the line is included.

I guess checkboxes could also be used, though I'm dubious about both the performance and looks of hundreds of checkboxes. It's probably not what they were intended for :)

christianguenter2:
Just submitted a PR replacing links with cheboxes. I seems like a good idea from both performance and UX perspective.

PR #2684, opened 2019-05-18
Patch: replace links with checkboxes #2673

christianguenter2:
#2673
With this commit we replace the links on the patch page with checkboxes.
This has several advantages:
- better performance
- simplified and less cluttered UI
- less and easier to understand code

before:
image

after:
image

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -839,7 +839,7 @@ LinkHints.prototype.tooltipActivate = function (oTooltip) {
 
   if (elInput) {
     // case 1) toggle the checkbox
-    elInput.checked = !elInput.checked;
+    elInput.click();
   } else {
     // case 2) click the link
     oTooltip.parentElement.click();
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -1114,12 +1114,12 @@ Patch.prototype.preparePatch = function(){
 
 Patch.prototype.registerClickHandlerForFiles = function(){
   // registers the link handlers for add and remove files
-  this.registerClickHandlerForPatchFile("input[id^='" + PatchFile.prototype.ID + "']");
+  this.registerClickHandlerForSelector("input[id^='" + PatchFile.prototype.ID + "']", this.onClickFileCheckbox);
 };
 
 Patch.prototype.registerClickHandlerForSections = function(){
   // registers the link handlers for add and remove sections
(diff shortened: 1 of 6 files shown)


src/ui/zabapgit_js_common.w3mi.data.js
```diff
@@ -1205,67 +1146,58 @@ Patch.prototype.registerClickHandlerForSelector = function(sSelector, fnCallback
...
+  sId = sId.replace(oRegex, PatchLine.prototype.ID);
+  return document.querySelectorAll("input[id^='"+ this.escape(sId) + "']");
 };
 
-Patch.prototype.togglePatchForElem = function(elLink) {
-  new PatchLine(elLink.id).toggle();
+Patch.prototype.getAllSectionCheckboxesForId = function(sId, sIdPrefix){
+  var oRegex = new RegExp("^" + sIdPrefix);
+  sId = sId.replace(oRegex, PatchSection.prototype.ID);
+  return document.querySelectorAll("input[id^='"+ this.escape(sId) + "']");
```
FreHu: Since the files are patched one by one, you could narrow the query to the section with the patched file instead of the whole document to boost performance even further.

FreHu: Also, getElementById can be used here, which is faster than the general querySelectorAll
https://web.archive.org/web/20160108040024/http://jsperf.com/getelementbyid-vs-queryselector

christianguenter2: Hm, I'm not quite sure how to do that. General idea is that you have the section id (e.g. patch_section_zbunkai_christian.prog.abap_1) and then find all checkboxes starting with the same id prefix (e.g. patch_**line**_zbunkai_christian.prog.abap_1).

image

Imho the query is quite narrow and getElementById cannot be used, because we don't know the exact id and we'd like to select multiple elements. Or am I missing the obvious?

Btw. perfomance improved already dramatically. Patching objects with 5000+ changed lines of code withinin the blink of an eye.

FreHu: My bad, I thought you were retrieving each line by its id separately.  The checkboxes should be more than enough then.

larshp:
@FreHu did you test and review this? please add a review, if you review it okay then I'll merge it 😄

FreHu:
I only went through the code, I think I'll get time to test it during the day.

christianguenter2:
Did some refactoring and fixed a minor issue with link hints. Ready to merge ;-)

christianguenter2:
> Tried it, the checkboxes are a bit harder to click than the links

Yeah, we can improve that afterwards. E.g. increase the checkboxes with some CSS tricks or make the whole cell clickable.

Issue #2698, opened 2019-05-23
Dump ASSERT_TRUE()/ASSERT_FALSE() 702 DevVersion


bigld:
ZCL_ABAPGIT_GUI_UTILS UT uses ASSERT_TRUE()/ASSERT_FALSE()

PR #2699, opened 2019-05-23
downport assert_true()/false() #2698

bigld:
downport assert_true()/false() #2698


FreHu:
If you need to downport asserts, maybe it is worth using something like this:
https://gist.github.com/FreHu/c8b3e32b115c82a7448c536cbed8427f#file-cl_abap_unit_wrapper-clas-abap

It wraps all cl_abap_unit_assert methods (minus deprecated), so you can just reimplement the missing assert_true/assert_false, but in one place. 

Make all your test classes inherit from this so you always have the asserts + no longer need to write `cl_abap_unit_assert=>` everywhere

PR #2707, opened 2019-05-26
fix for issue https://github.com/larshp/abapGit/issues/2706 for class/interface

arcanist123:
fix for issue https://github.com/larshp/abapGit/issues/2706 for class/interface


arcanist123:
I could make the code 7.02 compatible but the current implementation is not. so i am not sure is it is relevant

PR #2758, opened 2019-06-25
ADT jump for package link and transports

### LINKED ISSUE #2757: Jump to ADT for top-right package link
Opening a package from the repository contents will jump to ADT.
image

Clicking this link does not jump to ADT.
image

Both should behave the same.

FreHu:
Fixes #2757 

I also plan to do jump to transport requests/tasks, but I need to do some experimenting as these work slightly differently (they are not development objects).


FreHu:
Adding #2761, it's very related.

PR #2765, opened 2019-06-29
Improve css preprocessing

sbcgua:
Followup to #2610, #2700, #2733 
@flaiker very welcome to review :)

Goals:
- make css preprocessing more abstract, less explicit
- improve debugging (copy/paste HTML from system required non-trivial css links fixing )

Solution:
- introducing `zif_abapgit_gui_html_processor` and "default" `zcl_abapgit_gui_html_processor` which is 
  - triggered prior to `cache_asset` in `gui->render`
  - finds css links in html head
  - comments them out
  - collects urls from them
  - runs css processor over them
  - integrates the result into html
- the instance of processor is created in factory, external to GUI, and, in particular, populated with css links to preserve unchanged
- thus pages should not care about proper preprocess css naming as well as GUI itself
- `gui->cache_asset` separated into an interface, the code didn't change - it is SE80 reorderings. There are some ideas how to reuse the interface in future + to isolate critical methods like `free` ...
- minor
  - css processor reuse get_**text**_asset. Otherwise the code didn't change - it is SE80 reorderings. But some changes to UTs: for get_text_asset and also removed asset_man double - the existing class is abstract enough to be used in tests
  - improvements to asset_man UTs
  - there are some indent alignments I didn't do to make the commit clearer, can do before the merge or separately

P.S. will probably add some doc in a separate PR


sbcgua:
result sample

```html
  <head>
    <meta http-equiv="content-type" content="text/html; charset=utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=11,10,9,8" />
    <title>abapGit</title>
    <link rel="stylesheet" type="text/css" href="css/common.css">
    <link rel="stylesheet" type="text/css" href="css/ag-icons.css">
    <!-- by AG HTML preprocessor <link rel="stylesheet" type="text/css" href="css/theme-default.css">-->
    <!-- by AG HTML preprocessor <link rel="stylesheet" type="text/css" href="css/theme-belize-blue.css">-->
    <script type="text/javascript" src="js/common.js"></script>
    <style>.icon.large { font-size: inherit }</style>
    <!-- by AG HTML preprocessor --><link rel="stylesheet" type="text/css" href="css/bundle.css">
  </head>
```

src/ui/core/zcl_abapgit_gui_css_processor.clas.testclasses.abap
```diff
@@ -101,6 +70,7 @@ CLASS ltcl_single_file IMPLEMENTATION.
   ENDMETHOD.
 
   METHOD test_empty_file.
+    " TODO REMOVE ?? why test empty files ?
```
flaiker: Why not? Test all the things.

sbcgua: Well, yes, point is that real assetman actually **fails** when data is empty (opposite to the double).
So my statement is inaccurate. Probably the test should not be removed but changed to fail if no exception happened ... I'll fix

flaiker: Ah okay. Then it makes sense to align these two.
I just didn't expect the asset manager to fail if an asset is empty.

sbcgua:
> How exactly does this improve debugging?

Well, not that much, but ... the point is that the original CSS links remain in the html commented (above was an example). So it is easy to comment the "bundle" and uncomment "originals" and it works. No need to copy/paste or remember names. I prefer to save all the files to the PC and edit them in a good editor and view in chrome (although IEChooser is cool indeed). I have a [tool](https://github.com/sbcgua/abap_w3mi_poller) which simplifies file update so no manual SMW0 :)

> CSS preprocessing unnecessarily happened on every render

Yeap, it is the same. Good point, didn't think about this honestly. But imho it is not very big issue. It does not affect performance a lot from what I see, besides, what if specific page will need some specific additional css? Option is: some kind of global variable to hold the last set of css urls built in the bundle ... or ... maybe create the name of the bundle in some predictable and unique way like concatenation of css urls or some hash ( 'css/theme.css/css/theme-belize.css' or 'css/{ sha1('css/theme.css/css/theme-belize.css') }' ) and then remembering all the cached assets in GUI itself (and automatically prevent duplicates to load). What do you think ?

src/ui/core/zcl_abapgit_gui_css_processor.clas.testclasses.abap
sbcgua: I've adjusted in the last commit

sbcgua: > I just didn't expect the asset manager to fail if an asset is empty.

Yeap, I actually have a doubt if this is correct at all :( On the other hand what is the point of empty assets ? Anyway let's reconsider it when there a real issue

PR #2775, opened 2019-07-07
Improve error information - HTML error message

christianguenter2:
Follow up on #2764 

#1998

Error messages are issued with html instead of `MESSAGE`.

html_error

Additionally the messages are enriched with two new functions.

Go to source
Navigates to the source code position where zcx_abapgit_exception=>raise or zcx_abapgit_exception=>raise_t100 is called

Callstack
Shows a popup with the full callstack and the possibility to navigate to
the source of each line with double click

I open this as a draft because there are still some rough edges and more testing is needed. Especially the UI/UX can be improved for sure. Feedback welcome.


src/zcx_abapgit_exception.clas.abap
```diff
@@ -2,17 +2,25 @@
 CLASS zcx_abapgit_exception DEFINITION
   PUBLIC
   INHERITING FROM cx_static_check
-  CREATE PUBLIC .
+  CREATE PUBLIC.
 
   PUBLIC SECTION.
+    CONSTANTS: gc_memory_id TYPE char30 VALUE `ZABAPGIT_CALLSTACK`,
```
larshp: the memory id can be deleted?

larshp:
suggest moving gc_section_text + gc_section_token from the exception class to a new message utility class, would like to keep the exception class small

larshp:
ZCL_ABAPGIT_MESSAGE_HELPER, try moving the DEFINITION DEFERRED manually above the exception in the generated report and check if that works

christianguenter2:
> ZCL_ABAPGIT_MESSAGE_HELPER, try moving the DEFINITION DEFERRED manually above the exception in the generated report and check if that works

Doesn't work
image

larshp:
I think ZCL_ABAPGIT_MESSAGE_HELPER can be fixed by doing some changes in `abapmerge`, but we can open this as a separate follow up task

sbcgua:
really cool feature ! :) 👍 

Some issues/suggestions

- the `m` hotkey does not seem to work - both to hide and to re-show the message

UI/UX suggestions (IMHO, can be a separate PR)

- Imho clicking the message panel is superfluous. Better to display details on hover and also in a more compact way (see below)
- this will also remove JS which does not work in java anyway so would be more reliable
- by the way, why message number 5 digits - it is just 3 or I'm missing something ? 😕 
- maybe `source` link should have a title with the last class/method - can be a quick check without need to drilldown + can be used for error screenshots (see below)

So some experiments:

2019-07-20_00h38_26

```html
    <div id="message" class="message-panel">
      There are no tags for this repository
      <div class="float-right"><a class="close-btn" href="#" onclick="toggleDisplay('message')">&#x274c;</a></div>
      <div class="float-right message-panel-commands">
        <a id="a_goto_message" href="sapevent:goto_message" title="ZCL_BLAH_BLAH->FIRE_ERROR">Message (001)</a>
        <a id="a_goto_source" href="sapevent:goto_source">Source</a>
        <a id="a_callstack" href="sapevent:show_callstack">Callstack</a>
      </div>
    </div>
```
```css
div.message-panel {
  z-index: 99;
  box-shadow: 2px 2px 4px 0px hsla(0, 0%, 0%, .1);
  padding: 12px;
  margin-left: -48%;
  position: fixed;
  bottom: 12px;
  width: 94%;
  left: 50%;

  border: 1px solid;
  border-radius: 5px;
  border-color: hsl(0, 42%, 64%);
  background-color: hsla(0, 42%, 90%, 1);
}

.message-panel-commands {
  display: none;
  margin-right: 0.5em;
}

.message-panel-commands a {
  padding: 0em 0.5em;
  border-left: 1px solid;
  border-left-color: #ccc;
}

.message-panel-commands a:first-child {
  padding-left: 0;
  border-left: none;
}

div.message-panel:hover .message-panel-commands {
  display: block
}
```

P.S. just by the way - no need to align tags in code, the html does the the indentation automatically. In fact if it is emplicit - it does not
```abap
ro_html->add( |<div> ... |
           && |  <div>| )
```
result

image

src/ui/core/zcl_abapgit_gui.clas.abap
```diff
@@ -306,9 +307,8 @@ CLASS ZCL_ABAPGIT_GUI IMPLEMENTATION.
       CATCH zcx_abapgit_cancel ##NO_HANDLER.
         " Do nothing = gc_event_state-no_more_act
       CATCH zcx_abapgit_exception INTO lx_exception.
-        IF mi_error_handler IS BOUND.
-          mi_error_handler->handle_error( lx_exception ).
-        ENDIF.
```
sbcgua: I'd vote against removing this part (and the local class in the factory). The point was to remove explicit ROLLBACK from GUI - it can potentially block reusing the GUI in other code

src/ui/zabapgit_js_common.w3mi.data.js
```diff
@@ -1360,3 +1361,22 @@ function GitGraphScroller() { // eslint-disable-line no-unused-vars
...
+}
+
+function toggleMessageDetail(oEvent){
+  if (oEvent &&  ( oEvent.target.id === "a_goto_source"
+                || oEvent.target.id === "a_callstack"
+                || oEvent.target.id === "a_goto_message") ) {
+    return;
+  }
+  toggleDisplay("message-detail");
+}
```
sbcgua: Guess the whole JS part is unnecessary after the last change ?

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
```diff
@@ -153,6 +158,100 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
...
+        iv_class = `close-btn`
+        iv_typ   = zif_abapgit_html=>c_action_type-onclick ).
+
+    ro_html->add( |</div>| ).
+
+    ro_html->add( |<div class="float-right message-panel-commands">| ).
+
+    IF lv_msgid_and_msgno IS NOT INITIAL.
+      ro_html->add_a(
+          iv_txt = lv_msgid_and_msgno
```
sbcgua: Imho, I propose keep "message" text near 00001 - makes it clearer for a user (btw why 5 digits ? :), in fact is the number needed at all ? Having that a direct link is there. Maybe just `message` ?

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
```diff
@@ -153,6 +158,100 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
...
+    ix_error->get_source_position(
+      IMPORTING
+        program_name = lv_program_name ).
+
+    ro_html->add_a(
+        iv_txt   = `Goto source`
+        iv_act   = zif_abapgit_definitions=>c_action-goto_source
+        iv_typ   = zif_abapgit_html=>c_action_type-sapevent
+        iv_title = |{ lv_program_name }|
+        iv_id    = `a_goto_source` ).
```
sbcgua: Currently it looks like e.g. `"ZCL_ABAPGIT_TAG_POPUPS========CP"`, I'd propose to remove suffix and add method name - would be more descriptive. (maybe a separate PR ...)

Here is an example of a comparable feature from one of my devs

```abap
method internal_error.
  data:
        l_prog      type string,
        l_meth      type string,
        sys_call    type sys_calls,
        t_sys_stack type sys_callst.

  call function 'SYSTEM_CALLSTACK'
    exporting
      max_level    = 2
    importing
      et_callstack = t_sys_stack.

  read table t_sys_stack into sys_call index 2.
  l_prog     = sys_call-progname.
  l_meth     = sys_call-eventname. " <<<< METHOD

  if sys_call-eventtype = 'METH'.
    data oref type ref to if_oo_class_incl_naming.
    data oclif type ref to if_oo_clif_incl_naming.

    call method cl_oo_include_naming=>get_instance_by_include
      exporting
        progname       = sys_call-progname
      receiving
        cifref         = oclif
      exceptions
        no_objecttype  = 1
        internal_error = 2.
    if sy-subrc is initial.
      oref ?= oclif.
      l_prog = oref->clskey-clsname. " <<< clean class name
    endif.
  endif.

  raise exception type xxxxxxxxx
    exporting
      textid    = m001_internal_error
      str_attr1 = l_prog
      str_attr2 = l_meth
      str_attr3 = text
      chr2      = code.

endmethod.
```

src/ui/zabapgit_css_common.w3mi.data.css
```diff
@@ -674,6 +674,50 @@ div.info-panel .version-marker {
...
+  border-radius: 5px;
+  border-color: hsl(0, 42%, 64%);
+  background-color: hsla(0, 42%, 90%, 1);
+}
+
+.message-panel .close-btn{
+  position: absolute;
+  right: 20px;
+  bottom: 10px;
+}
```
sbcgua: propose to remove `.message-panel .close-btn` style - `position absolute` makes it unrelated to the parent div. Guess the whole design might need some padding tuning at the end, but position absolute is too heavy :)

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
```diff
@@ -153,6 +158,100 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
...
+    ro_html->add_a(
+        iv_txt = `Callstack`
+        iv_act = zif_abapgit_definitions=>c_action-show_callstack
+        iv_typ = zif_abapgit_html=>c_action_type-sapevent
+        iv_id  = `a_callstack` ).
+
+    ro_html->add( |</div>| ).
+    ro_html->add( |<div class="message-panel-commands">| ).
+    ro_html->add( |{ lv_longtext }| ).
+    ro_html->add( |</div>| ).
```
sbcgua: `if lv_longtext is not initial` ? Also probably style should be named differently.

And ... I'd add a preformatted style for long text instead of inserting `<br>`s above. Maybe... It is not without pitfalls ...

```css
.message-panel-longtext {
  display: none;
  font-family: monospace;
  white-space: pre;
  padding-top: 0.5em;
  font-size: small;
  color: hsl(0, 0%, 40%);
}
div.message-panel:hover .message-panel-longtext {
  display: block
}
```

image

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
sbcgua: 5 digits: ah ! it is message class ! stupid me  ! :)

christianguenter2:
@sbcgua thanks a lot for your valuable comments. Unfortunately I don't have much time currently. I try respond till the weekend.

sbcgua:
@christianguenter2 no prob. Actually another option is to merge the PR and fix the rest later ( I can help with this ) - at end this feature is not about design at all :)

larshp:
@sbcgua I think if we can merge this and fix some of the stuff later it will be good, this PR is quite large. We can open an issue describing the future improvements/changes

sbcgua:
> @sbcgua I think if we can merge this and fix some of the stuff later it will be good, this PR is quite large. We can open an issue describing the future improvements/changes

Yeap, totally. Let's do it

PR #2779, opened 2019-07-09
Revert behavior of second back button

FreHu:
This only reverts #2679. I'm going to leave the buttons be.

As discussed on Slack, the three buttons are not consistent across themes, so what works in some may be strange in others.

image


sbcgua:
I'd leave `CLEAR mt_stack.` - this was a bug fix in fact.
P.S. pity about inconsistency, the idea was actually not bad.

flaiker:
Sorry for breaking this feature by the way :/

PR #2787, opened 2019-07-14
UI Dark Theme #2610

sandraros:
Proposal - Right now, only the first obvious screens were checked, and there are still a few issues with them.


flaiker:
Thanks! Can you attach some screenshots so it's easier to review?

sandraros:
image
image
image
image
image
image
image

sbcgua:
:+1: :+1: Cool !

I started to work on it too some time ago but lack of time ... :(

Opinion: maybe make borders a bit dimmer - they popup too brightly. imho.

Also: do you reuse variables from default theme or add new ones ? I didn't check. But just in case I'd suggest reusing existing ones **OR** adding to both files (probably the latter as there are a lot of colors remaining without variable for the moment). This would simplify CSS improvements in future

sandraros:
Yes, currently I gave border-color: #fff to select, input and textarea! I changed the values of default variables, and added one variable, but didn't adapt common.css and default.css. Almost all colors are redefined as it's very difficult to reuse the colors in an inverse theme. I don't have much ideas how to make the CSS reusable. Feel free to take the theme and rewrite it as you wish, I don't feel the courage to refactor ;-)

FreHu:
My few bits:
- The code looks quite nice, but I suggest slightly adjusting the background to the vscode grey `#1e1e1e`. 
- I agree the borders are too light, this gives me flashbacks to the eclipse dark theme
- It looks like the abapgit logo needs to have a transparent background

sbcgua:
> It looks like the abapgit logo needs to have a transparent background

Ut is because of explicit background color in css. The logo is transparent. Sadly, it seem that dark mode logo might need a dedicated image - IE does not support css filters :( `invert` would be not bad (this is CSS in Chrome :) ...

image

So some options to choose from

image

With the 1st being my favorite, and maybe 3rd is not bad.

sbcgua:
variations

image

sbcgua:
another idea about logo ... just split it into git vector logo and "abapgit" vector form and add both to the icon font. Thus it would enable regulating their colors from css (or the whole logo as an option but then the color must be same for both).

- git logo seems to be in CC license
- abapgit is just a text ... so should be ok

larshp:
@sbcgua can we merge this as is? And work from there? I dont think the current implementation will break anything?

sbcgua:
yes. The code is OK from my perspective.
BTW please write what do you think about the logo question above.

larshp:
I dont really have an opinion regarding the logo

Issue #2811, opened 2019-07-28
Improve error information - follow up


christianguenter2:
Follow up to #2775 and #1998.

- [x] Bug: Errorenous regex for message class and id. Best to get rid of it.
https://github.com/larshp/abapGit/blob/5081ddf7d08ec7cc02ec031cae0349763b123c9a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap#L163-L189
- [x] Refactor: Remove dynamic instantiation of ZCL_ABAPGIT_MESSAGE_HELPER somehow
https://github.com/larshp/abapGit/blob/5081ddf7d08ec7cc02ec031cae0349763b123c9a/src/zcx_abapgit_exception.clas.abap#L95-L114
- [x] Improve tooltip of "Goto source" link
![image](https://user-images.githubusercontent.com/17437789/62007471-0cb0f580-b14e-11e9-9afc-ea5e3cd7f9b0.png)
https://github.com/larshp/abapGit/blob/5081ddf7d08ec7cc02ec031cae0349763b123c9a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap#L233-L238
https://github.com/larshp/abapGit/pull/2775#discussion_r306390278
- [x] Check whether ROLLBACK should be decoupled from UI again.
https://github.com/larshp/abapGit/pull/2775#discussion_r305536671
- [x] Cleanup JS: Remove unnecessary functions (toggleDiplay, etc.)
https://github.com/larshp/abapGit/pull/2775#discussion_r306387234
- [x] Add tooltip for message class and id (e.g. raw message?)
![image](https://user-images.githubusercontent.com/17437789/62007542-ee97c500-b14e-11e9-8413-b701df0a661b.png)
https://github.com/larshp/abapGit/blob/5081ddf7d08ec7cc02ec031cae0349763b123c9a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap#L221-L227
- [x] Refactor CSS
https://github.com/larshp/abapGit/pull/2775#discussion_r306391068
- [x] Beautify message panel
https://github.com/larshp/abapGit/pull/2775#discussion_r306398751
- [x] Improve error information in log popup as well
- [x] Add icon in front of error message
![image](https://user-images.githubusercontent.com/17437789/62365931-bcf37500-b525-11e9-8c8e-af2fa3e236a8.png)

mbtools:
[...]

PR #2825, opened 2019-07-30
Completely dynamic JS hints

sbcgua:
to #2811

@christianguenter2 

draft for the moment, might be some polishing needed but works and ready for review

- hints are completely generated in JS, so no auxilliary `<span class="tooltiptext hidden">`
- some terminology fixes (tooltip vs hint) and some css
- I changed the JS code quite heavily ... sorry ... appeared to be an interesting subject, carried away :) ... hopefully for good.
- `ZCL_ABAPGIT_HTML ` - just removed the tooltip span, others are se80 reorderings
- same for ZCL_ABAPGIT_GUI_PAGE - just a rename in JS function
- known issue: a dropdown displayed with hint cannot be hidden ... actually no change from existing logic ... not sure if possible to fix cleanly ... maybe with timeout
- known issue: I added backspace for reverting hint steps back ... but works only on the first page - after IE perceives it as back button :( not sure how to fix

Doubt: why setting color in settings ? Imho it should depend on theme directly. Didn't change but imho not quite consistent.

### DIFF
--- a/src/ui/core/zcl_abapgit_html.clas.abap
+++ b/src/ui/core/zcl_abapgit_html.clas.abap
@@ -24,7 +24,6 @@ CLASS zcl_abapgit_html DEFINITION
         !iv_class TYPE string OPTIONAL.
   PROTECTED SECTION.
   PRIVATE SECTION.
-    CONSTANTS: co_span_link_hint TYPE string VALUE `<span class="tooltiptext hidden"></span>`.
     CLASS-DATA: go_single_tags_re TYPE REF TO cl_abap_regex.
 
     DATA: mt_buffer TYPE string_table.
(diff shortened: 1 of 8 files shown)


sbcgua:
P.S. I think JS file should be splitted in parts ... maintenance is quite inconvenient by now

larshp:
what is hints?

sbcgua:
The keyboard navigation feature Christian mentioned in #2775 
image

larshp:
any impacts on performance with these changes?

sbcgua:
No sure how to measure. In theory it should be faster because of pointers caching. But not sure how to measure it properly, especially in IE. I'm open for suggestions :)

christianguenter2:
> Doubt: why setting color in settings ? Imho it should depend on theme directly. Didn't change but imho not quite consistent.

Agree. I guess there's the option to set the color in the chrome vimium plugin. I thought it would be cool to have the same in abapGit. But to be honest, I didn't change the standard color since then. And we hadn't themes when I implemented this ;-)

christianguenter2:
Did some tests. LGTM from a functional point of view.

christianguenter2:
> I changed the JS code quite heavily ... sorry ... appeared to be an interesting subject, carried away :) ... 

Sure, no problem at all :-) I can learn a lot.

christianguenter2:
> known issue: a dropdown displayed with hint cannot be hidden ... actually no change from existing logic ... not sure if possible to fix cleanly ... maybe with timeout

I can be hidden with the hint. That's ok for me.

christianguenter2:
> any impacts on performance with these changes?
Didn't notice any negative impact during my tests.

FreHu:
@sbcgua
>No sure how to measure. In theory it should be faster because of pointers caching and less dynamic dom changes. But not sure how to measure it properly, especially in IE. I'm open for suggestions :)

You can profile using iechooser, although you need to export it and then import it in a normal IE window.

larshp:
looks like the JS is only run when hotkeys are enabled, good,
image

sbcgua:
@christianguenter2 

- color: ok, if you don't mind I'll remove it then :) 
- what do you think about backspace feature ? did you try ? I thought it would be nice to undo typings ... but since it works only on the first page (most frequently the tutorial ...) it does not make sense :( I'm considering removing it
- "dropdown can be hidden with the hint" - hmm, you mean remembering the last triggered dropdown and clearing it on hotkey press ? Interesting idea. Thought can be a bit tricky - I'll have a look. P.S. sadly `li` does not support `:focus` state - it would solve the issue cleanly, just click anywhere and focus changes together with CSS

sbcgua:
@FreHu @larshp did some performance tests, with good old `console.time`

Scenario: init (obviously), show all hints, hide all hints, show all again, type 146 for filtering (each time less items to filter), total number of hints 3466. last 2000 hints were one above another ... made a separate test to check if block elements would be different but the difference seems to be minor.

Stages below named differently for new and old logic, but follow the same order - it has the same meaning

IE
```
New logic
deployHintContainers: 421ms
displayHints true:    489,9ms
displayHints false:   152,5ms
displayHints true:    441,4ms
filterHints 1:        283,8ms
filterHints 4:        59ms
filterHints 6:        44,2ms

Old logic
querySelectorAll:       2ms
toggleAllTooltips true: 471,2ms
toggleAllTooltips false: 401,8ms
toggleAllTooltips true: 399,2ms
filterTooltips 1:       326,7ms
filterTooltips 4:       77,6ms
filterTooltips 6:       44,8ms
```

Initialization is obviously taking more time in the new logic as it creates elements. Though half a second per 3,5k hints ... seems to be OK for me. Displaying, surprisingly for me, is a bit longer then in old one. Seems like because internal structure of the hint in the new logic is more complex (2 independent spans) - made a quick test for this. Hiding is faster in new logic. And also filtering (probably because of those 2 independent spans - back side of the coin - that was why I actually made them). But in general - same figures imho.

Now to extend the study, let's run it in Chrome

```
New logic
deployHintContainers: 17.22ms
displayHints true:    19.35ms
displayHints false:    9.38ms
displayHints true:    18.66ms
filterHints 1:        22.15ms
filterHints 4:        10.61ms
filterHints 6:         8.61ms

Old logic

querySelectorAll:         0.21ms
toggleAllTooltips true:  54.86ms
toggleAllTooltips false: 68.09ms
toggleAllTooltips true:  35.89ms
filterTooltips 1:       127.24ms
filterTooltips 4:        72.48ms
filterTooltips 6:       105.40ms
```
Yeah ... was is the same PC ? :)

christianguenter2:
> * color: ok, if you don't mind I'll remove it then :)
okay

> * what do you think about backspace feature ? did you try ? I thought it would be nice to undo typings ... but since it works only on the first page (most frequently the tutorial ...) it does not make sense :( I'm considering removing it
Yeah, if I misstype something I press the activation hotkey twice and start over again.

> * "dropdown can be hidden with the hint" - hmm, you mean remembering the last triggered dropdown and clearing it on hotkey press ? Interesting idea. Thought can be a bit tricky - I'll have a look. P.S. sadly `li` does not support `:focus` state - it would solve the issue cleanly, just click anywhere and focus changes together with CSS

Note quite sure whether we are talking about the same thing. I mean this:
abapGit5

sbcgua:
OK, committed the discussed changes.

> Dropdown logic

Ah, OK, I see what you mean. This should also work :)
But I also added some minor improvements in the last commits. 
- if you force-close the hints (second "f"), and a dropdown was displayed - it is closed automatically. 
- Also if you open a dropdown, and show hints again (like in your gif), and mistype the command, then all hints are closed (it was so before) and also opened dropdown closes too.

I release the PR from draft, imho it is complete. But I'll be away for a week, so no bugcare :)) Maybe wait a bit with the merge if you prefer :))

christianguenter2:
Sorry for the force push. Somehow did something wrong while merging master.

christianguenter2:
Small bug. When a page is newly rendered, hint on a dropdown doesn't work the first time.
abapGit6

I try to fix it.

christianguenter2:
#2807

PR #2856, opened 2019-08-08
Allow horizontal scrolling in docs code blocks

flaiker:
https://docs.abapgit.org/development/api.html

Before:
image

After:
image


sbcgua:
I faintly remember that there was some issue with scrolling. Maybe on small screens or something and i disabled it. Cannot check now. Just for info. Maybe there will be some UX effect.

PR #2894, opened 2019-08-26
APACK: Introduce versioned dependencies

### LINKED ISSUE #2893: APACK: Export versioned dependencies
APACK currently doesn't feature any possibility to declare version numbers for dependencies. Both strict, half-open and unbounded version ranges (see also https://dzone.com/articles/manifestmf-and-featurexml ) should be supported to enable import tools (such as the abapGit ADT plugin) to tell the user if there are inconsistencies or conflicts.

SebastianWolf-SAP:


SebastianWolf-SAP:
This addresses #2893

PR #2895, opened 2019-08-26
Diff/Patch: Add Jump to file

### LINKED ISSUE #2890: Patch - jump to file
Frequently requested by colleagues - in the patch page, add a dropdown which shows all filtered files and allows you to scroll to that file when clicked.

As a workaround you can currently use ctrl+f, but this can cause significant lag due to the amount of elements on the page.

christianguenter2:
resolves #2890 

First try implementing jump for Diff/Patch page.
image

Maybe someday we can add a CTRL+P like file navigation.

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -21,6 +21,7 @@
 /* exported toggleRepoListDetail */
 /* exported onTagTypeChange */
 /* exported getIndocStyleSheet */
+/* exported addMarginBottom */
 
 /**********************************************************
  * Polyfills
@@ -665,6 +666,11 @@ DiffHelper.prototype.highlightButton = function(state) {
     this.dom.filterButton.classList.remove("bgorange");
   }
 };
+
+// Add Bottom margin, so that we can scroll to the top of the last file
+function addMarginBottom(){
+  document.getElementsByTagName("body")[0].style.marginBottom = screen.height + "px";
+}
 
 /**********************************************************
  * Other functions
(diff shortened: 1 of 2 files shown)


christianguenter2:
@FreHu Thanks for your feedback. I tried to address both of your concerns. Can you retry?

FreHu:
Works perfectly. I got another idea and hacked together a floating menu, not sure where best to put it so it doesn't cover text.
image

christianguenter2:
Nice idea! I also thought about fixing the header somehow. But then it would probably be fix everywhere in abapGit. Not sure whether we want that.

FreHu:
Probably not, at least without testing (although diff for example is pretty much the same page). 
Maybe you could define the style in a class and toggle it on the correct pages.

FreHu:
Or maybe just fix the buttons and not the `logo > diff`.

FreHu:
Today I discovered this already has good keyboard navigation 
- click the jump link to focus
- select file with up, down arrows
- enter to jump. the jump list stays open
- home to go back to jump list

Disadvantages
- you can't scroll with arrow keys if the jump list is focused
- sometimes the focus gets messed up when you click something else and trying to refocus the jump list is a bit annoying

christianguenter2:
> enter to jump. the jump list stays open

Hm, enter doesn't work for me but spacebar does.

FreHu:
Maybe ctrl+p could work instead of the floating menu. Maybe merge it and see how things evolve? My colleagues like it as it is already.

sbcgua:
+1 for merge. The feature is good and useful. Ctrl+p is not yet finalized anyway.

2 minor comments:
- I'd rename Jump to something else (Jump to file, or better Go to file ?) because Jump itself is confusing imho
- after file is selected the jump list scrolls away - its a potential stop for future improvements. Options are: keyboard shortcut or header that follows scrolling or some kind of hanging button at the corner. Or combination.

PR #2897, opened 2019-08-29
Command palette (Ctrl+P): Go to repo ...

sbcgua:
@FreHu @christianguenter2 
Speaking about Ctrl+P (#2895) ... - a good idea ! :)
Implemented for keyboard repo navigation ("Go to repo ...")
Key combination: F2 (with the prospect to use F1 for abapGit actions palette)
~~Draft but ready for beta-testing among highly-privileged beta members~~ 😜  ( = all interested) 

ezgif com-video-to-gif

Features:
- `CommandPalette` class is quite abstract and can be reused, see below
- arrow navigation works, with auto scrolling
- auto patch hotkeys help sheet (and reusable util `Hotkeys.addHotkeyToHelpSheet`)
- accepts clicks and enter
- fuzzy search ("abc" would find "**abc**efg" and "**a**1**b**2**c**3") (fuzzy search function is also separated and reusable outside of the class). Search is case insensitive.
- optional icons
- Backspace in empty filter closes the palette
- `^` before the `toggleKey` means Ctrl (e.g. `^g` = Ctrl+g)

Currently a draft, known issues/todo:
- ~~search is case sensitive~~
- ~~g cannot be typed - closes the popup - probably needs a ctrl combination.~~
- How to stop sap from closing on Esc ?!
- ~~seems like some conflicts with `enableArrowListNavigation`~~

Reuse guide
- to activate run `var xxx = new CommandPalette(enumerateTocAllRepos, {toggleKey: "^g", hotkeyDescription: "Go to repo ..."});`
- where `g` (ctrl+g in this case) and `Go to repo` are the hotkey and it's description
- and `enumerateTocAllRepos` - a function returning array of commands in form of
- there can be several commandPalettes on a page
```js
{
  action, // sapevent action without "sapevent:" prefix
  iconClass, // OPTIONAL, "icon icon_x ..."
  title, // command text - "my command X"
}
```

Maybe in future: potentially possible to bind the popup to around of specific element, if required - e.g. the case with diff file jump. So that it is located more consistently.

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -1492,8 +1492,7 @@ function CommandPalette(commandEnumerator, opts) {
     ul:      null,
     input:   null
   };
-  // this.selectedItem      = null;
-  // this.selectIndex       = 0;
+  this.selectIndex       = -1; // not selected
   this.filter            = "";
   this.renderAndBindElements();
   this.hookEvents();
@@ -1545,20 +1544,16 @@ CommandPalette.prototype.handleToggleKey = function(event){
 };
 
 CommandPalette.prototype.handleInputKey = function(event){
-  // if ((event.key === "ArrowUp" || event.key === "Up") && this.selectIndex > 0) {
-  //   this.selectIndex--;
-  //   this.highlight();
-  // } else if ((event.key === "ArrowDown" || event.key === "Down") && this.selectIndex < this.commands.length - 1) {
-  //   this.selectIndex++;
-  //   this.highlight();
-  // } else if (event.key === "Enter") {
-  //   this.exec(this.selectedItem);
-  // } else
-  if (this.filter !== this.elements.input.value) {
-    // this.selectIndex = 0;
+  if (event.key === "ArrowUp" || event.key === "Up") {
+    this.selectPrev();
+  } else if (event.key === "ArrowDown" || event.key === "Down") {
+    this.selectNext();
+  } else if (event.key === "Enter") {
+    this.exec(this.getSelected());
+  } else if (this.filter !== this.elements.input.value) {
     this.filter = this.elements.input.value;
     this.applyFilter();
-    // this.highlight();
+    this.selectFirst();
   }
   event.preventDefault();
 };
@@ -1569,7 +1564,6 @@ CommandPalette.prototype.applyFilter = function(){
     if (!this.filter) {
       cmd.element.style.display = "";
       cmd.titleSpan.innerText   = cmd.title;
-      // cmd.element.classList.remove("selected");
     } else {
       var matchedTitle = fuzzyMatchAndMark(cmd.title, this.filter);
       if (matchedTitle) {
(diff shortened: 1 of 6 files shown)


sbcgua:
@christianguenter2 could you please tell what was the intention behind `enableArrowListNavigation` ? Do I understand correctly that it is to enable keyboard navigation for Vimium hints ? So should be enabled just when hints are active and displayed ?

christianguenter2:
Idea is to activate to control dropdown menus completely via keyboard. Like this.
abapGit7

If I remember correctly, this is also useful in other, non linkhint related, circumstances.

sbcgua:
Problem is that it acts globally and it does not check if it is in any navigation mode or something. Just select the first available li with ".selected" child or, if not found, **any** element after actual active element. E.g. it sets "selected" class for UL element in my test and tries to focus a span.
I'm thinking how to fix it ... and that why I ask what was the design :)

sbcgua:
@christianguenter2 
did some refactoring on KeyNavigation, could you plz have a look ?

Expected behavior:
- navigates lists
- enter toggles (open/close) dropdowns or clicks the links
- backspace closes opened dropdowns (parent or right-below) + focuses it's trigger anchor
- should work with multi level - do we have those ???

sbcgua:
UX: appears that SAP GUI bypasses F1 and F2 to IE. Hmmm ??? 
F1 -> CP for actions, F2 -> CP for go to repo ???

sbcgua:
OK, I think ready for review. Updated the text at the beginning with some details.
Use `F2` to trigger (also `?` to see the cheat sheet with the assigned key)

sbcgua:
Expect right after :) (https://github.com/sbcgua/abapGit/tree/command-palette-more)
image

src/ui/zabapgit_js_common.w3mi.data.js
```diff
@@ -1407,3 +1438,253 @@ function GitGraphScroller() { // eslint-disable-line no-unused-vars
...
+  if (element.nodeName === "SPAN") element = element.parentNode;
+  if (element.nodeName === "I") element = element.parentNode;
+  if (element.nodeName !== "LI") return;
+  this.exec(this.getCommandByElement(element));
+};
+
+CommandPalette.prototype.exec = function(cmd) {
+  if (!cmd) return;
+  this.toggleDisplay(false);
+  submitSapeventForm(null, cmd.action);
```
christianguenter2: What if we have another action than sapevent? E.g. jump to file uses plain js function.

christianguenter2:
Great work so far! Imho it would be nice to have customizable hotkeys for the command palette(s).

christianguenter2:
Expect right after :)

https://github.com/christianguenter2/abapGit/tree/command-palette-jump
image

christianguenter2:
KeyNavigation works as expected.

src/ui/zabapgit_js_common.w3mi.data.js
christianguenter2: Maybe something like this?
https://github.com/sbcgua/abapGit/blob/6f46aa73892383bd80d98012c2d649bf5ec5eef3/src/ui/zabapgit_js_common.w3mi.data.js#L1669-L1673

sbcgua: Yes, I though about it but was lazy and decided to postpone
and now I see you implemented that in your jump CP :)
P.S. I just suggest not passing title to the action and `bind` it in enumerator instead - I'll leave the comment in your commit

sbcgua:
"Expect right after :)" 👍 👍 👍 

customizable hotkeys: maybe, but it would be quite hacky at the moment as it should be set in abap and then re-used in js. Probably will need some refactoring.

src/ui/zabapgit_js_common.w3mi.data.js
christianguenter2: Thx, fixed it in my branch.

christianguenter2:
It's really a pitty that we can't use the CTRL+P hotkey, because there is no way to disable the print popup.

sbcgua:
Yeap. However F1 and F2 are also fine. In fact `F1 = Ctrl+P` in vscode (both work by default) so at least some consistency with popular external tools. By the way I was surprised as F1 in intercepted by Chrome for example and Sap passes it by.

Issue #1906, opened 2018-09-09
Support for Object Type UENO - Entity Type


d023604:
Support for the serialisation and de-serialisation of the Data Modeller (Tx SD11) object types UMDO - Data Model, UENO - Entity Type presently missing. Whilst perhaps not the most commonly used objects, relatively simply objects to implement.

d023604:
Happy also to implement - have started already.

larshp:
@d023604 are you actively working on this? if not I will remove your assignment 😄

d023604:
Hi Lars,

Thanks for the nudge. In short - maintain the assignment - and I will
complete it in the next week.

Regards
Andrew

d023604:
Testing completed. Pull request shortly

larshp:
merged in https://github.com/larshp/abapGit/pull/2911, closing

PR #2911, opened 2019-09-11
UENO: Entity Type Implementation (#1906)

d023604:
Initial implementation of UENO - Entity Type
Update of the documentation of supported objects


d023604:
small example added to UENO object

PR #2933, opened 2019-09-28
Order by on repo page

### LINKED ISSUE #2852: Add order by function to repo page

christianguenter2:
resolves #2852 
follow up on #2877

abapGit8


sbcgua:
cool :+1: 
Imho: is the reset order button really needed ? There is no such thing as "completely unordered list" - I think it would be fine just to always leave sorting , especially if it is based on user preference. Also imho that crossed circle looks a bit ... out of design. I'm trying to remember any site where there was a separate reset order function and I cannot remember any.

christianguenter2:
Yeah, the design of the reset button is questionable. It was the best I found ;-)
The button is imho needed, because the diffs are rendered on the top by default. If we order the list there's otherwise no way back the original layout.

abapGit9

sbcgua:
Hmmm, but that means that you don't want to reset, you want to order by "diffs first" ... I think there might be more obvious representation for this. E.g. adding at the end the column (the name of the column where diff status is) something like "Diffs" or "Diffs first". From usability perspective it would be much clearer.

This by the way opens a deeper question:
- may I want to see diffs first and also ordered by diff status ?
- may I want to see diffs first and yet sorded by type ? (I mean maybe diff first is a separate view param. maybe, not sure, just thinking aloud)

christianguenter2:
Yeah, I don't want to open that box yet.

sbcgua:
mmm

image

also non code files go to the bottom ... is it intended ?

image

src/ui/zcl_abapgit_gui_view_repo.clas.abap
```diff
@@ -761,12 +787,16 @@ CLASS ZCL_ABAPGIT_GUI_VIEW_REPO IMPLEMENTATION.
         ENDIF.
 
         " Repo content table
-        ro_html->add( '<table class="repo_tab">' ).
+        ro_html->add( '<table class="repo_tab db_tab">' ).
```
sbcgua: `db_tab` ... mmm ... better to add css for repo_tab
```css
/* common: */
.repo_tab th {
  text-align: left;
  padding: 0.5em;
  border-bottom: 1px solid;
  font-weight: normal;
}
/* theme-default: */
.repo_tab th {
  color: #888888;
  border-bottom-color: #ddd;
}

```

sbcgua:
I mean something like this

image

```html
<th class="cmd">
  <a href="sapevent:toggle_diff_first"><i class="icon icon-check blue" ></i>diffs first</a>
</th>
```
blue or grey depending on activation

```css
.repo_tab th.cmd .icon{
  padding-right: 8px;
}
/* added ", .repo_tab th.cmd" */
.repo_tab td.cmd, .repo_tab th.cmd {
  text-align: right;
  padding-left: 0.5em;
  padding-right: 0.7em;
}
```

christianguenter2:
@sbcgua thank you very much for your valuable feedback. I tried to apply your suggestions. This is the current state.

abapGit10

TODO: 

- [ ]  Order by `PATH`

image

- [ ]  There are cases where it's not possible to restore the original sort order as calculated here.
https://github.com/larshp/abapGit/blob/22d19023e40d45315a876a73e33fd0da710c5ece/src/zcl_abapgit_repo_content_list.clas.abap#L255-L258

sbcgua:
Looks very nice and works like a charm (tested), thanks ! 👍 

P.S. Cannot judge about path - I always just hide it :)
P.P.S. hmmm the titles strangely jump when sorting switched. Seems like this is because of the this symbol, but it does not happen in Chrome. Seems like some bug of IE - reproducible if started from desktop too. Well, not sure if can be fixed.

christianguenter2:
We can merge the current state. Above mentioned issues are minor and can be adressed later if needed.

PR #2977, opened 2019-10-16
Add object type IAXU

CASAG-MariusRaht:
#2889 - Object type ignored, not supported: IAXU

Was able to test it with source code on a customer system and it seems fine. But it would be better if someone could validate this solution!


christianguenter2:
Can you add a testcase here?
https://github.com/abapGit-tests/IAXU

src/zabapgit.prog.xml
```diff
@@ -17,6 +17,18 @@
...
+     <LANGUAGE>D</LANGUAGE>
+     <TEXTPOOL>
+      <item>
+       <ID>R</ID>
+       <ENTRY>abapGit</ENTRY>
+       <LENGTH>7</LENGTH>
+      </item>
+     </TEXTPOOL>
+    </item>
+   </I18N_TPOOL>
```
christianguenter2: Can you remove the german text from the pull request?

CASAG-MariusRaht:
Found an error in the read method: cl_w3_api_template=>if_w3_api_template~load is only compatible with IATU - have to look into it more closely.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,231 @@
...
+
+    DATA: lt_source   TYPE w3htmltabtype,
+          lv_source   TYPE string,
+          li_template TYPE REF TO if_w3_api_template.
+
+
+    cl_w3_api_template=>if_w3_api_template~create_new(
+      EXPORTING p_template_data = is_attr
+                p_program_name = is_attr-programm
+      IMPORTING p_template = li_template ).
```
christianguenter2: What happens if the object already exists?

CASAG-MariusRaht:
Closed - have to make further changes. Will be reopened when I'm ready.

PR #2978, opened 2019-10-16
Add object type IAXU

mariusraht2:
#2889 - Object type ignored, not supported: IAXU
#2977 - First pull request

Figured out how to create IAXU objects (s. https://github.com/abapGit-tests/IAXU) and rewrote the whole class on base of my tests.

@christianguenter2: Regarding your question in the first pull request: If the object already exists an exception is being thrown which leads to a message on the GUI. Just copied this behaviour from other object classes.


christianguenter2:
Yeah, problem is that deserialization is not only for creating objects but also for update (e.g. if remote changed or to reset local). That would currently fail for IAXU. Strategy in this case is usually to either update the object or delete and recreate it. Maybe you can incorporate something in that direction.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,216 @@
+CLASS zcl_abapgit_object_iaxu DEFINITION PUBLIC INHERITING FROM zcl_abapgit_objects_super FINAL.
+
+  PUBLIC SECTION.
+    INTERFACES zif_abapgit_object.
+
+  PROTECTED SECTION.
+  PRIVATE SECTION.
+    DATA: mo_xml_api           TYPE REF TO cl_w3_api_xml3,
```
christianguenter2: suggest to remove mo_xml_api and always use a local variables instead as in `delete`.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,216 @@
...
+ENDCLASS.
+
+
+
+CLASS zcl_abapgit_object_iaxu IMPLEMENTATION.
+
+
+  METHOD read.
+
+    DATA: li_service TYPE REF TO cl_w3_api_xml3,
```
christianguenter2: li_service is not used. Suggest to remove it.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,216 @@
...
+        error_occured       = 4
+        OTHERS              = 5 ).
+
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( 'error from w3_api_xml3~load' ).
+    ENDIF.
+
+    lo_xml_api->if_w3_api_object~set_changeable( abap_true ).
+    lo_xml_api->if_w3_api_object~delete( ).
+    lo_xml_api->if_w3_api_object~save( ).
```
christianguenter2: These methods raise a bunch exceptions. Suggest to handle them.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,216 @@
...
+        object_already_existing   = 3
+        not_authorized            = 4
+        action_cancelled          = 5
+        OTHERS                    = 6 ).
+
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( |Error from w3_api_xml3~create_new| ).
+    ENDIF.
+
+    mo_xml_api->if_w3_api_object~save( ).
```
christianguenter2: The method raises a bunch exceptions. Suggest to handle them.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,216 @@
...
+        p_attributes        = es_attr
+      EXCEPTIONS
+        object_not_existing = 1
+        permission_failure  = 2
+        data_corrupt        = 3
+        error_occured       = 4
+        OTHERS              = 5 ).
+
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( 'error from w3_api_xml3~load' ).
```
christianguenter2: Maybe add the subrc to the exception text? 
```
      zcx_abapgit_exception=>raise( |error from w3_api_xml3~load subrc={ sy-subrc }| ).
```

CASAG-MariusRaht:
Thanks for the feedback - will look into asap.

src/objects/zcl_abapgit_object_iaxu.clas.abap
CASAG-MariusRaht: I agree. Many other classes (e.g. IATU or IASP to just name some) are to be reworked then too, right?

CASAG-MariusRaht:
Maybe it would be useful to encapsulate the xml api methods into local methods to reduce the lines of redundant source code due to multiple calls if these api methods?

src/objects/zcl_abapgit_object_iaxu.clas.abap
CASAG-MariusRaht: Opened issue #2982

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,280 @@
...
+        p_xml               = lo_xml_api
+      EXCEPTIONS
+        object_not_existing = 1
+        permission_failure  = 2
+        data_corrupt        = 3
+        error_occured       = 4
+        OTHERS              = 5 ).
+
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( 'error from w3_api_xml3~load subrc={ sy-subrc }' ).
```
christianguenter2: needs to be a string template, also others below
```suggestion
      zcx_abapgit_exception=>raise( |error from w3_api_xml3~load subrc={ sy-subrc }| ).
```

christianguenter2:
> Maybe it would be useful to encapsulate the xml api methods into local methods to reduce the lines of redundant source code due to multiple calls of these api methods?

Feel free

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,305 @@
...
+
+    TRY.
+        w3_api_load( is_name = ls_name ).
+
+        rv_bool = abap_true.
+
+      CATCH zcx_abapgit_object_not_found.
+        rv_bool = abap_false.
+
+    ENDTRY.
```
christianguenter2: Can't import test repo
image

`zcx_abapgit_not_found=>raise` raises exception of type zcx_abapgit_exception  w3_api_load which isn't caught here.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,305 @@
...
+        permission_failure  = 2
+        data_corrupt        = 3
+        error_occured       = 4
+        OTHERS              = 5 ).
+
+    IF sy-subrc = 1.
+      zcx_abapgit_object_not_found=>raise( |Error from w3_api_xml3~load subrc={ sy-subrc }| ).
+    ELSEIF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( |Error from w3_api_xml3~load subrc={ sy-subrc }| ).
+    ENDIF.
```
christianguenter2: Hm, we already have `zcx_abapgit_not_found` exception class. Is the new one `ZCX_ABAPGIT_OBJECT_NOT_FOUND` really needed? I would prefer to use plain `zcx_abapgit_exception` as it simplifies the logic. See bug above.

mariusraht2: If I would use ZCX_ABAPGIT_EXCEPTION I couldn't catch it in the method for the existence check and set the boolean to abap_false. Additionally ZCX_ABAPGIT_NOT_FOUND doesn't inherit from ZCX_ABAPGIT_EXCEPTION and therefore hasn't a method RAISE which would allow me to throw a message up to the GUI.

mariusraht2: Ah, now I understand... Method Raise in ZCX_ABAPGIT_OBJECT_NOT_FOUND doesn't throw ZCX_ABAPGIT_OBJECT_NOT_FOUND but ZCX_ABAPGIT_EXCEPTION and therefore isn't being catched. Ok, I'll change it according to your suggestions.

src/objects/zcl_abapgit_object_iaxu.clas.abap
mariusraht2: Fixed it

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,309 @@
...
+    ls_name = ms_item-obj_name.
+
+    w3_api_load( EXPORTING is_name = ls_name
+                 IMPORTING eo_xml_api = lo_xml_api ).
+
+    IF lo_xml_api IS BOUND.
+      rv_bool = abap_true.
+    ELSE.
+      rv_bool = abap_false.
+    ENDIF.
```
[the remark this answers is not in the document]
mariusraht2: I can't do that in the exists-method because some exceptions (sy-subrc > 1) have another meaning than "object is missing" (sy-subrc = 1). If I catch ZCX_ABAPGIT_EXCEPTION I don't know what happened in the underlying method which throws the exception. I could analyse the exception text which holds sy-subrc but that seems a bit too much. Am I missing something?

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,309 @@
...
+        object_not_existing = 1
+        permission_failure  = 2
+        data_corrupt        = 3
+        error_occured       = 4
+        OTHERS              = 5 ).
+
+    IF    sy-subrc <> 1
+      AND sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( |Error from w3_api_xml3~load subrc={ sy-subrc }| ).
+    ENDIF.
```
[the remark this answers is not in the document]
mariusraht2: s. comment above

src/objects/zcl_abapgit_object_iaxu.clas.abap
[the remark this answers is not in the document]
mariusraht2: I've implemented an alternative approach. Also thanks for your continuous reviews and suggestions! :)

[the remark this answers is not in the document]
christianguenter2: Imho it would be sufficient to interpret every catched exception as `rv_bool = abap_false` and not differentiate, but I agree that's not the cleanest solution.

What about using `CL_W3_API_XML3=>S_CHECK_EXIST` method instead of ` W3_API_LOAD`? Then we can get rid of `EV_SUBRC`  and simplify the exception handling in `EXISTS` further.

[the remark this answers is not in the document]
mariusraht2: That simplifies a lot..

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,299 @@
...
+
+  ENDMETHOD.
+
+
+  METHOD w3_api_set_changeable.
+
+    io_xml_api->if_w3_api_object~set_changeable(
+      EXPORTING
+        p_changeable                 = abap_true
```
christianguenter2: Use importing parameter `iv_changeable` instead of `abap_true`?

christianguenter2:
If I install the demo repo and imediately uninstall I get this error:
image

Suggest to add 
```   
 w3_api_set_changeable( io_xml_api    = lo_xml_api
                           iv_changeable = abap_false ).
```
at the end of `zif_abapgit_object~deserialize` to unlock.

mariusraht2:
> 
> 
> If I install the demo repo and imediately uninstall I get this error:
> image
> 
> Suggest to add
> 
> ```
>  w3_api_set_changeable( io_xml_api    = lo_xml_api
>                            iv_changeable = abap_false ).
> ```
> 
> at the end of `zif_abapgit_object~deserialize` to unlock.

I would say it belongs to the local save-method because after every change it has to be unlocked.

mariusraht2:
> 
> 
> If I install the demo repo and imediately uninstall I get this error:
> image
> 
> Suggest to add
> 
> ```
>  w3_api_set_changeable( io_xml_api    = lo_xml_api
>                            iv_changeable = abap_false ).
> ```
> 
> at the end of `zif_abapgit_object~deserialize` to unlock.

Added it to local method "save" because we would have to unlock it always after changes were made and saved.

src/objects/zcl_abapgit_object_iaxu.clas.abap
```diff
@@ -0,0 +1,302 @@
...
+
+
+  METHOD zif_abapgit_object~exists.
+
+    DATA: ls_name  TYPE iacikeyt,
+          lv_subrc TYPE sysubrc,
+          lx_exc   TYPE REF TO zcx_abapgit_exception.
+
+
+    ls_name = ms_item-obj_name.
```
christianguenter2:
```suggestion
    ls_name-service = ms_item-obj_name.
```
would make it more obvious what's going on.

christianguenter2:
@mariusraht2 
one last small remark ;-)

mariusraht2:
> 
> 
> LGTM

Thanks for the help :)

PR #2981, opened 2019-10-17
DOMA: clear VALPOS

ibnbr:
issue #1053

### DIFF
--- a/src/objects/zcl_abapgit_object_doma.clas.abap
+++ b/src/objects/zcl_abapgit_object_doma.clas.abap
@@ -356,6 +356,7 @@ CLASS ZCL_ABAPGIT_OBJECT_DOMA IMPLEMENTATION.
           lv_masklen TYPE c LENGTH 4,
           lt_dd07v   TYPE TABLE OF dd07v.
 
+    FIELD-SYMBOLS <ls_dd07v> TYPE dd07v.
 
     lv_name = ms_item-obj_name.
 
@@ -391,6 +392,10 @@ CLASS ZCL_ABAPGIT_OBJECT_DOMA IMPLEMENTATION.
     SORT lt_dd07v BY
       valpos ASCENDING
       ddlanguage ASCENDING.
+
+    LOOP AT lt_dd07v ASSIGNING <ls_dd07v>.
+      CLEAR <ls_dd07v>-valpos.
+    ENDLOOP.
 
     io_xml->add( iv_name = 'DD01V'
                  ig_data = ls_dd01v ).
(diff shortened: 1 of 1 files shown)


larshp:
closes #1053

larshp:
please open PR with changed files to https://github.com/abapGit-tests/DOMA

ibnbr:
this object has no value range, so I have no changes to open a PR. can I create de value range and open a PR?

larshp:
https://github.com/abapGit-tests/DOMA_fixed_values then

ibnbr:
PR Done.
I had to make some adjustments.

larshp:
Suggest also clearing DD07V-DOMNAME, as it is always the same as the object name

ibnbr:
I made some changes, and I'm doing some tests, when DOMA has translation is not working properly. I don't know if it was caused by my changes, I'll investigate.

ibnbr:
Removed VALPOS and DOMNAME. All tests OK now.

src/objects/zcl_abapgit_object_doma.clas.abap
```diff
@@ -82,10 +83,13 @@ CLASS ZCL_ABAPGIT_OBJECT_DOMA IMPLEMENTATION.
       " Domain values
       lt_dd07v_tmp = it_dd07v.
       LOOP AT lt_dd07v_tmp ASSIGNING <ls_dd07v>.
+        lv_valpos = <ls_dd07v>-valpos.
         READ TABLE lt_dd07_texts ASSIGNING <ls_dd07_text>
-          WITH KEY ddlanguage = <lv_lang> valpos = <ls_dd07v>-valpos.
+          WITH KEY ddlanguage = <lv_lang> domvalue_l = <ls_dd07v>-domvalue_l
+                   domvalue_h = <ls_dd07v>-domvalue_h.
```
larshp: is this change related to clearing of the fields?

ibnbr: Yes, because "it_dd07_texts" don't have valpos filled. I'm not sure if we can use "sy-tabix" to read, probably it'll work but I think reading using the key is most reliable

larshp: hmm

larshp:
If I pull a DOMA serialized with the new code, with an old version of abapGit, what will happen?

ibnbr:
I tested only old DOMA serialized on a new version of abapGit and it's work.

I'll prepare the system to test this, but I think the function module  that update DOMA will return a exception.

ibnbr:
As I thought an exception is raised in the function module, but this message doesn't appear in the UI.

image

I'll try understand the UI of abapGit, because I'm lost on this.

larshp:
is the error caused by clearing VALPOS or DOMNAME? perhaps we can clear just one of them and have it working on old versions

ibnbr:
For both.  DDIF_DOMA_PUT returns 'NAME_INCONSISTENT' for DOMNAME empty and 'DOMA_INCONSISTENT' for VALPOS empty.

larshp:
sorry about the delay

Suggest this approach: add the code to fill the empty values now,

In 4(?) months: add the code that clears the fields

This way users will have a chance to update to latest abapGit, and we can avoid some problems

ibnbr:
OK. So I will split this changes in 2 PRs, right?

larshp:
yes please

larshp:
closing

PR #2993, opened 2019-10-22
Stage improvements: commit selected button

sbcgua:
Follows #2988 
Main purpose: press "me" filter - then immediately press commit, without explicit adding

Long explanation: before there were 2 commit options "Commit all" and "Commit those which are manually marked to add". Now I added the 3rd - "Commit all those which are filtered by current filter" - which appears when filter is active AND there is nothing manually selected. So from UX perspective it is a shortcut for quick filter + commit (2 clicks vs 3 clicks). Plus it is very "intuitive" to filter and then click commit which before was commiting **everything** (which is "unexpected") and I stumbled into this a couple of times.

Sorry cannot attach a picture - some github issue, will try later. But it is better to click it live. Stage -> filter something -> see how button changes -> click commit -> check commit list at the bottom

To do in future: diff statuses in stage and commit page, and probably auto adding files, deleted locally (there was an issue somewhere about it)

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -264,17 +264,21 @@ function StageHelper(params) {
   this.pageSeed        = params.seed;
   this.formAction      = params.formAction;
   this.user            = params.user;
-  this.choiseCount     = 0;
+  this.selectedCount   = 0;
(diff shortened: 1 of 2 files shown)


sbcgua:
2019-10-22_21h57_14

finally attaching works

PR #3003, opened 2019-10-25
Diff state at stage and commit pages

sbcgua:
Stage page
2019-10-25_22h55_40
Commit page
2019-10-25_22h55_52

P.S. review welcomed, imho a bit crutchy, but not sure how to do better without major refactoring
P.P.S `apply_order_by` and `render_order_by` unchanged - reordrings

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_stage.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_stage.clas.abap
@@ -255,7 +255,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_STAGE IMPLEMENTATION.
           ev_filename = ls_file-filename ).
 
       READ TABLE ms_files-status ASSIGNING <ls_status>
-        WITH KEY
+        WITH TABLE KEY
           path     = ls_file-path
           filename = ls_file-filename.
       ASSERT sy-subrc = 0.
@@ -462,7 +462,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_STAGE IMPLEMENTATION.
       READ TABLE lt_changed_by INTO ls_changed_by WITH KEY item = <ls_local>-item. "#EC CI_SUBRC
       READ TABLE lt_transports INTO ls_transport WITH KEY item = <ls_local>-item. "#EC CI_SUBRC
       READ TABLE ms_files-status ASSIGNING <ls_status>
-        WITH KEY
+        WITH TABLE KEY
           path     = <ls_local>-file-path
           filename = <ls_local>-file-filename.
       ASSERT sy-subrc = 0.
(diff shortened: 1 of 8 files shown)


christianguenter2:
I did some tests and didn't spot any issues. Nice feature, btw.

src/ui/zcl_abapgit_gui_page_stage.clas.abap
```diff
@@ -252,6 +254,12 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_STAGE IMPLEMENTATION.
           ev_path     = ls_file-path
           ev_filename = ls_file-filename ).
 
+      READ TABLE ms_files-status ASSIGNING <ls_status>
+        WITH KEY
+          path     = ls_file-path
+          filename = ls_file-filename.
```
christianguenter2: Suggest using `WITH TABLE KEY` to optimize key access as it's a hashed table with keys path and filename. Same for others usages below.

sbcgua: adjusted, though seems to be unnecessary  (see comments below)

sbcgua:
Regarding table key - sap seems to auto identify this. Made some test:

2019-10-26_15h02_41

Source is here: https://github.com/sbcgua/benchmarks

P.S. surprise ! linear+binary faster than native sorted tab, strange, does not make sense - it should be the same approach ...

src/ui/zcl_abapgit_gui_page_stage.clas.abap
christianguenter2: Interesting! I thought that omitting `TABLE` always leads to a linear search for hash tables. Obviously I was wrong. Thanks.
https://help.sap.com/doc/abapdocu_752_index_htm/7.52/en-US/abapread_table_free.htm
image

christianguenter2:
@sbcgua Do you think that feature should also work for patch? I'm not sure, but it would be nice.
Currently it looks like this. But we could do this also in a follow up PR.
image

sbcgua:
hmmm ... I don't know patch codebase well ... but I'll have a look ... should be as easy as inserting one read table somewhere hopefully

sbcgua:
@christianguenter2 
> ... should be as easy as inserting one read table somewhere hopefully ...

as usual after a phrase like this ... it is not that easy. :) 

lets imagine an example of MM file (modified from both sides)
What if I select only those lines modified locally  ? Should the status in commit look like MM or M_ ? And will this work correctly at all (I don't use patch a lot - I don't know)
Also I'd do some refactorings in diff page first. E.g. lo_state is created at stage page but it's empty and not returned there, so I'd move creation into diff page and some other minor stuff
So to summarize I propose to make it in another PR later. I have some more small improvements to stage page to implement soon.

christianguenter2:
Yeah agree. Let's postpone that. I already guessed that it wouldn't be too easy.

PR #3007, opened 2019-10-27
stage UX improvements: ctrl-enter, autoadd deleted, remote state bugfix

sbcgua:
OK, final cherry on this series

1. Ctrl+enter on stage page activates current commit action 
2. Files deleted locally autoadded to the commit (potentially resolves the issue from #2944) - watch out, imho the state is sometimes deceiving @christianguenter2 
3. Stupid bug fix for remote state from the previous PR
4. Autofocus on filter field

image

image

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -302,6 +302,7 @@ function StageHelper(params) {
 
   this.setHooks();
   if (this.user) this.injectFilterMe();
+  this.dom.objectSearch.focus();
 }
 
 StageHelper.prototype.findCounters = function() {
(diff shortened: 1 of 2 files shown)


christianguenter2:
@sbcgua 
Very nice. Did some tests. Overall works as expected.

What's confusing for me is that the link says that there will be one file added to commit, but in fact 3 files are added.
image

image

Should we adjust that to three in this case?

Can we add the CTRL+ENTER hotkey to the hotkey overview?
image

sbcgua:
@christianguenter2 good points, thanks. Will have a look.

sbcgua:
should work now

Commit shortcut - decided to do like this
image

PR #3034, opened 2019-11-04
Improve 'Duplicates' error caused by table maintenance FUGR

### LINKED ISSUE #3027: 'Duplicates' errors caused by table maintenance views
Related: #3020 
I debugged to see where my strange duplicates are coming from.

- The duplicate occurs for each `xxxxf00` include + xml in each table maintenance FUGR. 
- These are autogenerated and excluded from versioning.

The first time it is returned here along with the other includes:
```abap
  CALL FUNCTION 'RS_GET_ALL_INCLUDES'
      EXPORTING
        program      = lv_program
*       WITH_RESERVED_INCLUDES =
*       WITH_CLASS_INCLUDES    = ' ' hmm, todo
      TABLES
        includetab   = rt_includes
      EXCEPTIONS
        not_existent = 1
        no_program   = 2
        OTHERS       = 3.
    IF sy-subrc <> 0.
      zcx_abapgit_exception=>raise( 'Error from RS_GET_ALL_INCLUDES' ).
    ENDIF.

```
Then this part adds it again:
```abap
* handle generated maintenance views
    APPEND INITIAL LINE TO rt_includes ASSIGNING <lv_include>.
    IF ms_item-obj_name(1) <> '/'.
      "FGroup name does not contain a namespace
      <lv_include> = |L{ ms_item-obj_name }T00|.
    ELSE.
      "FGroup name contains a namespace
      lv_offset_ns = find( val = ms_item-obj_name+1 sub = '/' ).
      lv_offset_ns = lv_offset_ns + 2.
      <lv_include> = |{ ms_item-obj_name(lv_offset_ns) }L{ ms_item-obj_name+lv_offset_ns }T00|.
    ENDIF.
```

My approach was to remove the duplicate:
image

This fixes the duplicate error, with a side effect: some of the autogenerated includes come back.

image

This is because the duplicate has a side effect: in `zcl_abapgit_objects`, `check_duplicates` throws
```abap

    li_obj->serialize( lo_xml ).
    lo_files->add_xml( io_xml      = lo_xml
                       is_metadata = li_obj->get_metadata( ) ).

    rs_files_and_item-files = lo_files->get_files( ).

    check_duplicates( rs_files_and_item-files ).
```
which is caught in `zcl_abapgit_serialize->run_sequential` and prevents it from being included
```abap
TRY.
        ls_fils_item = zcl_abapgit_objects=>serialize(
          is_item     = ls_fils_item-item
          iv_serialize_master_lang_only = mv_serialize_master_lang_only
          iv_language = iv_language ).

        add_to_return( is_fils_item = ls_fils_item
                       iv_path      = is_tadir-path ).
        IF NOT mi_log IS INITIAL.
          mi_log->add_exception(
              ix_exc  = lx_error
              is_item = ls_fils_item-item ).
        ENDIF.
    ENDTRY.
```

To get rid of the `duplicates` errors, first we need a better way to exclude these includes.
I already addressed this in #2674 , but did not eliminate all of them. They're gone in the current version, not yet sure when it happened.

FreHu:
Resolves #3027 

- Instead of incorrectly reporting 'Duplicates', the error message now informs about the skipped FUGR.
- No changes to what is serialized.

New message:
image

I would like it to be an info message instead of error though. Does abapgit support different error categories?


mariusraht2:
ZCL_ABAPGIT_GUI_CHUNK_LIB renders different message boxes: warning (e.g. if master language of the repo and system are different), error for exceptions and info for e.g. the version notes of abapGit. But it seems there's no blue info message box which would be cool in my opinion to give the user some feedback about some processes.

FreHu:
The control flow here relies on an exception occurring to exclude the group. I think a reasonable change would be to add severity as an attribute of `zcx_abapgit_exception` for `zif_abapgit_log` to decide on.

FreHu:
Seems to work:

image

After [staging the deletions](https://github.com/FreHu/FUGR_maintenance_view/commit/2f179b7337b94ac56afd3abfc191f1a4b5419397)
image

mkaesemann:
I've checked out the changes in our project and the change skips the whole function group. This is not a good solution, because generated function groups can contain non-generated coding or even function modules (e.g. required for adapted maintenance views) and custom events may be stored in includes linked to the function group itself.
I'm currently testing a solution, which determines if an include is custom 
or SAP, based on the namespace ownership and so far it looks to be working quite well.
The system is closer to how SAP identifies coding and works for Z/Y coding just as well as partner coding (/xy/ Namespaces, like in our add-on).
If I skip the maintenance groups completely, we would loose a good chunk of our add-on coding and not have it in our repository.

FreHu:
@mkaesemann I don't particularly like my or the existing code, bu it seemed to work for my use-cases. I was not aware people put more development on top of the autogenerated content though. This is a dealbreaker. I'm therefore closing this in favor of your upcoming solution.

jfilak:
Please, no smart detection of generated code which works on every system differently - use a robust solution "version everything and let people easily add to ignore list".

PR #3063, opened 2019-11-13
Issue 3060: Check repo URLs before creating a new online repo

### LINKED ISSUE #3060: check repo URLs before creating a new online repo
I had to deal with a shortdump where the user added their package with the url https://github.com/. Then it was not possible start abapgit because zcl_abapgit_url=>name did not parse out rv_name and dumped on the assert.

mariusraht2:
Fixes #3060: Check repo URLs before creating a new online repo

Method regex just splits the url into single components; validates calls regex and didn't check if the relevant parts/components (host and repo name) were provided by the user


filak-sap:
What about to use the method zcl_abapgit_url=>name and replace the assert with an exception?

mariusraht2:
Also we have to make sure that the repo name is only the name between a "/" and ".git" or the end of the url.

PR #3095, opened 2019-11-28
Add adt link to objects on diff page

FreHu:
This PR makes ADT jump possible from the diff/patch page as well. Useful if you noticed a mistake you want to quickly fix. 

5RXNSHC2e1

- Links are not rendered for locally deleted/nonexistent objects
- ADT link generation extracted to separate class

For now in draft form as I want to discuss a few things:
- links for beacons/lines as well?
- the path on the left is actually the path on remote. Is this worth changing (put object name on the local side, path on the right)?


FreHu:
Alright, let's do it this way.

PR #3109, opened 2019-12-08
Commit icon added

mariusraht2:
Related to #3040

- New commit icon (icon-code-commit)

https://github.com/abapGit/icon-font/commit/f2410fdbd20a379e00f5b9bcded4202e167fcc1a


larshp:
for some reason the full files show up as changed? did you change newline format or something?

mariusraht2:
> 
> 
> for some reason the full files show up as changed? did you change newline format or something?

Yes, wondered about it myself: Only copied the source lines from GitHub into Notepad++, saved it and uploaded it again.

larshp:
how does it look after import via abapGit? I guess the files will show up with diffs?

mariusraht2:
Yes:
grafik

sbcgua:
Certainly looks like EOL char issue. Notepad++ would save in CRLF (win) format by default. Also the difference is visible in offline diffviewers as well as file is not ended with newline. 
image

2019-12-16_19h50_33

Not that this is critical but I'd suggest not to edit the files but upload them directly to SAP via SMW0, or to use w3mipoller (see https://docs.abapgit.org/development/developing-ui-css.html) - SMW0 would probably be easier in this case as you just want to overwrite the files with a new icon build.

I would also suggest adding the abap changes in this PR as well - icon is added but __not clear how it will be used__ - it would be nice to couple these changes together.

mariusraht2:
I don't see any problem to just merge these changes into master with the files appended to this pr. I switched several times now from master to my branch revert_commit on different machines and never got a problem. Are there any other problems we would experience, e.g. on older SAP systems?

mariusraht2:
Uploaded the new icon files to SMW0 and due that overwrote the icon files copied via Notepad++.

PR #3110, opened 2019-12-08
Time to String and Date to String converter added

mariusraht2:
Related to #3040

- Added converter methods for date to string and time to string
- Added related unit tests


src/utils/zcl_abapgit_convert.clas.abap
```diff
@@ -78,13 +78,29 @@ CLASS zcl_abapgit_convert DEFINITION
         ev_size   TYPE i
         et_bintab TYPE lvc_t_mime.
 
+    CLASS-METHODS date_to_string
+      IMPORTING
+        iv_date        TYPE rsd_chavl_ext
```
g-back: this data type does not exist in my 740 release

src/utils/zcl_abapgit_convert.clas.abap
```diff
@@ -298,4 +314,44 @@ CLASS ZCL_ABAPGIT_CONVERT IMPLEMENTATION.
...
+        i_chavl_int = iv_date
+        i_inttp     = lc_date
+      IMPORTING
+        e_chavl_ext = rv_date
+      EXCEPTIONS
+        OTHERS      = 0.
+
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( |Date to String Conversion failed. sy-subrc: { sy-subrc }| ).
+    ENDIF.
```
g-back: I'm not familiar with setting OTHERS to 0, but shouldn't the sy-subrc check be pointless because it's always 0?

mariusraht2: Will change it to anything greater 0

g-back: I think you can remove the `EXCEPTIONS` part altogether? The function itself does not raise any exceptions, so the `subrc` should always be `0`.

Regardless: I think you may have to change the function anyway, it seems to be a part of SAP_BW?

mariusraht2: Oh, ok I'll close it for now and search for another solution - thanks nonetheless for the review.

mariusraht2: Have changed the conversion methods - thanks for the review!

larshp:
Hi,
Can you add a few more details on what the usage of these functions is? If we do backend stuff, then I'm pretty sure we dont want anything user specific? If its user specific UI, then just use `WRITE` to trigger the conversion exit?
No need to add references to DDIC data types `sydatum` and `syuzeit`, just use the built-in ABAP data types.

Example:
```abap
REPORT zlhvp005.

DATA time TYPE t VALUE '112233'.
DATA date TYPE d VALUE '20191224'.
DATA str TYPE c LENGTH 10.

WRITE time TO str.
WRITE str.

WRITE date TO str.
WRITE str.
```

mariusraht2:
Didn't know this trick :). Thought write would always write something to the UI?

The added methods are for the conversion of utc time and date to show in the checkout commit popup:
grafik

PR #3124, opened 2019-12-16
Commit icon added

mariusraht2:
Related to #3040 and PR #3109 

-   New commit icon (icon-code-commit)

abapGit/icon-font@f2410fd

Did the import only via SMW0 - no Notepad++ involved this time. After that commited the changes to this branch.


sbcgua:
hmm, strange. the css still have different line endings. But this is how it is built. Both on Win and linux - just checked. How come that it was just LF in the abapGit repo ?

Well, anyway, I think this is OK. I'd approve, but do you plan to add actual abap change ? It would be nice to actually see the intended change (i mean this new icon in UI somewhere), why not to add it to the same PR ?

mariusraht2:
Yes, I'm currently still working on the new feature 'commit checkout' and try to split the necessary commits into small PRs whenever possible else we would have too much to review in the related PRs.

PR #3146, opened 2020-01-08
ODSO Support - new

pawelwiejkut:
Hello,

New implementation of ODSO support. Includes last comments and prevents issues like #3105 by the dynamic call of classes and data creation.

I wasn't sure should I close the old one, so decided to leave it as it is.


christianguenter2:
@pawelwiejkut 
can you provide a test repo here?
https://github.com/abapGit-tests/OSOD.git

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,349 @@
...
+  ENDMETHOD.
+
+
+  METHOD zif_abapgit_object~is_active.
+
+    DATA: lv_dsona TYPE  c LENGTH 30.
+
+    lv_dsona = ms_item-obj_name.
+
+    rv_active = cl_rsd_odso=>factory( i_odsobject = lv_dsona )->is_active( ).
```
g-back: looks like this class is a SAP_BW component, probably has to be called generically?

christianguenter2:
@pawelwiejkut 
Btw. you can call methods dynamically and supply parameters statically. So you don't have to use `          PARAMETER-TABLE lt_col_par.` neccessarily.

pawelwiejkut:
@christianguenter2 , didn't know. Could you please provide some examples?

christianguenter2:
Sure
https://github.com/larshp/abapGit/blob/1a896613b4394258f4785bf5fd42800d0b01028b/src/objects/zcl_abapgit_object_dcls.clas.abap#L69-L79

pawelwiejkut:
Thanks for your comments guys. I work on this, please give me some time. I also have to push extension to IOBJ support before, because for now, we don't support all the features. After tests, I will let you know.

pawelwiejkut:
@christianguenter2 @larshp What is the proper way to handle import/deletion queue if objects have dependencies between themselves? In the example, when object 'XX' should be created after 'AA' but deleted before. Could you please also share an example if it exists ?

larshp:
import: see https://github.com/larshp/abapGit/blob/master/src/zcl_abapgit_objects.clas.abap#L1040

deletion: see https://github.com/larshp/abapGit/blob/master/src/zcl_abapgit_objects.clas.abap#L504

pawelwiejkut:
please tell me if I'm wrong but  "prioritize_deser", is handling import queue based on different object types, right? If I have a case where iobjXX is part of iobjAA, should I try to implement a check in this method? Like, create some kind of loop by all iobj, check dependencies and finally sort order?

pawelwiejkut:
> @pawelwiejkut
> can you provide a test repo here?
> https://github.com/abapGit-tests/OSOD.git

Done

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,381 @@
...
+        CALL METHOD lo_collection->('DELETE').
+
+        CLEAR: lt_col_par, ls_col_par.
+        ls_col_par-kind = cl_abap_objectdescr=>importing.
+        ls_col_par-name = 'E_T_MSG'.
+        GET REFERENCE OF lt_msg INTO ls_col_par-value.
+        INSERT ls_col_par INTO TABLE lt_col_par.
+
+        CALL METHOD ('CL_RSO_APPLICATION_LOG')=>('APPL_LOG_MSG_READ')
+          PARAMETER-TABLE lt_col_par.
```
christianguenter2: What do think, would supply the parameters statically like in `is_active` simplify the code?

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,381 @@
...
+      TABLES
+        infoobjects          = <lt_infoobjects>
+        navigationattributes = <lt_navigation>
+        indexes              = <lt_indexes>
+        indexesinfoobjects   = <lt_index_iobj>.
+
+    IF ls_return-type = 'E'.
+      zcx_abapgit_exception=>raise( |Error when geting details of ODSO: { ls_return-message }| ).
+    ENDIF.
+*chage to method
```
christianguenter2: ?

src/objects/zcl_abapgit_object_odso.clas.abap
pawelwiejkut: Sure, I just didn't have enough time to do this, but I will.

src/objects/zcl_abapgit_object_odso.clas.abap
pawelwiejkut: Sorry for this commit, I should leave this on my disc. This is my note to change ASSIGNMENT, to the method like @larshp done in https://github.com/larshp/abapGit/blob/1a896613b4394258f4785bf5fd42800d0b01028b/src/objects/zcl_abapgit_object_iobj.clas.abap#L299

pawelwiejkut:
> @pawelwiejkut please see comments

@christianguenter2, please check

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,371 @@
...
+
+  METHOD zif_abapgit_object~delete.
+
+    DATA: lv_odsonam    TYPE c LENGTH 30,
+          lv_objname    TYPE sobj_name,
+          lo_collection TYPE REF TO object,
+          lt_msg        TYPE rs_t_msg,
+          ls_msg        TYPE bal_s_msg,
+          ls_col_par    TYPE abap_parmbind,
+          lt_col_par    TYPE abap_parmbind_tab.
```
christianguenter2: Unused variables can be deleted.

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,371 @@
...
+    TRY.
+        CALL METHOD lo_collection->('ADD_TLOGO')
+          EXPORTING
+            i_objnm  = lv_objname
+            i_modify = abap_true
+            i_delete = abap_true.
+
+        CALL METHOD lo_collection->('DELETE').
+
+        CLEAR: lt_col_par, ls_col_par.
```
christianguenter2: Not used.

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,371 @@
...
+          lr_indexes     TYPE REF TO data,
+          lr_index_iobj  TYPE REF TO data,
+          ls_return      TYPE bapiret2.
+
+    FIELD-SYMBOLS:
+      <ls_details>      TYPE any,
+      <lv_tstpnm>       TYPE any,
+      <lv_timestmp>     TYPE any,
+      <lv_conttimestmp> TYPE any,
+      <lv_owner>        TYPE any,
```
christianguenter2: Unused variables.

src/objects/zcl_abapgit_object_odso.clas.abap
pawelwiejkut: done

src/objects/zcl_abapgit_object_odso.clas.abap
pawelwiejkut: done

src/objects/zcl_abapgit_object_odso.clas.abap
pawelwiejkut: done

src/objects/zcl_abapgit_object_odso.clas.abap
```diff
@@ -0,0 +1,363 @@
...
+
+  ENDMETHOD.
+
+
+  METHOD zif_abapgit_object~delete.
+
+    DATA: lv_odsonam    TYPE c LENGTH 30,
+          lv_objname    TYPE sobj_name,
+          lo_collection TYPE REF TO object,
+          lt_msg        TYPE rs_t_msg,
```
christianguenter2: `RS_T_MSG` is part of BW. As compatibility issues are likely I'd suggest to use `STANDARD TABLE OF BAL_S_MSG` instead.

pawelwiejkut: fixed

pawelwiejkut: I have to try to automatize somehow this kind of check.

christianguenter2:
@pawelwiejkut  
When I import the demo repo in my NW752 SP04 system I get this diff. Is this something to worry about?
image

pawelwiejkut:
@christianguenter2, I don't think so. The difference is that I have created my own info area for storing ABAPGIT objects. For now, we don't have support for info area but I will provide it in one of my next PRs (this is AREA object type). Change of object assignment between info areas is usually possible even on PROD systems.
Screenshot 2020-01-17 at 15 49 18
Screenshot 2020-01-17 at 15 55 08

larshp:
thanks guys 👍

PR #3164, opened 2020-01-19
fugr: do not uncoditionally add T00 include

filak-sap:
I know don't why but on our systems the "include for generated maintenance
views" is returned from the function module RS_GET_ALL_INCLUDES, so
when we add it in the method zcl_abapgit_object_fugr=>includes
we have the file twice which cause duplicate errors on our systems.

I didn't remove the code because we are supposed to run on all ABAP
rutime versions and the problem with duplicates for the T00 includes
has started occurring only recently, so the older versions need this
artificial include (I guess).


filak-sap:
Without this patch, colleagues of mine could not commit files belonging to generated maintenance views because serialization was interrupted by a duplicate error.

PR #3167, opened 2020-01-24
Fix for issue #1044

BlackmanCC:
unicode systems always add the byte order mark to the xml, while non-unicode does not
so we will always add the byte order mark if it is not in the xml


larshp:
merged https://github.com/larshp/abapGit/pull/3166, and pulled upstream changes to your branch, correct errors from abaplint should be shown now 😄

BlackmanCC:
On our non-unicode system abapGit now is usable against repositories from unicode systems.

src/objects/zcl_abapgit_objects_files.clas.abap
```diff
@@ -170,8 +170,10 @@ CLASS ZCL_ABAPGIT_OBJECTS_FILES IMPLEMENTATION.
 
   METHOD add_xml.
 
-    DATA: lv_xml  TYPE string,
-          ls_file TYPE zif_abapgit_definitions=>ty_file.
+    DATA: lv_xml        TYPE string,
+          lv_hex(1)     TYPE x VALUE '23',
```
larshp: suggest

```suggestion
          lv_hex     TYPE x LENGTH 1 VALUE '23',
```

similar for the next definition

BlackmanCC: ok. I will change

BlackmanCC:
@larshp Sorry, it is my second contribution project. Are there any further steps left for me?

larshp:
also change `lv_hex_bom(3) TYPE x VALUE 'EFBBBF',` to use `LENGTH`, but no, code looks good

I typically leave the pull requests for a few days for people to read / review / try out

jrodriguez-rc:
It looks really good but still adds BOM character for abapgit.xml file:
image
image

This could be because the modification adds BOM character to local serialized file instead remove character from the repository files
image

BlackmanCC:
On our system it looks like this:
image

jrodriguez-rc:
It is same I have shown on the first screen, the second screen is a file (package) that doesn't detect changes because of BOM character is added in local serialized file instead remove in remote file

BlackmanCC:
@jrodriguez-rc Please check again.

BlackmanCC:
@larshp maybe it would be better to include the new logic for BOM into zcl_abapgit_convert=>string_to_xstring_utf8( lv_xml ).
What do you mean?

larshp:
zcl_abapgit_convert: yea, but then I suggest adding it as a new method, string_to_xstring_utf8 I would expect to do raw conversion

BlackmanCC:
new method like "fix_bom_for_nonunicode"?

larshp:
perhaps `string_to_xstring_utf8_with_bom` ?

PR #3181, opened 2020-01-26
Checkout Commit 2/2

mariusraht2:
New Feature "Checkout Commit"; Part 2/2
Related to #3040 - Revert to any single commit

**NOTE:** This commit depends on branch "checkout_commit_2"!

### Change Log
**zif_abapgit_popups**
- Specified some more import parameters for method popup_to_select_from_list to be able to reuse popup logic for commit list

**zcl_abapgit_popups**
- Added eventhandler on_double_click to be able to react on double click executed on a line
- Differenciate between multiple and single selection mode to be able to control behaviour and look of the popup
- Removed obsolete variable lv_line in method on_select_list_link_click
- Added some logic in popup_to_select_from_list to be able to control the content, header and width of single columns

**zcl_abapgit_services_repo**
- Adjusted source code popup_overwrite to be consistent with other source code within its class, use field-symbols instead of work areas and use global defined types

**zcl_abapgit_git_object**
- New class to consolidate same source code between zcl_abapgit_git_commit and zcl_abapgit_branch_overview

**zcl_abapgit_git_commit 
- New class to be able to get commits from remote by specific hash

**zcl_abapgit_branch_overview**
- Moved some source code to zcl_abapgit_git_object because it shares it with zcl_abapgit_git_commit

**zcl_abapgit_gui_router**
- Register action checkout_commit and commit_switch though both call the same action logic

**zcl_abapgit_services_git**
- Implement logic to determine commits (max. on level 99), open popup, handle selection and switch/load commit
- Moved logic for deletion of obsolete objects to method "handle_popup_del_obsolete_obj" to be able to reuse it though we don't reuse it in the newest version of checkout_commit
- Empty commit hash if branch is being switched

**zcl_abapgit_gui_view_repo**
- Moved constructor to top
- Add menu item "Checkout commit" to "Advanced"-dropdown

**zcl_abapgit_gui_page_boverview**
- Adjust call of zcl_abapgit_gui_chunk_lib=>render_repo_top

**zcl_abapgit_gui_page_commit**
- Adjust call of zcl_abapgit_gui_chunk_lib=>render_repo_top

**zcl_abapgit_gui_page_main**
- Adjust call of zcl_abapgit_gui_chunk_lib=>render_repo_top

**zcl_abapgit_gui_page_merge**
- Adjust call of zcl_abapgit_gui_chunk_lib=>render_repo_top


mariusraht2:
Sry in advance for the great PR - I tried to split it up but encountered numerous dependencies :(.

PR #3210, opened 2020-02-19
Patch: add beacon at beginning of objects

christianguenter2:
Before:
image

After:
image


FreHu:
Maybe it doesn't matter much, but for some files there will always be an empty beacon. Seems unnecessary aside from visual consistency.
image

sbcgua:
+1, for no beacon at the beginning of the file

christianguenter2:
@FreHu @sbcgua do you have general doubts about the beacon at the beginning or is it just the visual appearance for .xml files?

Reason why I introduced it was this case:

Without it's not possible to just select the first few changes en bloc.
image

with it is
image

Maybe we could optimize it this way, when the function of the beacon is select all diff lines of the like the one above we skip it?

sbcgua:
Ah, I missed the point about patching (I don't use it a lot). Patching makes sense. My point was about visually unnecessary blocks. Well, is it possible to add the starting beacon just in case of "patch" mode ? I think this would solve the problem.

PR #3213, opened 2020-02-19
Fix patch: multiple files with same name

### LINKED ISSUE #3208: Patch - files with the same filename in different folders are not treated correctly
Example: All subpackages will contain a `package.devc.xml` file. 
- On the patch screen, if you select changes in one file, it will select all `package.devc.xml` files.
- When you go commit, only one `package.devc.xml` will be staged.

(you can still commit the files by not using patch for this)

christianguenter2:
- unify id generation
- generate unique filenames

fixes #3208


FreHu:
I noticed the jump feature also needs unique filenames or it will always jump to the first object with the same name.

christianguenter2:
I'll have a look into that. Should we separate that in a separate PR?

Issue #3209, opened 2020-02-17
Only serializing master language is not respected for longtexts


FreHu:
Even with this option enabled, longtexts in non-master languages will still be serialized. This is either a bug or a missing feature.

sbcgua:
sounds like a bug, could you give a clearer examples pls ?

FreHu:
Use case: I'm trying to compare two systems, one has more translations than the other so I see a lot of differences.

If I check this,
![image](https://user-images.githubusercontent.com/5097067/74716219-66822580-522e-11ea-899b-1e64e8c5427e.png)

the longtexts do not disappear from my message class. Example longtext:
```xml
    <item>
     <DOKIL>
      <ID>NA</ID>
      <OBJECT>xxx</OBJECT>
      <LANGU>1</LANGU>
      <TYP>E</TYP>
      <VERSION>0003</VERSION>
      <TXTLINES>00002</TXTLINES>
     </DOKIL>
     <HEAD>
      <TDOBJECT>DOKU</TDOBJECT>
      <TDNAME>xxx</TDNAME>
      <TDID>NA</TDID>
      <TDSPRAS>1</TDSPRAS>
      <TDFORM>S_DOCU_SHOW</TDFORM>
      <TDSTYLE>S_DOCUS1</TDSTYLE>
      <TDVERSION>00003</TDVERSION>
      <TDLINESIZE>072</TDLINESIZE>
     </HEAD>
     <LINES>
      <TLINE>
       <TDFORMAT>U1</TDFORMAT>
       <TDLINE>&amp;CAUSE&amp;</TDLINE>
      </TLINE>
      <TLINE>
       <TDFORMAT>AS</TDFORMAT>
       <TDLINE>foo_text</TDLINE>
      </TLINE>
      <TLINE>
       <TDFORMAT>U1</TDFORMAT>
       <TDLINE>&amp;WHAT_TO_DO&amp;</TDLINE>
      </TLINE>
      <TLINE>
       <TDFORMAT>AS</TDFORMAT>
       <TDLINE>bar_text</TDLINE>
      </TLINE>
     </LINES>
    </item>
```

FreHu:
Solved.

PR #3217, opened 2020-02-19
DOCU master lang only

sbcgua:
To #3209 @FreHu 

Should work...
Although the code change is minor (`serialize` only - others are reorderings) maybe need a review (@larshp ?) - this is not the most familiar part of the AG code to me :)

P.S. Long text translation UI is a disaster ! Appeared I didn't do this before ... UX is "fantastic" ... just whining ...


sbcgua:
P.P.S. there are more calls to `DOCU_READ` in the code ... others are in DSYS and DOCV. Didn't touch them but maybe they will popup later

PR #3234, opened 2020-03-03
Multiple sicf with identical names is not imported #3230

jrodriguez-rc:
I removed some code, that gets the first entry if not found checking hashed path. I don't know why it was added


jrodriguez-rc:
I found commit with this change 55013696ae9d8bd9d341221127f611d2780fa3aa but doesn't specify the error that causes the problem

Issue #3165, opened 2020-01-21
Diff/patch - add option to collapse files


FreHu:
Add a mechanism to collapse files similar to github's recent "viewed" feature (no need for any progress bar). Useful if you want to collapse a large diff that does not interest you.

![image](https://user-images.githubusercontent.com/5097067/72812006-0add7000-3c61-11ea-88e1-f9842e4c2a17.png)

PR #3248, opened 2020-03-10
Diff - Add option to collapse files

EduardoCopat:
Implements #3165 

@FreHu @sbcgua


src/ui/zcl_abapgit_gui_page_diff.clas.abap
```diff
@@ -1111,6 +1111,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_DIFF IMPLEMENTATION.
       ro_html->add( '<th class="num">new</th>' ).           "#EC NOTEXT
       ro_html->add( '<th class="mark"></th>' ).             "#EC NOTEXT
       ro_html->add( '<th>code</th>' ).                      "#EC NOTEXT
+      ro_html->add( '<th class="collapse" onclick="onDiffCollapse(event);">▼</th>' ). "#EC NOTEXT
```
EduardoCopat: Any idea how could I solve this abaplint issue?

**Contains non 7 bit ascii character
7bit_ascii**

Should I try to use an ASCII character ou replace it with an icon?

larshp: suggest: add "zcl_abapgit_gui_page_diff" to exclude in abaplint.json, under the rule

larshp: yea, moving it to the JS file might also work

EduardoCopat: I added the exception. Thanks!

FreHu:
A few ideas:

I think the ideal location for the collapse button would be the left side, the screen is wide and there is nothing to click on the right. 

image
vs
image

It could be placed left the added/removed/modified counters, there is more space for it there, you could have a more noticeable icon.

When I collapse everything, I would like to have an overview of my staged changes. 
- add expand/collapse all options to the top menu
--------------------------------
Collapsing of partially staged files

image

In the first file, I have no lines selected, but in the second one I have one of two lines selected. They both make it look like nothing is selected. 
- There could be an extra line shown in collapsed mode which would summarize the user's selection ("1 of 2 lines staged")
- Different background colors for `nothing staged`/`something staged`/`everything staged`

EduardoCopat:
@FreHu, thanks for the ideas! I will work on them.

I think we could follow how Github collapses:

image

image

They use the collapse button close to the counters.

larshp:
@FreHu does the change look okay from your point of view?

FreHu:
The collapsing works, but I'd merge it after arrows are moved to the left. The rest can go in different PRs.

EduardoCopat:
ezgif com-video-to-gif

- I added a nice icon (chevron) for collapse/expand. Similar to github.

As mentioned, we can create issues for the other comments.

sbcgua:
Heh, I'm too late everything is already reviewed. :)
From my side no suggestions, nice and clean code, cool and useful feature. :+1:

src/ui/zcl_abapgit_gui_page_diff.clas.abap
sbcgua: Just FYI: there are unicode symbols escaping in html. This one would be `&#x25bc;` I guess

image

But chevron icons look better anyway

larshp:
thanks guys 👍

PR #3251, opened 2020-03-11
Use xml syntax highlighter for html as well

FreHu:
I came upon a `w3ht.data.html` object without syntax highlighting. We can reuse the xml highlighter for html to fix it.


larshp:
how does the xml highlighter handle tags without closing tags, eg `<br>`, just want to make sure it does not throw an exception or something bad

FreHu:
Seems to be okay with weird inputs:
image

PR #3253, opened 2020-03-11
Minor ux improvements

FreHu:
My colleague does not like that long names are split into two lines:
image

He suggested having the stage list full width just like the repo page.
SapGuiServer_IE2kuBnOxP

image

This is easy to do but possibly there's now too much space, plus there are inconsistencies in column placement which can be discussed/resolved.
- Transport column in the repo page is in a different position, repo page has an icon with hover text instead
- Status icons are on opposite sides
- No ADT links on the stage page. Clicking the path opens the diff.
- There is no last change on the repo page, could be useful


sbcgua:
The code is OK. But the feature is very controversial. From time to time I'm thinking about the **opposite** for the repo page :))) The content is very stretched, it makes it difficult to follow left and right parts of it (I have 21:9 monitor for example, but have issues even on half-screen windows).

Thoughts:
1) maybe make it switchable ? "centred" and "wide" layouts (including the heading part) like in github
2) make hover highlight
3) rearrange it more properly, I actually agree with your bottom comments and was thinking in the same direction e.g. it's confusing to click "view diff" on the repo page roght side and file name at stage. So unifying the layout would be great. Suggestion:
    - move status to the left for both
    - first icon at the repo page - remove ? it looks nice but does not bring value
    - then obj type/name with a link to DIFF (!!!)
    - right aligned in the same cell is "open" link (visible just on hover)
    - then filename, if displayed. And by the way maybe, if not displayed, hide it at stage screen as well - does it make sense to commit a part of an object (except of manual patch)
    - then additional read only infos like transport and changer
    - now, question: where stage commands should be ? This is tricky. abapGit commands gravitate to the right. But ... it means name and command are stretched. I would probably move it after the name (before infos) but they will be in the middle then ... not sure. Needs experiments.


Maybe do it all ? What do you think ?

P.S. also sometimes I'm thinking about complete separation of UI and server part and implementing the UI as a external tool/page with electron or similar ... but it is so much work ...

sbcgua:
more: I'm also not sure that wrapping is bad. Maybe name could be longer ... but point is that if the column is dynamic then the following infos are always placed at different distance from the left edge which makes it less readable

sbcgua:
for the benchmark - github solves it by hiding part of name and hints

image

FreHu:
As for UI overhauls, I was thinking of simply using Fiori fundamentals inside the abap browser, it should be ie11 compatible.

FreHu:
I think these are good suggestions. I'll try to add them over time and see how it looks.

FreHu:
> maybe make it switchable ? "centred" and "wide" layouts (including the heading part) like in github

On second thought, github does not do this. 
The _diff_ is switchable, but abapgit has that too. Github actually centers content quite heavily. It's almost the same width as abapgit's stage currently, which is 900px.

image

Looks like a lot of wasted space on 1440p, but it's actually meant to fit on half of your screen.

1080p, half width
image

1440p, half width
image

And starts breaking with less space:
image

(I don't have access to crazy aspect ratio screens, feel free to post screenshots)

I think a better direction to take would be to try and center the content.

image

image

FreHu:
The patch button was a bit misplaced so I moved it to the top (it does not interact with the stage list, just changes pages). 
image

FreHu:
Actually, another important difference: repo starts with object type/name, stage only has paths. If you move status icons to the left, in objects with many parts (FUGR for example) you would get many icons associated to the `path` which is visually far.

Status and path should be next to each other. So I would actually switch the first column on the repo page to path, then move the status icons to the left. Maybe even the file icons would start making sense if they were drawn for each path istead, to distinguish code from metadata.

Github uses icons too 
image

FreHu:
I would like to merge this as it is generating way too many ideas and it might be better to do in smaller increments. 

I created #3265, #3266, #3267, #3268 and #3269 and #3270 as follow-up.

This PR just adds a bit more space between things, centers the stage list and moves the patch button.
image
image

sbcgua:
will check it tomorrow

sbcgua:
### Comments

> The patch button was a bit misplaced ...

:+1:

> Status and path should be next to each other

Mostly agree, just that some people do not display path at repo page - just objects. But maybe you're right about the order: status, path, object - where path is optional, it makes sense. Icon at github just distinguish folders and files.

> line-height 1.5

:+1:

### screenshots for history

conditions: 21:9 screen, 150% font scale (!)

2020-03-17_19h09_53
2020-03-17_19h09_59


### Final thoughts

I'm a bit tentative to approve it ... followup issues are ok. But this particular change does not look finished. E.g. changing the size of stage table but leaving the header at full width - they should be consistent imho (e.g. buttons look scattered). What do you think ?

I'd propose the following: let's make the stage table to be full width (100%). Thus this change will be dedicated to the length of filename and line spacing. And then we can work on the width separately.

FreHu:
Alright, let's do it like you propose. I was trying to mimic github which also has a full-width header, but there are two levels of header and one has centered content so I guess you're right.
image

PR #3256, opened 2020-03-12
CHDO: clear UZEIT and UTIME dynamically

### LINKED ISSUE #3255: Method zif_abapgit_object~serialize - components udate and utime not available
Hi all,

While trying to activate the current release in our ECC6 EHP8 system I got 4 syntax errors (see screenshot below).

image

There are no components "udate" and "utime" in the objects "<ls_objects>" and "<ls_objects_text>".
"ls_objects" is referencing database table "TCDOBS".
"ls_objects_text" is referencing database table "TDCOBTS".
Both database tables do not have columns called "udate" or "utime".
There are no date or time columns in those tables at all.

What are "udate" and "utime" supposed to be used for?
Should they reference the columns "datum" and "uzeit" of database table "TCDRPS" maybe?
Is there an older build available which doesn't have that issue?
Please assist.

Best regards
Dominik

christianguenter2:
fixes #3255


mbtools:
See https://github.com/larshp/abapGit/issues/3255#issuecomment-598374733

PR #3260, opened 2020-03-13
CHDO: clear UZEIT and UTIME dynamically (#3256) bis

marcellourbani:
Had this floating around since Monday but couldn't find the time to rebase.
Would have dropped it as I saw you posted a fix, so I tried it.

My solution is less elegant but more compact, and more importantly doesn't dump:
image


larshp:
what line/code causes the dump?

marcellourbani:
Funnily enough it didn't say.
Sorry I forgot the abaplint bit, will fix it now

marcellourbani:
I spinned it up to replicate:
image
Changing the string format doesn't do the trick:
image

The system isn't THAT old btw:
image

marcellourbani:
No dump in 750sp5 btw

christianguenter2:
Reason for the dump was that the key fields of `<ls_objects>` are flagged as read-only. And that's because `ls_change_object-objects` is a sorted table.
image

https://github.com/larshp/abapGit/blob/7bf443d7a85d36bcd7b948efc0aeb0ac61ddb8f2/src/objects/zcl_abapgit_object_chdo.clas.abap#L42

PR #3262, opened 2020-03-14
Handle customizing

goreraks:
#3244
> Customizing transport to create relevant files based on object name & object type

> Change the customizing in one of the object and refresh the ABAGIT screen


goreraks:
Since AbapGit supports only the TADIR objects, it doesn't make sense integrate the support of customizing here.
As it requires several tweaks in the abapgit code, which may have some effects on the performance.
Also it make sense to create a new report to handle only the customizing content using the APIs, see https://docs.abapgit.org/development/api.html.

Issue #2664, opened 2019-05-10
Soft fail of remove fetch


sbcgua:
Currently if remote fetch fails the list of object is not shown, just an error. However it would be nice to see it for at least 2 reasons:

1. to observe the list of objects
2. to see the status change compared to checksums ( in other words if all the changes were committed before )

PR #3271, opened 2020-03-17
Patch: new refresh functions

christianguenter2:
Three new refresh functions on the patch page.˙

1) Local refresh of single object 
2) Local refresh of all objects
3) Full refresh of all objects, local and remote
image

abapGit15


FreHu:
Big thanks, I consider this a huge improvement to the committing experience. Some initial ones, I'll do some more thorough testing tomorrow:

- The jumping behavior is a bit confusing.
- I managed to get an error by trying to refresh `.abapgit.xml`.
image

christianguenter2:
- Jumping: Yeah I'm not sure about that too. Otherwise it's also very annoying if the page jumps to the top after refreshing an object. 
- `Refresh object` depends currently on the object. As there is none for `.abapgit.xml` there error occurs. I disable `Refresh` for non object files. Will try to improve this further.

FreHu:
After committing through patch, it would be interesting if abapgit had the option to go straight back to patch after committing, width updated diffs. What do you think about this?

christianguenter2:
Good Idea, I think this should be possible. I'll investigate. 
I think that's then something for a follow up PR?

FreHu:
Of course, split things up as you see fit.

christianguenter2:
Yeah already working on it. Do you think we can merge this PR as is?

FreHu:
- Currently refreshing jumps to the top, but does not jump back to the file. I think this is worse than jumping there and back, it's very easy to get lost.

fWQj46OO5b

- Deleting a file and then refreshing will lead to errors and a blank page:
image

From this state, doing refresh local will show a different error and refresh all will dump.

christianguenter2:
- Ok, then I'll reenable jumping to the file.
- I'll check the deletion case

christianguenter2:
@FreHu 
- jump to refreshed object enabled again
- refresh of deleted objects should work now
Can you check again?

FreHu:
Confirmed working. But I found another dump when creating a new class, then doing refresh local.
image

christianguenter2:
Thanks I'll check

christianguenter2:
@FreHu fixed. Please try again.

FreHu:
Works correctly now.

christianguenter2:
I just added hotkey support for refresh all local objects, as I need that the most ;-) Now I'm done.

christianguenter2:
Btw. using the hotkey jumping is much more smoother than when refreshing single objects

FreHu:
I've never given hotkeys a chance, I'll try to give them a go.

christianguenter2:
@larshp do you think we can merge this? I'd like to prepare a follow up PR :-)

larshp:
hmm, there are changes to `src/zcl_abapgit_repo.clas.abap`, its the central point that needs most refactoring, adding additional methods makes it more difficult than it already is

christianguenter2:
Hm, ok I see. Any idea how we can tackle this?

sbcgua:
Imho. I'd propose to merge and then refactor separately. Repo needs refactoring for some time by now ... I've been proposing some idea before in #2216 - to separate repo as a container and local-remote "strategies".  

P.S. There is also #2664 hanging from my side :( can't get hands on it ... maybe soon.

PR #3338, opened 2020-05-02
Gui refactoring 2

sbcgua:
Step 2 to #3286, ufff, I thought it would be times easier.

## So the goal

Components self register as a hotkey providers, line the event handlers in the previous PR

```abap
* ZCL_ABAPGIT_GUI_PAGE_MAIN->RENDER_CONTENT
...
mi_gui_services->get_hotkeys_ctl( )->register_hotkeys( me ).
```

Consequently several components may independedntly register own hotkeys, the rendering logic compresses the hotkeys, removing duplicates (later are of higher priority).

## Tech

All the stuff for hotkey are now collected in `zcl_abapgit_hotkeys`: hotkey control, hotkey rendering component (with some exception which I will fix in later PRs), own hotkey (`?`) also unified to fit in the same flow. This means that `zcl_abapgit_hotkeys` plays 3 roles and it is a multifunctional mess, but ... it is all in one place and class is not long so maybe it is OK.

## Settings

... it took ~4x times than the "goal" and I'm not sure the time was invested properly. So now it looks like this

2020-05-02_12h49_58

Although the feature itself is cool as well as the dynamic stuff in the original code, but I have very controversial feelings. The interface defines static methods, hotkeys are collected using `cl_oo_class` stuff ... ufff. A bit heavy imho. So if this feature is not used by people, I'd vote to remove it at all ... comments are highly welcomed.

UPD: ah, yes. The interface was renamed and moved to core package, empty implementations of the old interface in all the pages removed

UPD2: and I didn't check the one-file version ... all my current systems are full-abapgit-enabled 😝


FreHu:
I tried the merged version. There is a dump when you go to settings (invalid dynamic method call).
image

As for hotkeys, I don't use them personally. I've tried to start using them multiple times but always fall out of the habit quickly. I think there are very few hardcore enough abapgit users to warrant configurable hotkeys and the feature is largely unknown. Based on personal observation I'd estimate most users don't even visit the settings. I'm all for removing bloat and I think it is bloat, but I can't speak for everyone.

sbcgua:
@FreHu :+1:, thanks, will have a look

> most users don't even visit the settings
=), yeah, true ...

sbcgua:
P.S. hotkeys are cool. Especially the possibility to enter stage, where there are only remote changes so no "Stage" button at the screen 😝 - used it several times :)))
More practical - S key = refresh + stage - quicker when no changes to the repo on abapgit run, but then you change the code and need to refresh and then stage separately (and stage does refresh once gain so it is also longer)

sbcgua:
@FreHu should work now, have a look. Forgot to copy `where` from the old code

FreHu:
works now 👍

PR #3347, opened 2020-05-05
Highlight hovered line

### LINKED ISSUE #3265: Highlight hovered line
Add highlighting to hovered line in 
- repo
- stage
- diff

Colors should not clash when using different themes.

FreHu:
resolves #3265 

In the diff page i added a top and bottom border around the line instead of background.

SapGuiServer_IQQHqZqc0t
image
SapGuiServer_ZDPXZhaAPn

SapGuiServer_jmVYmyyYnu
SapGuiServer_Ue0RtMfW8G
SapGuiServer_imNNN0vm4u


sbcgua:
will have a look tomorrow. Conceptually look fine, but I think colors could be better. And also the highlight method (background in repo and borders in diff - I think should be unified)

FreHu:
I added the borders not to clash with the diff colors, but it never occurred to me to go back and use borders everywhere. Will try.

sbcgua:
- Maybe use just bottom border to underline hovel line (but I'm not sure it is a good idea)
- or maybe add a shadow effect to "rise" the hover item above the table
- or just use the color but with opacity so that diff main color is tinted but not lost completely

Without visual test I think that the last could be the best.

- just in case, adjusting the font (e.g. bold) would be bad idea because of element size change. Color of font is also not an option as there are these icon-like things

larshp:
pending @FreHu ?

FreHu:
Will do some experiments in the evening.

larshp:
no hurry, just getting an overview of what the status is

FreHu:
In the end i decided not to add highlight for the diff

*   a border does not look good in the old browser, the borders always look very rough, if you compare to vscode's border.



*   effects are nice to play around with (increasing font size is fun) but the result is way too jumpy like scbcgua said.
*   a background color can clash with both syntax highlighting and diff colors, so i think it's too much.

The yellow was changed to a less aggressive color

sbcgua:
yellow and yellow ... hmmm.
let me look at it tomorrow. I have one idea to check. otherwise I'll approve.

src/ui/zabapgit_css_common.w3mi.data.css
```diff
@@ -251,6 +251,11 @@ table.repo_tab {
 }
 .repo_tab tr:first-child td { border-top: 0px; }
 
+.repo_tab tr:hover {
+  background: #f0e5bf !important;
```
sbcgua, line 255: `background-color` ? to be more specific

src/ui/zabapgit_css_common.w3mi.data.css
```diff
@@ -280,6 +285,10 @@ input.stage-filter { width: 18em; }
   font-weight: bold;
 }
 
+.stage_tab tr:hover {
+  background: #f0e5bf !important;
```
sbcgua, line 289: `background-color` ? to be more specific

FreHu:
I am known for liking bright colors way more than other people, so I'll go with your suggestion.



Also fixing some syntax errors in the dark theme css.

larshp:
@FreHu please update/rebase the branch to master, I cannot click the "update branch" button for some reason

PR #3348, opened 2020-05-06
Don't run code inspector in parallel when disabled

FreHu:
If this is checked in the settings,
image

then code inspector/syntax check will now run synchronously.


src/zcl_abapgit_code_inspector.clas.abap
```diff
@@ -127,11 +133,7 @@ CLASS zcl_abapgit_code_inspector IMPLEMENTATION.
     mv_name = |{ sy-uname }_{ sy-datum }_{ sy-uzeit }|.
 
     " We have to disable parallelization in batch because of lock errors.
```
christianguenter2, line 135: Maybe we should move that comment to the ```decide_run_mode``` method?

PR #3355, opened 2020-05-09
Gui refactoring, chapter 3

sbcgua:
OK, more steps towards disentangling html components.

Main idea: 
- a new primitive - `ZCL_ABAPGIT_HTML_PARTS` - which is a collection of named sets of zcl_htmls
- a managed instance of it is accessible via `zif_abapgit_gui_services` and allows to store html chunk e.g. scripts or postponed hidden forms (stage_page) and render them later
- dev note: `zcl_abapgit_gui_component->c_html_parts` is the place to define "standard" cross-page part names

In addition one example is implemented to illustrate the flow within this PR - hotkeys script completely removed from `gui_page` and now rendered within `hotkeys` class where it should be rendered. `gui_page` is completely unaware of it which is good.

Further refactorings planned for `gui_page->get_events` (hidden forms) and maybe something else. Fill free to join :)

Ths is not the final chapter of GUI refactoring though - there is one conceptual bug I'g going to fix, see the comment in `hotkeys->render` code. And some more cleanups of `gui_page` are planned

P.S. some clutter in this PR: removed alias for `cache_asset`, could not pass by sorry.

### DIFF
--- a/src/ui/core/zcl_abapgit_gui.clas.abap
+++ b/src/ui/core/zcl_abapgit_gui.clas.abap
@@ -339,6 +339,8 @@ CLASS ZCL_ABAPGIT_GUI IMPLEMENTATION.
     ENDIF.
 
     CLEAR mt_event_handlers.
+    mo_html_parts->clear( ).
+
     IF mi_router IS BOUND.
       APPEND mi_router TO mt_event_handlers.
     ENDIF.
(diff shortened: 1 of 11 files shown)


sbcgua:
abaplint seems to hang ... hmmm :)

jinigirl:
Hello! Updated my abapGit installation just now, and I'm having short dump when running tcode ZABAPGIT. I think it's caused by the removal of the cache_asset alias.

image

sbcgua:
ref #3286

PR #3371, opened 2020-05-18
Set filter for file open and save dialogs / Add new setting for default commit comment

mbtools:
Set the file filter dropdown and default to *.ZIP in file open and save dialogs


mbtools:
(still an aG beginner... didn't means to combine those changes, sorry)

mbtools:
Some screenshots of the new setting with $FILE:

image
image
image

or using $OBJECT

image
image
image

larshp:
personally I think the predefined commit comment is an anti pattern, but its behind a user flag, so ok

mbtools:
Thx for approving! It's a great simplifier esp. when you're the only one commiting to your own repos 👍

PR #3381, opened 2020-05-21
New syntax-highlighters for CSS, JS, JSON

mbtools:


mbtools:
Some examples using the new "Local View" (see coming experimental feature, which I will post to an issue).
Image1
Image3
Image4

larshp:
LGTM

@sbcgua any comments for the CSS changes

sbcgua:
Wow, new cool parsing stuff :)) Nice, nice ! Tested, works fine.

Colors:
- I would remove red coloring from brackets (operator) - they seem popping up too much from the rest of the code
- maybe change color of values to green - to conform with strings in abap (arguable - violet looks fine, just out of consistency considerations) - this supposes creating a selector instead of using `attr_val`
- if yes for the green above then probably `selector` should be changed *from green* as well for consistency. E.g. to violet ?

Anyway, a matter of preference. I'd just recommend removing red brackets :) Otherwise looks very cool and works.

abap
image
css
image
js
image

mbtools:
I removed the brackets since they where causing also too much workload. Also changed the text style as suggested. Should we change the attr_val style (for xml/html) to green as well to be consistent?

mbtools:
Test repo: https://github.com/mbtools/abapgit-syntax-highlight
You can switch between branches to get diffs to render.

mbtools:
I'm done with this one. Anymore wishes regarding syntax highlighting?

larshp:
@sbcgua any additional comments/suggestions?

sbcgua:
Sorry, have been a bit busy. The syntax CSS looks fine. But one thing. I'd better move
cab09cf (hover style changes) to another PR. The change is not related to syntax highlighting and, I'd say, it is very controversial ... there are better ways to emphasize active element. Let's separate into another PR. Otherwise - OK from me.

image

image

image

P.S. `zcl_abapgit_file_status` reordering: looks like no changes ... is it related to syntax highlight ?

mbtools:
Removed hover. 

PS zcl_abapgit_file_status was just a pretty print which reordered the method (which is a different issue).

PR #3388, opened 2020-05-21
Master lang only - clear descriptions for clas, intf

FreHu:
Example removed part:

    \<SEOCOMPOTX>

    \<CLSNAME>cx\_blah\</CLSNAME>

    \<CMPNAME>CONSTRUCTOR\</CMPNAME>

    \<LANGU>닱\</LANGU>

    \<DESCRIPT>\[ĈŎŃŜŢŘŮĈŢŎŘ∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙∙\]\</DESCRIPT>

   \</SEOCOMPOTX>

### DIFF
--- a/src/objects/zcl_abapgit_object_clas.clas.abap
+++ b/src/objects/zcl_abapgit_object_clas.clas.abap
@@ -358,12 +358,15 @@ CLASS zcl_abapgit_object_clas IMPLEMENTATION.
   METHOD serialize_descr.
 
     DATA: lt_descriptions TYPE zif_abapgit_definitions=>ty_seocompotx_tt,
-          lv_master_lang_only TYPE abap_bool.
+          lv_language     TYPE spras.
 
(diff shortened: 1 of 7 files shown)


mbtools:
Hmmm. sy-langu is not necessarily the same as mv_language (the master language in object_clas/intf).  For example, when master language is DE in the repo but you are logged on in EN. 

I think the logic for descriptions should be like for textpool (deserialize/serialize_tpool), where we pass mv_language to ZIF_ABAPGIT_OO_OBJECT_FNC

FreHu:
I see, every object has a mv\_language. I'll change this in the other texts I excluded recently as well.

FreHu:
Fixed. DTEL and DOCT 055d1b8 use the masterlang flag instead of language, so these should be okay.

PR #3392, opened 2020-05-22
Replace assert in gui_page_stage with check

FreHu:
This is connected to the strange bug in #3073 which remains unsolved. Every time I update abapgit on our systems I need to change this line or people eventually run into this dump and I get e-mails about it.

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_stage.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_stage.clas.abap
@@ -601,6 +601,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_STAGE IMPLEMENTATION.
           path     = ls_file-path
           filename = ls_file-filename.
       IF sy-subrc <> 0.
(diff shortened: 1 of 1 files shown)


larshp:
this will give unexpected results to users? some parts of the commit will disappear?

perhaps raise an exception instead?

FreHu:
Yeah, it will skip the file so let's go with an exception instead. Not sure if I should mention files with spaces in the message as the suspect because I can't offer a reliable solution - "it goes away eventually" is not very professional, but I've tried debugging it multiple times and this is in fact how it goes.

larshp:
"staging files failed, reason unknown, try again"

FreHu:
I came up with           

Unable to stage { ls\_file-filename }. If the filename contains spaces, this is a known issue. Consider ignoring or staging the file at a later time.

what do you think?

larshp:
good, but might be too long to display, lets try anyhow

FreHu:
It's not a standard message, longer texts look alright.

PR #3401, opened 2020-05-23
Add currently checked-out commit hash to repo

mariusraht2:
Related to #3040

Adds the currently checked-out commit hash to repo (only attribute and getter and setter).


mariusraht2:
Is dependent on future PRs which are themselves dependent on PR #3321. Will transport these changes in another, future PR then like you suggested - closing.

PR #3403, opened 2020-05-23
Gui refactoring - part 5

sbcgua:
This is more technical chapter. fine tuning. **makes sense to review by commit** - they are dedicated and clear. Otherwise looks like many changes in many places.

1) refactor mi_gui_services to getter - removed obtaining of gui_services in constructor of gui_component - this was causing infinite loop in hotkeys (there is a removed comment in code)
2) zif_html->set_title - this is more for debug. Postponed parts are not anyhow connected to originating class - could be difficult to debug in case of issues. Now they are named. (**for reviewers**: zcl_html was not changed, just `zif_abapgit_html~set_title` method added - other changes are se80 reorderings)
3) gui optional rollback on error - an option to disable rollback on error. It is more to possibility of reusing GUI code in other projects. I added this feature some long time ago then it was removed when html error handler was introduced - getting it back :)
4) docs on the implemented stuff and recommendations on new components, hotkeys, events, blah blah ...

P.S. This part is probably the last for some time.  A potential next step is to completely disconnect gui_page from components and remove subclassing. To do this the following steps are to be done:
- refactor `ms_control` in gui page - create ways to pass page title and global menu from sub-components without subclassing - probably via html parts mechanism (`gui_services( )->get_html_parts( )->add_part(...)`)
- remove all subclassing, make gui page final - a HOC component  - and adjust how pages are instantiated. Not difficult but a lot of small changes here and there.

UPD: P.P.S or it can be done other way around - first separating smaller reusable components from existing pages. E.g. grid view - and reuse it in repo-list, repo-view, stage, commit, ... wherever tables are user

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_commit.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_commit.clas.abap
@@ -392,6 +392,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_COMMIT IMPLEMENTATION.
 
     CREATE OBJECT ro_html.
 
+    ro_html->zif_abapgit_html~set_title( cl_abap_typedescr=>describe_by_object_ref( me )->get_relative_name( ) ).
     ro_html->add( 'setInitialFocus("comment");' ).
 
   ENDMETHOD.
--- a/docs/collections/_development/developing-ui.md
+++ b/docs/collections/_development/developing-ui.md
@@ -9,22 +9,27 @@ This doc covers page creation, html rendering and event handling.
 
 ## TL;DR
 
-- To create a new page in abapGit you subclass `ZCL_ABAPGIT_GUI_PAGE` and redefine `RENDER_CONTENT` method
-- Use `ZCL_ABAPGIT_HTML` to collect HTML content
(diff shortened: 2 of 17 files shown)


src/ui/zcl_abapgit_gui_page_commit.clas.abap
```diff
@@ -392,6 +392,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_COMMIT IMPLEMENTATION.
 
     CREATE OBJECT ro_html.
 
+    ro_html->zif_abapgit_html~set_title( cl_abap_typedescr=>describe_by_object_ref( me )->get_relative_name( ) ).
```
larshp, line 395: does this work in the compiled version?

sbcgua: Yes

image

P.S. found one forgotten place, will add a small commit in a moment

larshp: cool

sbcgua: ok, done. Forgot to add title in hotkeys scripts because it was not unified in the first place

sbcgua:
ref #3286

PR #3426, opened 2020-05-25
CLAS: 702: Fix for missing suppress_dialog parameter

mbtools:


larshp:
what happens in 702, then it will show a dialog?

mbtools:
I didn’t get are popup for classes but for interfaces it will show a popup asking to activate the changes. So you have to press enter once.

larshp:
guess its better than a dump

mbtools:
it's a minor inconvenience but I don't see any other way

PR #3445, opened 2020-06-03
DSYS serializer overhaul

### LINKED ISSUE #3374: abapGit does not serialize DSYS documents when the name contains a namespace-prefix

Elberet:
The serializer class for DSYS objects has a number of issues, including #3374 and two more problems I encountered while working on a fix:
* most notably, in its current version, the serializer will only interact with version `0001` of the text, even if that is not the active version and/or newer versions exist on the system.
* texts are always handled in the current user's logon language, even if that is not the document's master language.

This PR rewrites the DSYS serializer from the group up using the `longtexts` facility, while retaining the ability to deserialize objects written by the current version of the deserializer. Additionally, texts are always opened in their master language when jumping to the text edit tool.

Closes #3374.
Closes two unreported issues as mentioned above.


larshp:
any changes to the serialized XML files, will it also work with older serialized XML files?

Elberet:
Yes, the deserializer checks the version recorded in the XML metadata. For `v1.0.0` it uses more or less the same logic as before (but updates the latest version of the document). New XML files are written with metadata version `v2.0.0`.

larshp:
okay

can you help to submit a 1.0.0 example to https://github.com/abapGit-tests/DSYS_old
and the new serialized XML to https://github.com/abapGit-tests/DSYS_new

so we have it for testing and future reference

Elberet:
Done - see abapGit-tests/DSYS_old#1 and abapGit-tests/DSYS_new#1.

larshp:
@goreraks can you help testing this PR?

goreraks:
In logon langauge EN, imported https://github.com/abapGit-tests/DSYS_old/pull/1, document created in DE language but abapGit shows the difference b/w local and remote. is it the correct behaviour?
image

Elberet:
I'm assuming you're using the old version of `zcl_abapgit_object_dsys`, since the serializer version on your diff's local side is still `v1.0.0`.

If so, you're seeing the effect of one of the additional problems I've mentioned in the PR:

> * texts are always handled in the current user's logon language [...]

Basically, on import the document is created in language DE, but it gets compared to language EN because the serializer always tries to read version 1 in your logon language, regardless of what actually exists on your system.

goreraks:
Tested https://github.com/abapGit-tests/DSYS_new/pull/1, works as expected i.e. including the translation of document.
The behaviour observed in my previous comment is also valid for new version v2.0.0 i.e. if the document is created in different language(Master language = true), after the pull request the object shows the difference.

Elberet:
@goreraks can you post the diff you get after importing the v2 XML with the v2 deserializer?

goreraks:
Sorry, the difference is not similar to version 1.0.0 but the difference is with the DOKL-MASTERLANG.
image

Elberet:
Hm, I'll have a look into that when I'm back in the office next week.

It could be that it's impossible to create a document with a master-language that's different from the current logon language, in which case this would be a user-error because the repo's master language doesn't match the document's master language (altho, in all fairness, I probably should have included an `.abapgit.xml` in the test-data PRs).

Elberet:
Yep, as mentioned above: this happens when the log-on language and document language are different. As long as the master lang in `.abapgit.xml` matches the master lang in serialized documents, this should never result in an actual problem. I'll update the test repo accordingly, but I don't think additional code changes are strictly necessary.

goreraks:
I am not sure but seems the behavior is not in sync with the transport import. For example, if the same document is transported to other system then, masterlang would have been updated as 'X'.
One more thing, I think because of the difference, I can also stage the file which would overwrite masterlang in the remote file.

Elberet:
Yes and no. You can do those things right now, because your abapGit repository is using master language EN. If the repository is setup correctly, then abapGit will refuse to pull the document unless you're logged in with the correct logon language; if your logon language is DE, none of these issues appear.

larshp:
yea, sure, always feel free to raise new issues

thanks both 👍

esjewett:
I believe this broke compilation in 7.31 and below. :(

image

larshp:
thanks @esjewett, please open it as a new issue

"In Release 4.0" hmm?

Elberet:
That's what you get for trusting the linter. 😄 

I think we have a 7.02 system for... reasons - but nothing between that and 7.4x; I'll have a look on Monday.

larshp:
I dont have a 702 system, so dont have that option

will fix the linter, one bug at a time

PR #3475, opened 2020-06-08
Enhanced object list for "Debug Info"

mbtools:
Display supported objects with description, class name, version, and other metadata and includes link to implementing class

image


larshp:
perhaps move the browser information to before the list, 

image

mbtools:
last two lines are deferred script. not sure if that can be moved up

sbcgua:
```
ro_html->add( 'debugOutput("Browser: " + navigator.userAgent + ' &&
      '"<br>Frontend time: " + new Date(), "debug_info");' ).
```
Browser info is added to the end of debug_info div. So just extract supported objects to it's own div in `render_content` and it will be displayed below Browser info

PR #3479, opened 2020-06-10
Changelog news

jrodriguez-rc:
I worked in a change for News functionality, to get the version from APACK and show README, removing the restriction to abapGit repository and making generic.


jrodriguez-rc:
There is a problem with abaplint checks because it is checking the APACK interface if_apack_manifest

mbtools:
Now we can do news for any of our APACK repos. Cool! Could you pls make is_abapgit public so we can use it in other places (maybe move it to ZCL_ABAPGIT_URL)?

jrodriguez-rc:
Updated!

mbtools:
I tested it and it works... under certain conditions. The changelog file must adhere to the same format as used by abapGit. The parsing is quite delicate (i.e. regex). I had a dash between date and version for example and then it won't work. -> We need to document it properly

You changed the filename detection to "changelog*" or "CHANGELOG*". This allows for example CHANGELOG.md (markdown) which is unlikely to work with the parsing. I guess it should stay as .txt.

jrodriguez-rc:
I left as a pattern because GitLab has defined their filenames for readme, contributing, and changelog. For changelog is "CHANGELOG"
image

larshp:
thanks both 👍

jrodriguez-rc:
I created issue #3524 to document it

PR #3486, opened 2020-06-12
Improved validation of URLs

mbtools:
Only http or https are allowed. Other protocols with now cause "Malformed URL" error

Closes https://github.com/larshp/abapGit/issues/3484


mbtools:
needs more testing since some menu links become broken

mbtools:
Now it should work properly everywhere

PR #3498, opened 2020-06-13
Repo overview as main page

### LINKED ISSUE #3269: Redesign home page
Core idea: promote repository overview page to the home page.
- The list can be toggled to only show favorite packages (should be faster to load).
- With no favorite packages, the tutorial can still be shown above the list

image

image

You can also have buttons in the grid for syntax check/code inspector. These don't require serializing the repo so you get a good place from which to quickly check all of your repositories.

Then in the repo page, this part can go away. If you have a lot of favorites, it will start eating up space. The hamburger menu is not very useful either if the system contains a large number of repositories.
image

The top abapgit button can go home instead of the abapgit website, which is already linked to in two other places (help and footer).
image

FreHu:
End goal is for repo overview to be the main page, show tutorial if there are no favorites.
Resolves #3269.

Implemented parts:
- the abapgit logo goes to repo overview
- added main menu to repo overview
- added toggle favorites filter
- removed favorites bar, dropdown

545E97A7-4924-4C54-8F63-A6B16077B659


mbtools:
Lovely... My wish list:
- Add link to package with jump to se80 package tree or ADT
- Add link to branch for "switch branch" popup

FreHu:
>  arguably "Toggle only favs" also not needed if sorted by default

this one i disagree with, on large systems there is a lot of noise in the repository list and you definitely want to hide it.

the repo overview will be merged with the main page as intended

maybe put toggle detail on the right and toggle favs on the left?

sbcgua:
> on large systems there is a lot of noise in the repository list and you definitely want to hide it.

OK, if you see a practical case, let it be so. I think in a long run it should become some kind of tags. But it is a separate feature for sure.

> maybe put toggle detail on the right and toggle favs on the left?

No, no, commands on the different sides of a large screen ... please no :))))) I meant something like this:

image

```html
      <a href="#" onclick="gHelper.toggleRepoListDetail()">Toggle detail</a>
      <span class="separator">|</span>
      <a href="#">Toggle fav only</a>
```
```css
span.separator {
  padding-left: 0.5em;
  padding-right: 0.5em;
  opacity: 0.25;
}
```

FreHu:
There is an annoying issue where the back button will exit abapgit instead of going home.

src/zif_abapgit_repo_srv.intf.abap
```diff
@@ -1,6 +1,10 @@
 INTERFACE zif_abapgit_repo_srv
```
FreHu, line 1: these are unrealated changes, todo remove

src/ui/zabapgit_js_common.w3mi.xml
```diff
@@ -15,7 +15,7 @@
      <RELID>MI</RELID>
      <OBJID>ZABAPGIT_JS_COMMON</OBJID>
      <NAME>filename</NAME>
-     <VALUE>common.js</VALUE>
```
FreHu, line 18: todo remove

FreHu:
Remaining: 

*   what to do with empty favorites? Disable the toggle so this can't be done?
*   abapgit will exit if you go home->repo->back. this is because repo is not a page, but perhaps there is a simpler solution
*   anything i haven't noticed

mbtools:
- When viewing a repo, we use a check mark to toggle for example "Diffs only". Suggest we use the same here 
- No favs: Disable toggle
- It does look weird with the header and then tutorial page. I really liked the original tutorial page, when you start AG the first time. What does it look like on your version, when you click "Help > Tutorial"?

sbcgua:
will have a look tomorrow

src/ui/zcl_abapgit_gui_page_main.clas.abap
```diff
@@ -102,6 +123,36 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_MAIN IMPLEMENTATION.
...
+    CREATE OBJECT ro_html.
+
+    ro_html->zif_abapgit_html~set_title( cl_abap_typedescr=>describe_by_object_ref( me )->get_relative_name( ) ).
+    ro_html->add( 'setInitialFocus("filter");' ).
+    ro_html->add( 'var gHelper = new RepoOverViewHelper();' ).
+
+  ENDMETHOD.
+
+
+  METHOD render_text_input.
```
sbcgua, line 137: maybe move to html_chunks ?

src/ui/zcl_abapgit_gui_repo_over.clas.abap
```diff
@@ -37,24 +51,18 @@ CLASS zcl_abapgit_gui_page_repo_over DEFINITION
...
-      parse_filter
-        IMPORTING
-          it_postdata TYPE cnht_post_data_tab,
+    DATA: mv_order_descending TYPE abap_bool,
+          mv_filter           TYPE string,
+          mv_time_zone        TYPE timezone,
+          mt_col_spec         TYPE zif_abapgit_definitions=>tty_col_spec,
+          mt_overview         TYPE tty_overview.
+
+    METHODS: render_text_input
```
sbcgua, line 60: duplicates the one in main ?

src/ui/zcl_abapgit_gui_repo_over.clas.abap
```diff
@@ -410,55 +436,6 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_OVER IMPLEMENTATION.
   ENDMETHOD.
 
 
-  METHOD zif_abapgit_gui_event_handler~on_event.
```
sbcgua, line 413: Mmmm... I would leave this here. Point is that with new event_handler registration add any component into the event pipe `gui_services( )->get_hotkeys_ctl( )->register_event_handler( me )`. And it makes sense to process events of the component inside the component. Also there would be no need to expose additional public methods.

sbcgua:
minor: maybe separate JS formatting changes into a separate PR ? Would make it easier to review. Just imho, I don't insist

sbcgua:
maybe 2: add a short cut `f` to focus filter from keyboard ... can be a separate feature just an idea

FreHu:
The only thing that makes sense in the second breadcrumb is the repo name, which is already below. But at the same time I would like the remote link to stay visible and it won't fit in the breadcrumb...

FreHu:
I'll start with separating the tutorial like you proposed and then i'll get to the rest of the comments.

sbcgua:
> The only thing that makes sense in the second breadcrumb is the repo name, which is already below. But at the same time I would like the remote link to stay visible and it won't fit in the breadcrumb...

Yeah. And what do you think about the first option ? Remove "+" at repo view and add "Repolist" command there ? On the one hand it is logical - repo management in one place. On the other: it makes no advantage to have "+" only in repo overview page, besides this is one of the main function of AG so should be easily accessible. 

The 3rd option would look something like this

image

But also not ideal.

FreHu:
> And what do you think about the first option ? Remove "+" at repo view and add "Repolist" command there

I'm starting to like this idea more and more.

src/ui/zcl_abapgit_gui_repo_over.clas.abap
FreHu: cleaned up

FreHu:
Latest iteration:



I removed all the other menus from the repo page, none of the advanced functions require a repo to be open.



But I think there should actually be a generic back chevron next to the page title, like it is in the fiori shellbar.

https://experience.sap.com/fiori-design-web/shell-bar/

This would look familiar even allow for a (smaller!) abapgit logo to be on the top.

flaiker:
> But I think there should actually be a generic back chevron next to the page title, like it is in the fiori shellbar.

Do you mean this one?

image



Just thinking it would look weird if you are using the default SAP GUI theme and have two back-chevrons. Like SAP _just_ got rid of the 3 back buttons next two other and split them across the screen.

FreHu:
Ah, the belize theme comes back to bite me. On the other hand, maybe we want the ui to be able to work completely independent of sapgui and a back button is important. 

@sbcgua thoughts?

mbtools:
If and when we run in a browser, then the browser has back and forward controls. No need to put our own.

sbcgua:
> On the other hand, maybe we want the ui to be able to work completely independent of sapgui and a back button is important. 

This is a very huge and separate topic imho because it depends on what will be the tool do display AG ... Persnally I though that UI should either go in the direction of vscode (and the the UI will be fundamentally different) or some kind of "in tray" progressive web app on chrome and react ... not sure. Needs discussion

sbcgua:
Will check the code later today. Seems like a big change :))

In terms of look ... looks nice. `Remove "+" at repo view and add "Repolist" command there` - yeah, appeared to be better than expected, I also like it. I would just keep Advanced and Help at the repo page.

Logo. Once @larshp declined me to remove bottom logo. And now it's the top one :P @larshp what do you think ?

Just in case a couple of new image renders to fit the size

ag-h60
ag-h80
git-h40
git-h60


see also some old renders here https://github.com/sbcgua/abapgit-design. Maybe it's time to do some refresh on logo ?


image

sbcgua:
another-bordered

sbcgua:
image

mbtools:
Sure, there's an evolution of a brand but do we need a logo change? I like it at the top left. Without it, it could be any tool in the world. 

I don't like removing the top menu (online/offline/advanced/help) on the repo view. I fear user will be "where the did xyz go?" plus it adds more steps to get to things. I rather keep this top part of abapGit identical on all pages. 

While we are at it, we should address that we have two "Advanced" menus. 

Suggestion:
- Change title from "Repository List/Overview" to "Home" 
- Keep logo top left (click on logo goes home)
- Replace "Advanced > Repo Overview" with "Home" as the first menu item 
- Move "Advanced > Settings" to new gear as last menu item
- Rename "Advanced" to "Utilities" (or something else)

Result: 
**Home | +Online | +Offline | Utilities | Help |** image



[lol]
remember? I'm still undecided if mysap or sap gold ranks worst...

image
[/lol]

FreHu:
The logo is fine to keep I guess, but with a few notes:

*   shouldn't look ugly in the dark theme, and if we're doing a new one, I'd replace the bottom logo as well (as a separate PR before this is merged? this one is getting a bit too large)



*   I wouldn't make it clickable to go to docs or home, but keep the back link as it is now
    *   Docs are accessible from tutorial and it's just not where you need to go most of the time
    *   as @sbcgua mentioned, it is not obvious that it's a button (and everyone who tried now probably remembers not to click it because it goes to a place they don't need to go to)
*   it doesn't need to be so big

_I don't like removing the top menu (online/offline/advanced/help) on the repo view. I fear user will be "where the did xyz go?"_

Actually, technically the top menu was only accessible from one page - home and repo were one page before I split them, and home was mostly useless. Stage/diff/commit have their own menus and I think this is fine.

The double "Advanced" is why I had the idea to just get rid of all the top menu links on the repo page. 

*   none of the top advanced functions interact with the open repo
*   Help is useful to have everywhere, but it's not in the other page menus either
*   there is no more advanced->repo overview because it is the home page now
*   I like renaming the global advanced to utilities to distinguish them, but i'd still keep the utilities on the home page only

Currently known issues:

*   standard back button leads to dump
*   "show last repo" function needs reworking and cleanup

mbtools:
Already. Sounds good. 

I was already working on the logo for dark theme. I will make it svg, so we can scale it with css to the size we like.

sbcgua:
Uff, much to comment. :) Will separate to keep subjects.

## Logo

According to my measurement the height of logo that fits the current header is 29px. Render here https://github.com/sbcgua/abapgit-design/blob/master/logo/abapGit-orange-29px.png (width = 118px). But better retry in other env - I may have unusual display scale settings. The render has transparent BG so should look fine in dark as well.

SVG is not supported by IE sadly.

Separate PR ? Mmm, I don't know, the header is already changed in this PR so why not add logo change here as well. It would be incomplete without it. Imho.

> I wouldn't make it clickable to go to docs or home

+1

sbcgua:
## Menu

1) "Repository List/Overview" vs "Home". I'd go for "Repository <whatever>". "Home" does not reflect the action essence. Especially speaking about menu command. "Repository list" is clearer. Consequently the title should be similar. Maybe it is possible to find a better word than so long "Repository list" but should clearly indicate that it would switch the overview of available repos.

2) "...Rename "Advanced" to "Utilities"..." +1 Or maybe change to an icon - [this](https://fontawesome.com/icons/tools?style=solid) or [this](https://fontawesome.com/icons/wrench?style=solid) - if OK, I can add to the font quickly.

3) ... and help could become [this](https://fontawesome.com/icons/question-circle?style=regular)

4) I'd keep Utils and Help everywhere. Help actually should be everywhere, in all pages. I'm going to refactor page header a bit to simplify this. So can be unified later.

mmmm ......... 

2020-06-23_19h25_40


5) "none of the top advanced functions interact with the open repo". Yes, but there are flows that could be natural. E.g. I have some issue with *current* repo, I go to data base util. Or (real case) I'm checking the code before deploy to a client, double check everything was commited and no new changes, Advanced -> Package to zip (with master lang only flag). So I'd vote to keep it everywhere.

6) Add online/offline. Here I'm tentative. If to keep them everywhere I'd unite them in a subgroup (Add repo and 2 subitems - online/offile). But I have a feeling that they really belong to the repo list. Actually even not the head menu ! But to the page menu (where "default" and "only favs" are). Not sure. What do you think ?

P.S. If this slows does the delivery, I'd probably suggest a separate PR for menu finetuning. Imho.

sbcgua:
## Minor

- Hotkeys stopped working at repo page, pls add this somewhere in the render_content `gui_services( )->get_hotkeys_ctl( )->register_hotkeys( me ).`
- Now when repo view is a "page" I'd rename the class (view_repo) to `zcl_abapgit_gui_page_repo`. But can be done separately not to mess it now.

FreHu:
Random tip: git has no problems with renames
image

As for the logo, I will just put it back and whatever version gets used will get used.

mbtools:
I like the idea of icons. Simple but effective.

image

1. Repositories 
- Overview
- Add Online
- Add Offline
2. Utilities
3. Help

larshp:
@sbcgua any thoughts on merging this, if only minor issues remain can it be moved to a new issue?

sbcgua:
@FreHu: do you plan to add/rollback anything in this PR ? I'd say we can merge and fix the remaining things separately and maybe in parallel.

P.S. this should also unlock merge for #3540
P.P.S @larshp from me OK to merge and go forward, but just if Frederik says he finished with the current set of changes

larshp:
cool, lets await a status from @FreHu

FreHu:
I still have these to do:

*   [ ] fix show last repo
*   [ ] fix dump with back buttons
*   [ ] fix hotkeys at repo page
*   [ ] add back logo

The rest can be done separately, this is getting too big. Feel free to merge anything before me, I'll resolve any potential conflicts.

FreHu:
All done. I also rolled back the formatting. Follow-ups:

*   icons in main menu + add menu back to repo page
*   logo changes
*   show tutorial page if no package is installed
*   main page filter is not remembered if you restart abapgit - session storage is used to store it. possibly it's enough to change to local storage

sbcgua:
quick ... :) :+1: I'll have a look today later

sbcgua:
@FreHu @mbtools FYI, added new icons to the font for future works

image

FreHu:
I added one more commit and i don't see it appearing in the PR.



I also get this when trying to update from master:



Any idea what's going on?

flaiker:
Seems like some kind of outage occurred: https://www.githubstatus.com/

Can you push another (dummy) commit?

PR #3528, opened 2020-06-19
Interface and implementing class in same package

mbtools:
Closes https://github.com/larshp/abapGit/issues/3473


mbtools:
Changing package assignment leads to many messages:

image

Should we simplify this first in a separate PR and replace with one message per object that says "Object xyz has been assigned to a different package. Run Pull/Stage to commit the change."?

larshp:
the files could have been moved in the remote git repository to an inconsistent state, in that case the message "Object xyz has been assigned to a different package. Run Pull/Stage to commit the change." would be misleading

mbtools:
https://github.com/larshp/abapGit/pull/3529

PR #3529, opened 2020-06-19
Condense messages in case of package move

mbtools:
If an object is assigned to a different package, abapGit shows multiple warning messages per object and file:

image

In case the files are identical in old and new folder (same SHA1), the messages are reduced to one per object:

image


FreHu:
:+1:  I'd also include package from/package to in the message.

mbtools:
How do you know which side is newer ie the to value?

Best,
Marc

FreHu:
Actually I think you can't, I didn't realize. Looks okay as it is currently then.

larshp:
add unit tests?

Issue #3417, opened 2020-05-24
Convert popups to pages


sbcgua:
Following the discussion in slack. Try creating dedicated pages for add online/local repo. Maybe for password popup as well.

NB:
- try to avoid major JS on such pages - this is the core flow that must work e.g. in java client
- maybe cache previous page for the cancel action, not to rerender
- password html page may open convenient possibilities to integrate with password managers (e.g. to enter github api key)
- what to do with package selection/creation ? rerender the page with new value ? Probably ok as popup pages will be simple

sbcgua:
some early design wip

![2020-06-08_11h09_39](https://user-images.githubusercontent.com/15635498/84007554-4612de00-a979-11ea-9999-6be298ab53ee.png)

larshp:
vertical better IMHO

sbcgua:
getting closer

![image](https://user-images.githubusercontent.com/15635498/84578311-cba8ea80-adcc-11ea-9ed4-7178ca49d071.png)

larshp:
👍 this is so much better than the SAP standard popup, plus we have better control of the page flow, plus no random DDIC dependencies

mbtools:
Great! Let's visually separate repo from local settings.

sbcgua:
@larshp is there an util for displaying existing search help in AG ? For usage like this: https://github.com/sbcgua/abapGit/blob/dialog/src/ui/zcl_abapgit_gui_page_addonline.clas.abap#L249

larshp:
I dont think so

abapGit is a tool for developers, so I think its okay to leave some of the search helps out in the first versions of this issue

sbcgua:
tnx.

One more subject.
I have an issue with branches. Do we need this field in this dialog ("add online" + potentialy "edit remote") ? Problem is that we need to know URL to show the branch selection. And it is tricky to do without JS (possible but crutchy). I'd probably prefer that branch is selected afterwards and it is always "master" at this point.

On the other hand, if branch is needed. Can it be a feature for JS clients only ?

And one more: branches - is it supposed to be HTML dialog too ? or abap popup is OK ?

larshp:
just make it an abap popup for now, that would be the easiest? we need some more experience in this area to figure out where to go

sbcgua:
[...]

PR #3533, opened 2020-06-21
Dialogs: add online repo HTML form

sbcgua:
OK, I think it can be tested and reviewed. Generally the code is ready. ~~I just want to try one tweak (string_map for form values below **DONE**)~~. The PR is rather large, apologies, but centered mostly in the new stuff. #3417

image

## What's inside

- html form abstraction `zcl_abapgit_html_form` that should be reusable in other places and simplify the page code
  - I will probably write a doc on it later. But example can be seen `ZCL_ABAPGIT_GUI_PAGE_ADDONLINE->RENDER_CONTENT` - clean and readable 👌 
  - dblclick on text field = side actions button
  - the `render` method accepts root css class which should allow to render same form differently just with css (if needed)
  - look at the validation approach ! Separate method collects validation message into a string map (`zcl_abapgit_string_map`) which is then uniformly processed by the html_form.
  - what I want to tweak (above) is to use same approach for values, no need to pass them from a structure explicitly imho. So inputs will need just `iv_name` and the values will be passed to `render` as a string map. Thus the form is a schema which is then mixed with values and validation errors at render time. **UPD: DONE**
- `zcl_abapgit_string_map` a primitive for key-value map. I probably invented a bicycle ... but it's quite convenient (there probably should be something like that in standard abap ... but I don't know)
- `zcl_abapgit_services_basis` - for scripts not related to repos, git or abapgit ... create package for now, maybe something else in future ? Transports ? whatever...
- `zcl_abapgit_gui_page_addonline` - the hero of the occasion
- `zif_abapgit_popups~popup_search_help` quick util to show search helps
- *important !* `repo_services->new_online` does not call popups anymore
- *change !* `zif_abapgit_repo_srv~new_online` accepts branch names without `refs/heads` (maybe I broke something ?? but seems to work)

## Considerations
- the solution is JS free (except dblclick handling) ... and it was hard ... 
- issue is that certain **secondary** actions on abap side would need data from the form. Thus **all** actions must be submits. Agh !!! 🤦 OK, it is kind of solved. But e.g. side actions (buttons next to text fields) are also submits, and they behave strangely in HTML. I don't like the look but this is as far as I could get it without completely crippling the html (which is already crippled a bit by unnecessary divs). Well. Maybe someone knows CSS better and can improve. The initial version (including screenshots from #3417) used `anchors` and it was more controllable and better looking. But they could not submit the form without JS. life's hard. I need some beer today. 🍺 or 🍻 

## Not to forget someday
- remove `zif_abapgit_popups=>ty_popup` when (if) all dialogs move to html, use `zif_abapgit_services_repo=>ty_repo_params`

@larshp @FreHu @christianguenter2 @mbtools and other - welcome to review

### DIFF
--- a/src/ui/core/zcl_abapgit_html.clas.abap
+++ b/src/ui/core/zcl_abapgit_html.clas.abap
@@ -30,6 +30,11 @@ CLASS zcl_abapgit_html DEFINITION
         !iv_hint    TYPE string OPTIONAL
         !iv_class   TYPE string OPTIONAL
         !iv_onclick TYPE string OPTIONAL .
+
+    CLASS-METHODS create
+      RETURNING
+        VALUE(ro_html) TYPE REF TO zcl_abapgit_html.
+
   PROTECTED SECTION.
   PRIVATE SECTION.
 
@@ -211,6 +216,11 @@ CLASS ZCL_ABAPGIT_HTML IMPLEMENTATION.
       EXPORTING
         pattern     = '<(AREA|BASE|BR|COL|COMMAND|EMBED|HR|IMG|INPUT|LINK|META|PARAM|SOURCE|!)'
         ignore_case = abap_false.
+  ENDMETHOD.
+
+
+  METHOD create.
+    CREATE OBJECT ro_html.
   ENDMETHOD.
 
 
--- a/src/ui/zabapgit_css_common.w3mi.data.css
+++ b/src/ui/zabapgit_css_common.w3mi.data.css
@@ -3,125 +3,6 @@
  */
 
 /* GLOBALS */
-.w500px { width: 500px }
-.dialog input::placeholder { color: #ccc }
-.dialog input:-ms-input-placeholder { color: #ccc }
-
-.dialog {
-  margin: 1em 1em;
-  padding: 1em 1em;
-  /*border: 1px solid #00000015;*/
-  border: 1px solid #cbcbcb;
-  border-radius: 4px;
-  /*background-color: #ffffff60;*/
-  background-color: #f4f4f4;
-  text-align: left;
-}
-.dialog li {
-  padding: 5px 10px;
-  display: block;
-  list-style: none;
-  position: relative;
-}
-.dialog button.field-cmd {
-  float: right;
-  position: absolute;
-  top: 0px;
-  right: 0px;
-  cursor: pointer;
-}
-.dialog li.dialog-commands {
-  text-align: right;
-  margin-top: 12px;
-}
-.dialog li.dialog-commands a {
-  /*background-color: #eaeaea;*/
-  border: 1px solid #ccc;
-  cursor: pointer;
-  text-decoration: none;
-  padding: 6px 12px;
-  border-radius: 3px;
-  font-size: smaller;
-}
-.dialog li.dialog-commands input[type="submit"] {
-  padding: 6px 12px;
-  border-radius: 3px;
-  cursor: pointer;
-  border: 1px solid #ccc;
-}
-.dialog li.dialog-commands input[type="submit"].main {
-  background-color: #64a8ff;
-  border: 1px solid transparent;
-  color: #fff;
-}
-
-.dialog label {
-  display: block;
-  font-size: 90%;
-  color: #444;
-  margin-top: 6px;
-  margin-bottom: 6px;
-  padding-left: 0.5em;
-}
-.dialog label span{
-  color: #ccc;
-  float: right;
-  padding-right: 0.5em;
-}
-.dialog small {
-  color: #999;
-  display: block;
-  font-size: 75%;
-  margin-top: 4px;
-}
-.dialog input[type="text"] {
-  /*line-height: 1.5em;*/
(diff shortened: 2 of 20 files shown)


sbcgua:
ready for review

g-back:
1. The UI is a bit broken for me:
image

2. Change "Master language only" to "Serialize Master language only" - I know that there is a tooltip but now that we have space that should clarify the meaning on first glance

sbcgua:
1) Interesting ... this is those submits ... could you tell what kind of screen you have and screen scaling % ? (I mean Windows settings -> Display -> Scale and layout)

2) Agree, will fix

mbtools:
this is awesome! the side buttons work for me. not sure what's wrong there.

I made a couple enhancements (see https://gist.github.com/mbtools/043d8ee6e449bc6576412963815f1957, not very git-like, i know)
- centered
- grouping of elements (optional in general but I added two groups here)

image

Suggestions:
- make form id mandatory
- add width as an optional attribute since it probably varies between popups (default 500px)

Issues:
- When you click on branch selection and the URL is not a valid git repo, then you get "HTTP 404" error. Needs a proper message.
- I really miss the field history for the URL (alt+down). We need the feature to switch repo URLs easily (as suggested somewhere else)
- Value help for package always shows in top left corner. We might have to use a callback in F4IF_FIELD_VALUE_REQUEST
- When you enter an invalid package, it shows an error (good). When you then use the value help to pick a valid package, the error remains. Should trigger validation after value help.

We're almost there. Thx Alexander!

mbtools:
These forms render so nicely, we should switch not just popups but also AG > Settings and Repo > Advanced > Settings to this.

sbcgua:
@g-back 
- tried different scaling but could not recreate. Anyway, please try again, I increased the padding a bit, hopefully it should work.
- *Serialize* Master language only - done

@mbtools, thanks for the suggestions ! :) Though I'd not agree with some of them
- centred - not really sure. It looks rather lonely on wide screens. **Need more opinions.**

image
image

- groups - nice feature, I'll add it, but in a separate PR, but will implement it differently and probably not for this dialog - there are just a few inputs, groups would make it more cluttered. Groups needed when there are more imho. But I rearranged the fields - folder logic before display name.

image


- "make form id mandatory" - why ? If there is not JS interaction with form, it is not really needed. Maybe it makes sense to autogenerate it randomly if not given - for double clicks to work. I'll think about it.
- "add width as an optional attribute" - I'd disagree. Form is placed in a container. The container can have certain width or height or red running dashed border - whatever. It is the style of the container. So it is better to be regulated by CSS and not by hardcoded style.
- "HTTP 404" - do you mean the one below ? I'm not sure it should be displayed as a validation error (in the form itself). It can be long. It can be some specific error which might need access to stack and all the stuff that is shown in the exception block. Not sure we have to catch all the exceptions and show in the field context.

image

- "field history" - yeah that might be an issue :( but not sure how to solve in this context. "switch repo URLs easily" - this is a separate feature.
- "Value help for package always shows in top left corner" - hmm, yeah. Do you know how to fix it ? I'm afraid I'm not very good at screens stuff :(
- "When you ... pick a valid package" - yeah good catch, done

Thanks for the comments !

flaiker:
> centred - not really sure. It looks rather lonely on wide screens. Need more opinions.

Agreed, everything else in SAP GUI is left aligned and there even is a maximum width of 255 columns for dynpros (which is like 2/3 of the screen width on 1440p with the already larger rendering Belize theme, I assume it's 1/2 on the other ones).

larshp:
@g-back do you still have problems showing the form? which SAP gui version are you running, any java GUI or something?

works for me,

image

g-back:
Can only test tomorrow, will report

flaiker:
Seems to be a theme related issue.

### Default

image

### Belize

image

flaiker:
One minor issue I noticed is that neither the value requests nor the create package popup get prepopulated with the input of the form.

mbtools:
Belize has only color changes (as it should be). It’s missing CSS from the default theme. As I wrote: CSS refactoring required. 

Best,
Marc

flaiker:
Font and font size are different in the Belize theme though, which might tip things.

g-back:
I can confirm what @flaiker said, issue is caused by the Belize (and Dark) abapGit theme.

sbcgua:
@flaiker @g-back

- Belize issue: got it, thanks. Fixed. Actually made it in a more reliable (though a bit cluttered) way. Now should work in all themes reliably because should doesn't depend on margins
- "create package popup get prepopulated" - fixed
- "value requests ... prepopulated" - sorry didn't get this one

larshp:
will merge this one tomorrow morning(Europe timezone)

mbtools:
FYI, the new form page contains invalid HTML: "Error: Element form not allowed as child of element ul" (there shouldn't be anything between UL and LI). Nevertheless, it seems to work fine and you can merge

larshp:
@sbcgua can you help updating the branch, looks like I dont have access

sbcgua:
@larshp rebased

larshp:
thanks 👍

PR #3540, opened 2020-06-22
UI: Fixes for dark theme

mbtools:
- abapGit color & background
image

- color for error messages (closes https://github.com/larshp/abapGit/issues/3487)
image

- darker color js debug info
image


FreHu:
There's a few more that bother me:

*   remote url is unreadable
*   borders are too white

FreHu:
There's a few more that bother me:

*   remote url is unreadable
*   borders are too white

sbcgua:
also branch border is rather light. 

I would also review the default text color, blue tint is a bit destructing. More common solution is either pure grey or a warmer color rather than colder. (light brown)

mbtools:
The css needs refactoring. Without that even small changes take a lot of trial&error. 

I will see what I can do

sbcgua:
maybe wait with the logo a bit, see latest comments in https://github.com/larshp/abapGit/pull/3498#issuecomment-648246966

mbtools:
Url, table borders, branch box border:

image

I didn't change the font color. The blue tint is for links, not text. Since you can click on mostly everything, you see a lot of "blue". Not sure which "link color" is good for dark theme.

larshp:
@FreHu & @sbcgua any comments, can this be merged?

sbcgua:
I'd propose merging it after #3498 (unless @FreHu is OK with it), 'cause they will conflist in css changes.

I'd also suggest separating the logo change. The design proposed here is good. But the only issue with the current logo is that it is not transparent. Changing it as I proposed in #3498 should fix the issue **imho** instead of adding the complexity.

2020-06-28_10h11_50
2020-06-28_10h12_20

P.S. sorry for fonts in dark screen shot - this was some old css to make it quickly

mbtools:
This PR currently changes logo for dark theme to have light font color. @larshp the alternative would be to use the same font color for default and dark theme as suggested above (which would not work for a "red" theme, haha ).

larshp:
@sbcgua to be merged?

PR #3564, opened 2020-06-29
Special characters in object/file names

mbtools:
Escape certain characters in file names to avoid issues with mapping between objects and files (independent of object type).
Closes https://github.com/larshp/abapGit/issues/3092

Test repo with various W3MI object names
https://github.com/mbtools/abapgit-test-esc-filename


larshp:
@mbtools you should now have access to create test repositories in https://github.com/abapGit-tests

mbtools:
It worked. Thank you

larshp:
looks like this changes the casing of filenames, in this example https://github.com/larshp/abapGitServer, but I guess there is the same problem with https://github.com/abapGit-tests/SICF

image

larshp:
perhaps just handle the characters that are not allowed as filenames on windows, ie. without space

mbtools:
I removed escape of space since spaces were allowed before and not causing problems (apparently). 

The problem with SICF is not caused by this PR but a bug in ZCL_ABAPGIT_OBJECT_SICF=>read_sicf_url where the hash is returned in lower case (by zcl_abapgit_hash=>sha1_raw). I will create a separate PR.

larshp:
hmm, yea, at least I cannot reproduce the error with SICF upper case anymore

mbtools:
https://github.com/larshp/abapGit/pull/3449 removed the TADIR fallback which kicked in if the sicf hash was not found. I believe the hash lookup never worked (since it was lower case) and therefore the fallback returned the tadir. 

Overall, sicf is a mess but with this and https://github.com/larshp/abapGit/pull/3600 should be working ok.

larshp:
ship it 🚢

PR #3567, opened 2020-06-30
Logo css cleanups and (maybe) orange logo

sbcgua:
This follows my suggestions from #3498

The PR contains:
- minor CSS cleanups (pls check - potentially may look differently in different scaling - I have 200%)
- and also orange logo variant which looks OK in both light and dark modes. **If not OK** - I'll reverse this point, it is a separate commit - and then we can go towards double logo variant suggested in #3540

2020-06-30_12h56_52
2020-06-30_12h57_11


g-back:
Looks good in Belize with 100% scaling:

image

sbcgua:
Thanks @g-back !

P.S. By the way the proportions text-icon definitely differs ... probably not solvable without SVG support or image re-scaling. Maybe not that critical also.

sbcgua:
Canceling this for now, there is an idea how to improve logo fundamentally, so on hold for the moment

PR #3578, opened 2020-07-02
CSS and logo fixes

sbcgua:
- logo is now vector and css controllable - migrated to icon font. Thanks @larshp for the idea in slack. @mbtools now it's easier to redefine the color for the dark theme
- internal refactoring of header and footer layouts - visually the changes are minor, mainly in footer

image

image


mbtools:
The idea with icon font was great. Thx

Just for future projects, the SAP GUI browser does support SVG. Here are a couple svgs I added for fun to the debug page: 

image

image

sbcgua:
Interesting. What GUI version do you have ? I certainly remember I was trying to make it work in AG a couple of years ago. But it didn't. I tried `<svg>`, `` and misc content stuff with css. So might be some combination of newer software. But thanks for the info.

mbtools:
7.60 p4. I couldn’t get it to work as attached asset but inline it works. It had to work since IE supports svg since IE8.

sbcgua:
🤷‍♂️

PR #3581, opened 2020-07-04
Fix open url in repo view

jrodriguez-rc:


jrodriguez-rc:
You're right, I moved to gui_router and deleted from other classes

Issue #3559, opened 2020-06-29
Potential zcl_html_form improvements


sbcgua:
Followups and potential improvements from #3533

- [X] grouping of elements (for particularly large forms if even happen)

<details>
  <summary>experiments</summary>

```css
fieldset {
  border: none;
  border-top: 1px solid #ccc;
  margin-top: 8px;
}
fieldset legend {
  color: #ccc;
  font-size: small;
  text-transform: uppercase;
  padding-left: 1em;
  padding-right: 1em;
  margin-left: 0.3em;
}
```
![image](https://user-images.githubusercontent.com/15635498/85991142-ae466400-b9fb-11ea-9837-ea45bd9ad5fa.png)
</details>

- [ ] implementation of dropdowns (though better to avoid them and use radio buttons instead - [good talk](https://www.youtube.com/watch?v=hcYAHix-riY&feature=emb_logo) on this subject)
- [X] auto-generated form id of not passed (for double clicks to always work for side actions)
- [ ] maybe possibility to add several side actions
- [X] adjust markup: bring ul inside form, use div instead, adjust css respectively (last comment from PR)
- [X] move colors to theme
- [X] dev docs
- [x] integer field with min/max (from #3904)

sbcgua:
I'll leave this for a while after #3582. The remainings are low prio nice-to-haves imho. So if needed, feel free to implement the remaining features or other features.

sbcgua:
- [ ] Unified validation (bool and uppercase can be unified too)

mbtools:
I don't think we need dropdowns or multiple side actions. Close?

sbcgua:
Yeah. Seem to be fine so far.

PR #3582, opened 2020-07-04
Html form improvements and docs

sbcgua:
to #3559

- auto generate form id, if not passed
- minor html changes to confirm validation (thank @mbtools for a hint)
- move colors to theme css (default only)
- implement form groups (below screenshot is test - **it is not implemented, and I'd not recommend adding groups to repo dialog**)
- development docs ... it's a duty, yeah ...

P.S. Not that I like the design of groups a lot, leaves room for improvement, but maybe will find it's usage. In large forms. By the way good designs seem to avoid groups, at least its native way. [1](https://uicookies.com/beautiful-css-forms/), [2](https://freefrontend.com/css-forms/)

image


mbtools:
Thanks for the continued improvements. Looks great. The grouping will be good to have when we replace the settings pages with nice forms. 

Interestingly, almost all forms shown in your 1,2 links are centered ;-).

sbcgua:
> Interestingly, almost all forms shown in your 1,2 links are centered ;-)

Centered is usually good. :) But it has to be observed in a context. My point with add-online dialog was that "center" will not add value. Look at this

image

It's ugly. I have a very wide screen, so let's try a normal one.

image

Not much better. I almost have to turn head to look around. Luckily we don't have commands in the upper right corner - it would be ugly and inconvenient at the same time. Frankly the left alignment *is not much better*, I confess. OK, there laptops, laptops have phisically smaller screens ... But it is still not the root of the problem. A more proper step would be to narrow down the whole view area, then the components would be nearby and consistent and center alignment would fit of course too (I actually discussed it with Frederik somewhere here before, so it is in some plans ... maybe :)

So imho no need to fight for the alignment. Actually I left the css to centralize the dialog and a related comment in `zcl_abapgit_gui_page_addonline->render` so feel free to adjust it if you like :)

mbtools:
I appreciate the exchanges. Lars, Fred and you have been with this much longer so I’m not going to push “my way”.

 Doing responsive design that works on small and large screens is possible but requires more css. Not saying that we need a mobile UI now but I could see some processes that ppl want to run via their phone (approving pull or merge requests). Anyway, step-by-step improvements is what we go for. :-)

PR #3590, opened 2020-07-06
Top menu improvements

FreHu:
@sbcgua 

*   Use icons
*   Add advanced, help back to repo page
*   Add menu to tutorial page
*   Refactoring

### DIFF
--- a/src/ui/zabapgit_icon_font_css.w3mi.data.css
+++ b/src/ui/zabapgit_icon_font_css.w3mi.data.css
@@ -59,7 +59,3 @@
 .icon-snowflake:before { content: "\f11e"; }
 .icon-star:before { content: "\f11f"; }
 .icon-tools-solid:before { content: "\f120"; }
-
-.icon-cloud-upload-alt {
-    padding-right: 5px;
-}
--- a/src/ui/zabapgit_icon_font_css.w3mi.data.css
+++ b/src/ui/zabapgit_icon_font_css.w3mi.data.css
@@ -59,3 +59,7 @@
 .icon-snowflake:before { content: "\f11e"; }
 .icon-star:before { content: "\f11f"; }
 .icon-tools-solid:before { content: "\f120"; }
+
+.icon-cloud-upload-alt {
+    padding-right: 5px;
+}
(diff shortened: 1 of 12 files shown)


sbcgua:
Cool 👍 On the first glance the code looks well. Will test it later.

P.S. we certainly need to refactor `super->zif_abapgit_gui_event_handler~on_event` ... (separately, just not to forget)

g-back:
I'm no UX designer, but maybe it would also be worth to add icons for "Repo list" (maybe something like a "home" icon?) and "+Offline" and "+Online" to keep the look unified?

I also tested the changed menus and it all works as it should.

mbtools:
+1 @g-back 

we already have the icons for online/offline/list repos that users are familiar with: https://docs.abapgit.org/ref-icons.html

I would keep the "Online", "Offline", "Repository List" texts since those are the core features but add icons.

FreHu:
\+ Online looks weird with the icon, so i changed it to _new_. This means i'll need to update the tutorial and probably some places in the documentation

FreHu:
Tutorial updated:

src/ui/zabapgit_icon_font_css.w3mi.data.css
```diff
@@ -59,3 +59,7 @@
 .icon-snowflake:before { content: "\f11e"; }
 .icon-star:before { content: "\f11f"; }
 .icon-tools-solid:before { content: "\f120"; }
+
+.icon-cloud-upload-alt {
```
sbcgua, line 63: This file is autogenerated (https://github.com/abapGit/icon-font), so better to fix it on common css.
I'd also try to unify it, it is strange to have global padding for one specific icon. Did you mean "icon in anchor"?
```css
a i.icon { padding-right: 5px }
```
just beware that it may have to reconcile with other icons near links (e.g. repo menu)

FreHu: What I was trying to do with this is that this icon was slightly wider, it looked like there was no space.

sbcgua: Ah, I see. It makes sense then. then the style is OK, just move it to commons pls.

larshp:
@FreHu check conflicts

FreHu:
@larshp now okay to merge

larshp:
thanks 👍

PR #3591, opened 2020-07-06
Add deleted objects to overwrite confirm popup

mbtools:
Previously, "reset local" would automatically recreate objects that had been deleted locally. Now user has to confirm these cases on the overwrite popup. 
Includes new unit tests for overwrite popup and closes https://github.com/larshp/abapGit/issues/2845


mbtools:
Updated popup text:

image

PR #3598, opened 2020-07-07
Fix issue 3597

shaiSAP:
When abapGit is being executed on SAP GUI for HTML, all the links don't work (don't do anything because SAPEVENT isn't being passed back to the application server).


FreHu:
It seems you made your change in the zabapgit report, this won't work as it is compiled from the classes. For a pull request you must find the class in the development version and make the change there.

mbtools:
Hi Shia, I had put a link to the correct class file in the issue you posted. Close this and put the changes into that class. Thx

Issue #3597, opened 2020-07-07
Links Don't Work ON SAP GUI for HTML


shaiSAP:
When abapGit is being executed on SAP GUI for HTML, all the links don't work (don't do anything because SAPEVENT isn't being passed back to the application server).

From technical point-of-view, the issue seems to occur due to the implementation of method zif_abapgit_gui_services~cache_asset for HTML files (method cache_html).
In standard, when being executed on SAP GUI for HTML, the method CL_GUI_HTML_VIEWER->LOAD_DATA parses the HTML and converts SAPEVENT links into external ones.
However, this seems to work when a text table is passed to data_table, but not with binary data (like is done in method cache_asset).
Hence, the code of method cache_asset should be adjusted to set a text table (W3HTMLTAB) instead of binary table when iv_text is passed.

mbtools:
Thx for the analysis! Did you test your suggested change? https://github.com/larshp/abapGit/blob/master/src/ui/core/zcl_abapgit_gui.clas.abap#L414

shaiSAP:
@mbtools ,
Yes, the following seems to do the job:
`    IF iv_text IS SUPPLIED. " String input
      DATA lt_html TYPE w3htmltab.

      zcl_abapgit_convert=>string_to_tab( " FM SCMS_STRING_TO_FTEXT
         EXPORTING
           iv_str = iv_text
         IMPORTING
           et_tab = lt_html
      ).

      mo_html_viewer->load_data(
        EXPORTING
          type         = iv_type
          subtype      = iv_subtype
          url          = iv_url
        IMPORTING
          assigned_url = rv_url
        CHANGING
          data_table   = lt_html
        EXCEPTIONS
          OTHERS       = 1 ) ##NO_TEXT.
    ELSE. " Raw input
      zcl_abapgit_convert=>xstring_to_bintab(
        EXPORTING
          iv_xstr   = iv_xdata
        IMPORTING
          ev_size   = lv_size
          et_bintab = lt_xdata ).

      mo_html_viewer->load_data(
        EXPORTING
          type         = iv_type
          subtype      = iv_subtype
          size         = lv_size
          url          = iv_url
        IMPORTING
          assigned_url = rv_url
        CHANGING
          data_table   = lt_xdata
        EXCEPTIONS
          OTHERS       = 1 ) ##NO_TEXT.
    ENDIF.`

shaiSAP:
I've sent a pull request.

PR #3601, opened 2020-07-08
Fix issue 3597 (#2 try)

shaiSAP:
When abapGit is being executed on SAP GUI for HTML, all the links don't work (don't do anything because SAPEVENT isn't being passed back to the application server).


mbtools:
Tested and works beautifully. Thanks!

image

larshp:
#3597

larshp:

PR #3615, opened 2020-07-12
gui_page cleanups

sbcgua:
ref #3286

Move test_changed_by from gui_page to services

P.S. call_browser is a reorder - no change there


sbcgua:
forgot a couple of event. now ready

PR #3626, opened 2020-07-13
Page HOC concept (refactoring)

sbcgua:
ref #3286

OK, here is one refactoring I wanted to come to for some time and mentioned in #3286. Currently just for one page as a demo. Eventually I'd suggest to migrate all pages to this approach.

Now all pages subclass `gui_page` which is cluttered: trickier to maintain, necessity to call `super`, confusion between `renderable~render` and `render_content`, redefinitions.

Idea: all generic page stuff - html head, footer, scripts - has to be a HOC (high order component). Which accepts a child component to render inside itself.

Consider changes to `zcl_abapgit_gui_page_addonline` - no redefinitions, no `super->on_event`,  proper reuse of interfaces,  completely independent and clean. Well, `create` knows about the gui_page but it is more for convenience sake.

`zcl_abapgit_gui_page_hoc` is a temporary piece - as soon as all pages will follow the proposed pattern it can be merged with `gui_page` (which will become non-abstract final class)

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_addonline.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_addonline.clas.abap
@@ -1,28 +1,25 @@
 CLASS zcl_abapgit_gui_page_addonline DEFINITION
   PUBLIC
-  INHERITING FROM zcl_abapgit_gui_page
+  INHERITING FROM zcl_abapgit_gui_component
   FINAL
   CREATE PRIVATE .
 
   PUBLIC SECTION.
+    INTERFACES zif_abapgit_gui_event_handler.
+    INTERFACES zif_abapgit_gui_renderable.
 
     CLASS-METHODS create
       " TODO importing prefilled form data
       RETURNING
-        VALUE(ro_page) TYPE REF TO zcl_abapgit_gui_page_addonline
+        VALUE(ri_page) TYPE REF TO zif_abapgit_gui_renderable
       RAISING
         zcx_abapgit_exception.
-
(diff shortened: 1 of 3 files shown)


mbtools:
I like it. "HOC" seems a bit abstract. I would call it LAYOUT. The current HOC/LAYOUT would be a "full-page" i.e. a container for one child component. but we could have other layouts like "main and side panels" (two children) or "grid" (several kids) etc. 

Maybe wrap the child content in a div for css:

ri_html = '<div class... id... >' && mi_child->render( ) && </div>.

sbcgua:
HOC - it is from react and js world - very elegant concept - https://reactjs.org/docs/higher-order-components.html
It does not much matter how it is called though because this class is temporary - should merge with gui_page someday.

wrap in div: potentially ... but I'd wait for a practical application case. Currently the code replicates `render_content` logic 1-to-1. Most (or all) of pages return a single tag currently. Also I don't thin that container should impact styles of subcomponents (css class). So let's really wait for practical need.

sbcgua:
@FreHu what do you think ? :)

FreHu:
Definitely looks simpler, and I had pretty much all the confusions you mentioned. As for how frontend is done these days, I don't really know as my FE career went from jquery to a frontend running inside abap backend :)

sbcgua:
@FreHu: :+1: thanks !

P.S. FE carrer: yeah, I'm far from expert in this area actually, just picking nice things here and there trying to make stuff convenient and ordered :)

Issue #3628, opened 2020-07-14
ENHS: Error message is not helping to find the issue


ed-holland:
With some projects i get this message.
![image](https://user-images.githubusercontent.com/53995991/87421291-fc9c5b00-c5d6-11ea-83fb-ddbcea623028.png)

It is unclear what the problem is.

Identified class ZCL_ABAPGIT_OBJECT_ENHS.

The message is used 3 times in different spots.

PR #3629, opened 2020-07-14
Make error messages from ENHS clearer

ed-holland:
Issue #3628

### DIFF
--- a/src/objects/zcl_abapgit_object_enhs.clas.abap
+++ b/src/objects/zcl_abapgit_object_enhs.clas.abap
@@ -58,6 +58,13 @@ CLASS ZCL_ABAPGIT_OBJECT_ENHS IMPLEMENTATION.
 
     DATA: lv_spot_name  TYPE enhspotname,
           li_enh_object TYPE REF TO if_enh_object.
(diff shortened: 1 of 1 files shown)


mbtools:
Works, but wouldn't this work too?

```
CATCH cx_enh_root INTO lx_enh_root.
  zcx_abapgit_exception=>raise( iv_text = lx_enh_root->get_text( )   " or get_longtext( )
                                ix_previous = lx_enh_root ).
```

ed-holland:
The text would not show the exception from what it is raised. It would be generic.
The ix_previous helps if you do it in the gui as one can go to the originating exception.
In background, you only have the message.

It is a bit more work, but the issue became much clearer to me. I was having problems with an ENHS from SAP (which does not have an active version). It really raised an IO Error.

mbtools:
Try this for fun https://github.com/mbtools/abapgit-test-exceptions ;-)

ed-holland:
With the help of your example, i was able to create this:
image

image

Is there an easier way to get the name of the exception?

ed-holland:
Closing as it is replaced by #3651

PR #3632, opened 2020-07-14
TOBJ: Clear package

mbtools:
Closes https://github.com/larshp/abapGit/issues/3627


mbtools:
done. SCP1_TOBJ now installs without diff. Uninstall works too

PR #3637, opened 2020-07-15
Condense messages in case of package move (2)

mbtools:
Follow-up to https://github.com/larshp/abapGit/pull/3529
Mark remote files as processed in the status calculation to avoid additional messages.


sbcgua:
I'd suggest covering with a UT. file status is one of the most complex pieces of code in AG afaik

mbtools:
It's the same UT I had added in https://github.com/larshp/abapGit/commit/ce8c4fa9e489919685ea7f37b4771cc071a16353. Makes it run a bit faster if large number of files change package assignment.

mbtools:
Taking it back. If an object is modified and assigned to a new package at the same time, then the existing logic must remain (without the clear statement).

Issue #3551, opened 2020-06-28
Repo overview - Action buttons


FreHu:
After #3498. In repo overview, add columns (or one column with a dropdown) for:

*   syntax check
*   code inspector
*   patch
*   stage

sbcgua:
Plus minor issue: review hotkeys - there are some hotkeys not in their place in overview and repo_view after split. E.g. uninstall repo in overview

PR #3645, opened 2020-07-17
Add action column to repo overview

FreHu:
#3551 #3266

Added commands to the main page, which are faster as they allow you to skip going to repo first.

todo:

*   change remote?
*   remove?

### DIFF
--- a/abaplint.json
+++ b/abaplint.json
@@ -1,5 +1,4 @@
 {
-  
   "global": {
     "files": "/src/**/*.*",
     "skipGeneratedGatewayClasses": true,
--- a/abaplint.json
+++ b/abaplint.json
@@ -1,4 +1,5 @@
 {
+  
   "global": {
     "files": "/src/**/*.*",
     "skipGeneratedGatewayClasses": true,
--- a/src/ui/zabapgit_css_common.w3mi.data.css
+++ b/src/ui/zabapgit_css_common.w3mi.data.css
@@ -915,6 +915,8 @@ table.commit tr .title {
 .repo-overview { font-size: smaller; }
 .repo-overview tbody td { height: 2em; }
 .ro-detail { display: none; }
+.ro-action { width: 190px; }
+.ro-go { font-size: 150%; }
 
 /* Branch Overview Page */
 .gitGraph-scrollWrapper, .gitGraph-Wrapper{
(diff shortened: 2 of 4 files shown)


mbtools:
I like the idea of "actions" on the overview page. "Pull" (i.e. update) would be my wish. Ideally with some job to check if there's an update available (not a diff scan but a version lookup - separate topic). I do a lot of testing, so "uninstall" would be nice, but probably not relevant for many others. Maybe we end up with some settings like for hotkeys where users can pick the actions they want to see on the overview.

On the overview, I think refresh (check for diffs), syntax check, and CI (and possibly others) should be "bulk actions": select multiple repos via checkbox, then trigger the action. Results should then be shown as icons (for each selected repo): are there any diffs? are there syntax errors? did CI run successfully?

FreHu:
Added pull, atlhough it is currently a bit mysterious without knowing the repo state. The rest are great suggestions, but there's more work involved. I'll create issues for them.

*   Background operations
*   Checkboxes + syntax check/inspector as actions on selected items
*   Diffs/status column

sbcgua:
after-thoughts: Pull vs Stage ... Stage also "blind" at this screen ... but Stage is cancellable - so it is ok

sbcgua:
aaah, hover actually brings the bg dows to grey ...
image

mbtools:
Or make it two lines:

image
(from WordPress Plugin page)

mwwhitted:
I like the WordPress plugin idea!

Thanks!
Mark

mbtools:
Not being able to cancel "pull" is another issue maybe.

There will be many links on the screen. maybe show them for the current line "on hover" only

FreHu:
### Action button placement

All the actions should be in the table header like the repo page, but operate on the selected repo. Selection would be either through line highlight, radiobutton or checkbox (later).

But to navigate to the repo, there should be an arrow, like this:

image

### Pull

I'll remove pull for now. 

In the future I think pull should be a page which combines all the popups that appear, lets you make a selection and continue/cancel.

### Maybe don't have syntax check for now?

You can always have syntax check be part of your code inspector variant so I guess it could go. 

This would only leave `Code Inspector | Stage | Patch` (i'll put code inspector first as you should check the code first).

sbcgua:
@FreHu did you finish the changes from the latest comment ? I though you mention you agree to remove pull for now but the code in unchanged.

Other suggestions (from Action button placement): these are good ideas, need some experimenting I guess. E.g. I would personally switch to repo based on any click on the table row - this is really most typical action and imho should most optimized. But there are some cons too: will work with JS only, cannot navigate easily to pkg and repo (I never use it by the way ... so for me it's ok, but needs more opinions for use cases).

FreHu:
All done. This is the final version:

PR #3671, opened 2020-07-28
Remember repo list filter

### LINKED ISSUE #3552: Remember repo overview filter on restart
#3498 followup. The filter is only remembered for the current session.

Either try local storage instead of session storage, or use backend persistence.

FreHu:
resolves #3552 

Changes:
- store filter in local storage instead of session storage, so it is remembered on abapgit restart
- filter is saved whenever it is changed instead of page unload
- removed debug output, it doesn't need to clutter the main page

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -252,7 +252,7 @@ RepoOverViewHelper.prototype.toggleItemsDetail = function(forceDisplay){
     var icon = document.getElementById("icon-filter-detail");
     this.toggleFilterIcon(icon, this.isDetailsDisplayed);
   }
-}
+};
 
 RepoOverViewHelper.prototype.toggleFilterIcon = function (icon, isEnabled) {
   if (isEnabled) {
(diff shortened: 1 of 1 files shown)


sbcgua:
Hmm, the code looks OK, but test didn't work for me - win10, 752(dev-edition), entered filter, pressed enter (filter worked), exited AG, run AG again - no filter

mbtools:
You may have the browser configured to deny permission to persist data (Or it was a private/incognito session). I can test tomorrow

FreHu:
Retested and it works for me.

sbcgua:
@FreHu ok, will have a look later today. It is strange indeed, I double checked - the localstorage is supported in IE, it really should work. 

P.S. just in case, GUI 7.40

g-back:
Works fine for me as well, GUI 7.60

sbcgua:
Ok ... did some debugging ...

1) saving "detail" and "Fav" flags do **work**
2) But based on PR description I was expecting that filter will also be saved - it does not work. But ... if was not intended - it is OK.
3) One thing that would suggest improving is double save onLoad. onLoad calls ``toggle...` which calls `saveFilter`. Not critical ... but could be improved.

debug output on start, SF = save filter, OPL = on page load. Don't mind the details, just note 2 "SF" at the end
```
storage OPL: true RepoOverViewHelperState
{"isDetailsDisplayed":true,"isOnlyFavoritesDisplayed":true}
storage SF: true RepoOverViewHelperState
storage SF: true RepoOverViewHelperState
```

FreHu:
Sorry for the confusion, I only meant to save the icon state. I fixed the page load, save is now only done on toggle.

PR #3689, opened 2020-07-30
Add string_map parameter to on_event handler

mbtools:
This is the foundation for refactoring/simplifying parameter handling (see https://github.com/larshp/abapGit/issues/3602). 

Pages can now get all field values with a single call

`io_parameter->to_abap( ls_structure_with_fields ).`

Old getdata and postdata remain for stay compatible with existing event handlers.

### DIFF
--- a/src/ui/core/zcl_abapgit_gui.clas.abap
+++ b/src/ui/core/zcl_abapgit_gui.clas.abap
@@ -1,8 +1,12 @@
 CLASS zcl_abapgit_gui DEFINITION
   PUBLIC
-  FINAL .
+  FINAL
+  CREATE PUBLIC .
 
   PUBLIC SECTION.
+
+    INTERFACES zif_abapgit_gui_services .
+
     CONSTANTS:
       BEGIN OF c_event_state,
         not_handled         TYPE i VALUE 0,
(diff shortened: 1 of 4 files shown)


mbtools:
no idea what lint is doing here. trash the branch and start over

sbcgua:
I'd suggest to sketch the final goal first. This is quite a core feature and would be nice to understand the final destination. E.g. I'd not write this code directly in GUI, I should be a separate class - with event params "in" and convenient selectors inside.
Also, how will the `on_event` shape change at the end ? Is it just importing stuff to change or also review exporting ? Maybe this is one essense (or maybe not). 
Also I'd suggest not to mix changes and re-formattings. Complicates the review a lot :)

mbtools:
Re-formatting? Well, tell that a) to the developers of pretty print and b) to the one who made our class serializer (which I will get back to soon). ;-)

You have the goal already "sketched" in ZCL_ABAPGIT_GUI_PAGE_ADDONLINE, ON_EVENT. But instead of building the string_map in PARSE_FORM in the page class, we move it up to GUI so every page will get the parameters as a string_map. 

I did this in my own tool (using a copy of your GUI framework) and it works beautifully. I can get any parameter I want with a simple string_map get (just like in your addonline form). 

Anyway, I was hurrying and will put this topic on hold for now. Not sure when I will find the time to do it properly.

sbcgua:
Reformatting: but it is not automatic, one should press a button to trigger it ;) Anyway, this was just a comment, I also do it sometimes (no pretty print though), but mixing code styling and changes makes it very difficult to review.

Regarding the design: I'll write to #3602

Issue #3688, opened 2020-07-30
Switching between forks


sbcgua:
Expect soon ...

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/15635498/88974954-179ee880-d2c2-11ea-8317-13ee225c99eb.gif)

P.S. ref #3451
P.P.S ... fed up with switching remotes ...

sbcgua:
done, closing

PR #3695, opened 2020-08-01
Switching to PRs

sbcgua:
Ready for testing and open for judgments, see the gif in #3688

## The flow
- in a repo view: press F4 (or advanced menu item)
- it will collect PRs to the current repo, **and**, if repo is a fork, add PRs from the source repo -> and display it as a command palette (pr num + name + author)
- choose the PR, the remote and branch will switch, the current one will be remembered
- at any time - switch back (remote+branch) via Advanced menu

## Usability
... is arguable, welcome to comment, haven't got enough energy to make it to the point ...
I don't like the menu, i think it is maybe time to reorganize the work with branches. Probably move them to command palette, probably merge branches + tags + prs together, unify acceess via "branch name" link ... opinions needed (but certainly for a separate PR)

Minor ideas:
- maybe PRs will be superfluous for part of users, maybe add some kind of "contribution mode" setting that will enable/disable features like this (and maybe some other). Which will in addition speedup the UI a bit !!! because now, PRs are requested by 2-3 additional http requests !!! Opinion needed
- how to indicate that origin is swicthed and remembered ? some kind of icon near "branch name" ? Maybe "reset" icon ? Maybe another color of branch name ?

## Portings
@larshp 
- I ported [ajson](https://github.com/sbcgua/ajson) - the reader part only for the moment. Maybe be useful elsewhere - there are a couple of places in the code which parses jsons with regex. May also port writer if needed ...
- I kind of ported [http_agent](https://github.com/sbcgua/abap-http-agent), though without some features and with some AG additions it is just yet another wrapper :) But ! What I want to say. Now there is `http_agent`, `http` class (finetuned for git) and `zcl_abapgit_2fa_auth_base` (finetuned for auth) which have much in common ... guess we should refactor and unify them one day ...

## Tech
*(probably will add to docs later)*
- utils
  - `zcl_abapgit_ajson` - json parser + UTs for it
  - `zcl_abapgit_http_agent` - http request wrapper, runs in one line, can return parsed json in particular. Created by `zcl_abapgit_factory=>get_http_agent`
  - interfaces for the above
  - `zcx_abapgit_ajson_error` - "compatibility" layer for ajson, inherits from abapgit_exception
- PR enumeration
  - `zcl_abapgit_pr_enumerator` - kind of factory/router - accepts repo, checks if GIT provider is supported, instantiates specific "strategy", trims exceptions to simplify handling
  - `zcl_abapgit_pr_enum_github` - "strategy" for enumaration of Guthub PRs
  - `zif_abapgit_pr_enum_provider` - interface for enum strategy to implement gitlab, bitbucket when anyone needs
- the changes to the existing code is "relatively" minor - some additions here and there - rendering, additional param in persistence, repo ... many places but more wiring stuff


sbcgua:
@larshp could you please also check linter messages, I'm not sure about the right way ...
- IF in IF - in this particular case imho this is the most proper approach (will adjust if requested)
- reduce depth - I'd propose increasing the depth in config, in this case the issue is really artificial
- `Statement does not exist in ABAPv702(or a parser error), "cl_abap_unit_assert"` - ?
- `Method "***" not found` - missing deps ?
- `Remove description for CONSTRUCTOR` - ? (for exception class)

src/http/zcl_abapgit_ajson.clas.testclasses.abap
```diff
@@ -0,0 +1,1191 @@
...
+    METHODS get_value FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS exists FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS value_integer FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS value_number FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS value_boolean FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS value_string FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS members FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS slice FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS array_to_string_table FOR TESTING RAISING zcx_abapgit_ajson_error.
+    METHODS get_date FOR TESTING RAISING zcx_ajson_error.
```
christianguenter2, line 289: Seems that `zcx_ajson_error` is missing?
image

christianguenter2: Should be `ZCX_ABAPGIT_AJSON_ERROR`?

sbcgua: yeap, good catch, thanks. Fixed

src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -255,6 +270,46 @@ CLASS ZCL_ABAPGIT_REPO_ONLINE IMPLEMENTATION.
...
+        set( iv_switched_origin = '' ).
+      ENDIF.
+    ELSE.
+      IF ms_data-switched_origin IS INITIAL.
+        set( iv_switched_origin = ms_data-url && '@' && ms_data-branch_name ).
+        set_url( iv_url ).
+      ELSE.
+        zcx_abapgit_exception=>raise( 'Cannot switch origin twice' ).
+      ENDIF.
+    ENDIF.
```
larshp, line 308:
```suggestion
    ELSEIF ms_data-switched_origin IS INITIAL.
      set( iv_switched_origin = ms_data-url && '@' && ms_data-branch_name ).
      set_url( iv_url ).
    ELSE.
      zcx_abapgit_exception=>raise( 'Cannot switch origin twice' ).
    ENDIF.
```

src/http/zcl_abapgit_ajson.clas.testclasses.abap
larshp: https://github.com/abaplint/abaplint/issues/1117 opened to catch this kind of error in the future

FreHu:
When opening a repo, I receive a rate limiting error which causes the pull url to be empty and abapgit dumps. Some http error handling needs to be added here.

sbcgua:
Will have a look. 
P.s.  the mentioned contribution mode looks a more and more better idea 
P.p.s one of further challenges would be password/token storage ... both for convenience and rate limiter issues. Any ideas where to store user passwords securely in sap?

FreHu:
For the contribution mode: It would be easiest to use if instead of the current branch popup, abapgit opens the command palette where you can choose what to do.

*   switch branch
*   create branch
*   delete branch

Then when switch branch is selected have the branch list + command to fetch pr branches.

sbcgua:
So requesting prs on demand? Definitely makes sense. 
Btw what do you think about support for non-js browsers (javagui)? Command palette will not work there probably :( could be a problem for some users

FreHu:
Unfortunately i think because of javagui the old popups will have to be kept and be switchable to the modern ui which is a big step forward for usability.

sbcgua:
> javagui and old popups

Yeah, probably it should be some global switch in settings to use this or that.

larshp:
linter error reference: https://github.com/abaplint/abaplint/issues/1036

larshp:
I dont think it is needed to call the Github API, the PRs are stored as special refs

sbcgua:
Ah yes. But there is one catch. It holds the pr branches in the original repo. And my solution analyses forks also. 

Just in case this is still wip - i have to work on Frederic's comments. A bit out of time these days. Maybe tomorrow.

larshp:
@sbcgua check conflicts

FreHu:
Small idea: I'd place this under `branch` rather than `advanced` so it's easier to find.

sbcgua:
@larshp done

@FreHu 

> place this under branch

certainly makes sense, done

> if you managed to get around the rate limiting

No, honestly I forgot about this. Here are some observations (but I'd say it should be solved in a separate PR - seems to need more studying and refactoring)
- the response from api has rate limit params. For auther user (`curl -u username:token https://api.github.com/user`) it is relatively high
2020-08-24_22h59_39
- for unauthed - relatively low
2020-08-24_23h26_53
- I added some code (c9c3662) which supposed to be a solution (but it is not)
  - `ZCL_ABAPGIT_PR_ENUM_GITHUB->CONSTRUCTOR` - access `zcl_abapgit_login_manager` for auth token
  - `ZCL_ABAPGIT_PR_ENUM_GITHUB->FETCH_REPO_BY_URL` reads response headers for debug
- but actually it is not that simple
  - A) it assumes that the user was authenticated before (commit happened or similar action that would ask auth from user)
  - B) credential after A are kept for `github.com`, not `api.github.com` - so need some refactoring
  - C) Using the force (the debug) one can workaround the B :) But it actually does not rise the limit for some reason ... so need to train the force more and rtfm the [sacred scripts](https://developer.github.com/v3/auth) to solve this :) (I'd propose to do it separately, I have no idea why it didn't work for the moment, git responds with `Basic` token, and api seems to require `token` token, or maybe they are not interchangeable ...)

larshp:
Suggest moving the json classes to `/src/json/` instead of `/src/http/`

sbcgua:
done

larshp:
thanks 👍 

if we get more github specific classes(not counting 2FA, as it will be removed), we should consider moving it to a separate package

PR #3698, opened 2020-08-02
Fix status calculation

### LINKED ISSUE #2264: Reset local: Subpackage is created with wrong text after switching branches
I have this [Demo repo](https://gitlab.com/christianguenter/test_trans.git) (abapGit plugins required) with this package structure in two branches.

- master
ZTEST_TRANS

- feature_1
ZTEST_TRANS
  |
  |----- ZTEST_TRANS_ODATA

When I switch from master to feature_1 and do reset local (or pull) the sub package is created, but with the wrong text. abapGit recognizes that and I need to reset local again to get to the correct state.

Repo state branch master
image

Repo state branch feature_1 before reset local.
image

Repo state branch feature_1 after reset local.
image

image

Correct state branch feature_1 after another reset local
image

The CTEXT of the subpackage is set here in the PATH_TO_PACKAGE logic.
image

I would be nice if abapGit can recognize this diff and deserializes the package in the first reset local.

mbtools:
Return checksum only if local item exists

During investigation of "Reset package name" https://github.com/larshp/abapGit/issues/2264, I found that the status calculation doesn't work correctly if 
- object existed before, was deleted, and then should be recreated (via pull or reset local for example)
- object existed in another branch but not in the current branch
- object had been deleted in another session

Root cause was that the local checksum still existed which made the status calculation incorrectly 'believe' the object was still there (due to the same checksum as remote). 

I created a new test case https://github.com/abapGit-tests/DEVC_main_sub with two branches. Master contains MAIN and 1 sub-package. Branch "feature" contains 2 sub-packages. 

This PR not only fixes #2264 but potentially several other open issues related to checksums and switching branches:

Branch created with no changes identified
https://github.com/larshp/abapGit/issues/3258

REFRESH_LOCAL_CHECKSUMS regression?
https://github.com/larshp/abapGit/issues/2086

Branching can cause changes to be flipped
https://github.com/larshp/abapGit/issues/2869

@christianguenter2, @7qiang, @sbcgua, @flaiker, @FreHu  please test.

### DIFF
--- a/src/zcl_abapgit_repo.clas.abap
+++ b/src/zcl_abapgit_repo.clas.abap
@@ -482,7 +482,12 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
     FIELD-SYMBOLS <ls_object> LIKE LINE OF ms_data-local_checksums.
 
     LOOP AT ms_data-local_checksums ASSIGNING <ls_object>.
-      APPEND LINES OF <ls_object>-files TO rt_checksums.
+      " Check if item exists
+      READ TABLE mt_local TRANSPORTING NO FIELDS
+        WITH KEY item = <ls_object>-item.
+      IF sy-subrc = 0.
+        APPEND LINES OF <ls_object>-files TO rt_checksums.
+      ENDIF.
     ENDLOOP.
 
   ENDMETHOD.


FreHu:
Unforunately does not resolve #2869.

PR #3710, opened 2020-08-04
Fix: Open repo in master lang opens repo

### LINKED ISSUE #3702: "Open in master language" should navigate to repo
Advanced > Open in master language navigates to the repo list. Imho it would make more sense to navigate to the repository. I think this changed while introducing the repo list as start page.

image

christianguenter2:
fixes #3702 

After this fix is applied the menu "advanced > open in master language" opens the repo directly and doesn't navigate to the repo list anymore.

The show_default_repo setting is already properly handled here, so that shouldn't be a problem.
https://github.com/larshp/abapGit/blob/59e36c47bde5233b3342dad057f24838b53dfa8a/src/ui/zcl_abapgit_services_abapgit.clas.abap#L289-L292


mbtools:
LGTM

PR #3724, opened 2020-08-10
Fix dump in Code Insp. Display for exception class

christianguenter2:
Fixes this dump for exception classes in code inspector display
image

image

image


mbtools:
There's always a reason to check return codes... Thanks @christianguenter2 :-)

PR #3731, opened 2020-08-12
show last commit hash with link if possible

bigld:
link to commit is supported for github.com and bitbucket.org for now


src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -318,4 +322,29 @@ CLASS ZCL_ABAPGIT_REPO_ONLINE IMPLEMENTATION.
...
+
+    rv_url = me->get_url( ).
+
+    FIND REGEX '^https:\/\/(?:www\.)?(github\.com|bitbucket\.org)\/' IN rv_url RESULTS ls_result.
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( |provider not yet supported| ).
+    ENDIF.
+    READ TABLE ls_result-submatches INDEX 1 ASSIGNING <ls_provider_match>.
+    CASE rv_url+<ls_provider_match>-offset(<ls_provider_match>-length).
+      WHEN 'github.com'.
```
FreHu, line 340: Can we do github without the `.com` to make it work for enterprise instances?

bigld: whats the DN pattern and the show-commit path?

sbcgua:
From my side, the code looks OK. Tested, works.
image

I'm a bit concerned the the commit extraction code is in repo class, but it requires refactoring in any case, so why not for now ...

P.S. hmmm, if I post an approve review will it override the one from Frederik or not ? Well, let's keep it a comment. :)

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
```diff
@@ -778,4 +788,31 @@ CLASS ZCL_ABAPGIT_GUI_CHUNK_LIB IMPLEMENTATION.
...
+
+        lv_icon_commit = zcl_abapgit_html=>icon( iv_name  = 'code-commit'
+                                                 iv_class = 'pad-sides'
+                                                 iv_hint  = 'Commit' ).
+        iv_html->add_a( iv_txt   = |{ lv_icon_commit }{ lv_commit_short_hash }|
+                        iv_act   = |{ zif_abapgit_definitions=>c_action-url }?|
+                                && lv_display_url
+                        iv_class = |url| ).
+      CATCH zcx_abapgit_exception.
+        iv_html->add( |<span class="url">{ lv_commit_short_hash }</span>|  ).
```
sbcgua, line 813: Ah, no. One minor bug. If the "provider not supported" - e.g. gitlab - then icon is not rendered
image

BTW: gitlab commit urls are: `https://gitlab.com/<user>/<repo>/-/commit/<hash>` 😉 🙏

src/zcl_abapgit_repo_online.clas.abap
bigld: can we use #3733 for this?

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
bigld: only online repo can have a "show url" and should know its provider and show url...

gitlab added - further provider => #3733

src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -318,4 +322,32 @@ CLASS ZCL_ABAPGIT_REPO_ONLINE IMPLEMENTATION.
...
+  METHOD get_commit_display_url.
+
+    DATA ls_result TYPE match_result.
+    FIELD-SYMBOLS <ls_provider_match> TYPE submatch_result.
+
+    rv_url = me->get_url( ).
+
+    FIND REGEX '^https:\/\/(?:www\.)?(github\.com|bitbucket\.org|gitlab\.com)\/' IN rv_url RESULTS ls_result.
+    IF sy-subrc <> 0.
+      zcx_abapgit_exception=>raise( |provider not yet supported| ).
```
larshp, line 336: this is pretty harsh? there is a large list of providers which we'll never get into the code, self hosted gitlab might not be named anything with gitlab, gitea etc.

Suggest not raising an exception, and instead silently not show the message if its not supported

larshp: okay, it catches the exception

src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -318,4 +322,32 @@ CLASS ZCL_ABAPGIT_REPO_ONLINE IMPLEMENTATION.
...
+
+
+  METHOD get_commit_display_url.
+
+    DATA ls_result TYPE match_result.
+    FIELD-SYMBOLS <ls_provider_match> TYPE submatch_result.
+
+    rv_url = me->get_url( ).
+
+    FIND REGEX '^https:\/\/(?:www\.)?(github\.com|bitbucket\.org|gitlab\.com)\/' IN rv_url RESULTS ls_result.
```
g-back, line 334: Repos can also be simple simple `http`, perhaps allow that as well?

larshp:
thanks, merging

PR #3759, opened 2020-08-18
Fix error when using "Advanced > Force Stage"

mbtools:
Show proper error message in case there are no diffs that could be staged.

image

Closes https://github.com/larshp/abapGit/issues/3650


sbcgua (approved):
Works. Nice fix

PR #3764, opened 2020-08-18
Add top repo menu to more pages

FreHu:
This menu 
image

is now in 
- diff, patch
- syntax check, code inspector
- repo settings


sbcgua (changes requested):
Nice one !
But I would add a visual separation in ci and setting - what do you think ?
Diff looks fine imho, it is visually separated because of following start of diff table

image

But in settings and ci, it kind of blurs with the following content

image

image

Options: either a single line 2-3 px wide or a full length margin would do imho

"div.settings_container: margin-top: 3px"
image

FreHu:
Fixed

PR #3765, opened 2020-08-18
Use title case for page titles

FreHu:
Other minor changes:

- patch page is no longer called diff
- "code inspector" in main page action menu renamed to "check"

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_repo_sett.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_repo_sett.clas.abap
@@ -70,7 +70,7 @@ ENDCLASS.
 
 
 
-CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_SETT IMPLEMENTATION.
+CLASS zcl_abapgit_gui_page_repo_sett IMPLEMENTATION.
 
 
   METHOD constructor.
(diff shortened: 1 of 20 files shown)


src/ui/zcl_abapgit_gui_page_repo_sett.clas.abap
```diff
@@ -70,12 +70,12 @@ ENDCLASS.
...
 
 
-CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_SETT IMPLEMENTATION.
+CLASS zcl_abapgit_gui_page_repo_sett IMPLEMENTATION.
 
 
   METHOD constructor.
     super->constructor( ).
-    ms_control-page_title = 'REPO SETTINGS'.
+    ms_control-page_title = 'Repository settings'.
```
mbtools, line 78: Repository **S**ettings

mbtools:
Thank you. Menu items next ;-)

FreHu:
Will do menu items as well.

mbtools:
Excellent. This gives the UI a little boost. Just a slight disconnect between "Database Util" (menu) and "Database Persistency" (page title)

sbcgua:
Hmmm ... some of renames look controversial. Why not just leave all the following works lowercased ?
image


Though in Menu it looks good, I agree
image

In settings, also controversial
image

Also ... all future changes (by other people) would normally capitalize just the first word, so it will eventually become more chaotic. I'd rather keep only first word capitalized

FreHu:
I was a bit indecisive about the longer labels, but then I looked at vscode settings: 



Prepositions are normally not capitalized, although there are several styles which might be opinionated about this.

I'll unify the database util terminology and add tooltips to stage.

sbcgua:
Good references (both Fiory and VSCode), though this does not explain why it is better ... well, anyway it is more about common style

PR #3767, opened 2020-08-18
Hide stage and patch action links for offline repo

FreHu:
Offline repositories don't have these actions, stage dumps and patch shows an error.


sbcgua (approved):
Kind of works :) :+1:

image

PR #3849, opened 2020-09-06
Always show restart message for a second and then restart without user interaction

### LINKED ISSUE #3838: Autostart without popup
It'd be cool if there would be an option in global settings to be able to suppress the autostart popup after update.

mariusraht2:
Closes #3838 (for real this time)

### DIFF
--- a/src/zcl_abapgit_repo.clas.abap
+++ b/src/zcl_abapgit_repo.clas.abap
@@ -242,6 +242,8 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
     CONSTANTS:
       lc_abapgit_prog TYPE progname VALUE `ZABAPGIT`.
 
+    DATA: li_progress TYPE REF TO zif_abapgit_progress.
+
     " If abapGit was used to update itself, then restart to avoid LOAD_PROGRAM_&_MISMATCH dumps
     " because abapGit code was changed at runtime
     IF zcl_abapgit_ui_factory=>get_gui_functions( )->gui_is_available( ) = abap_true AND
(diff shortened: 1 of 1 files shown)


flaiker:
Hm, I liked the popup. Now when you update and tab out and back in again you might be back at the repo overview wondering if the update was finished, right?

mariusraht2:
I don't think so: The update is done when you're on the repo view and the progress bar isn't being shown anymore..^^ That's what the progress bar is for. I didn't like the popup for the exact same reason: One had to specifically confirm a message which blocks a process you're just waiting for to finish so you can do what you wanted to do in the first place.

flaiker:
I agree an option would be overkill. The return to last repo on load option seems to already complicate things a bit :)

Is there a reason you removed the binding to that one for this feature? If you return to the abapGit repo after the non confirmed automatic restart then it is the same behavior as just pressing enter in the popup, as you say. But if you restart into the repo overview it's different.

mariusraht2:
Then the best solution would be to add an option :( If I for instance want to update AG and after that work with my own repos (that's the usual case for me) I wouldn't set the specific option to return to the last repo (which would have been AG) which would suppress the popup-free autostart.

mariusraht2:
Why specifically does the popup makes sense to you if restart into the repo overview page?

flaiker:
abapGit updates might take up to a coffee break. When I return and I just have the repo overview open I won't be reminded that I started the update 10 minutes ago :)

mariusraht2:
and we don't know if AG has been started after an update. If that would be the case we could show a message after the restart.. :(

flaiker:
One could export / import a flag to memory but makes it more and more convoluted.

I always have return to last repo off by the way. I don't want to start abapGit to then wait a minute for the repo to serialize to then switch to the repo I actually wanted to view.

PR #3855, opened 2020-09-08
DOMA: Clear VALPOS to show the right diff

### LINKED ISSUE #1053: DOMA: consider clearing VALPOS

mariusraht2:
Closes #1053

Related to PR #3094


larshp (changes requested):
For the texts, if not all values are translated, how does it find the value corresponding to the translated text?

image

mariusraht2:
If there's a translation for a value it finds it by DOMVALUE_L or its absence and sy-langu. The example you chose is a fixed empty value - in your example the attribute DOMVALUE_L has been suppressed because of that. But fixed values can't occur twice so it's not possible to have multiple translated entries with an empty value (keyfield). 

image
-> not activatable

XML:
image

mariusraht2:
If there's a translation for a value it finds it by DOMVALUE_L or its absence and sy-langu. The example you chose is a fixed empty value - in your example the attribute DOMVALUE_L has been suppressed because of that. But fixed values can't occur twice so it's not possible to have multiple translated entries with an empty value (keyfield). 

image
-> not activatable

XML:
image

mariusraht2:
If there's a translation for a value it finds it by DOMVALUE_L or its absence and sy-langu. The example you chose is a fixed empty value - in your example the attribute DOMVALUE_L has been suppressed because of that. But fixed values can't occur twice so it's not possible to have multiple translated entries with an empty value (keyfield). 

image
-> not activatable

XML:
image

larshp:
ie. the solution does not work

mariusraht2:
@larshp Just want to make sure: But you saw the added DELETE of empty untranslated right? It didn't still work then? After I added it it worked for me

PR #3874, opened 2020-09-11
PROG: Add program class to deserialize

mbtools:
Closes https://github.com/larshp/abapGit/issues/3870


jfilak:
How comes the remote file has the element CLAS while the local file does not?

filak-sap:
@mbtools Thank you!

BTW, if I were you, I would no expect reviewer clicks on the issue or future poor git log archeologist have access to the issue. Something like "we do serialize CLAS, so we also have to deserialize" would do the job.

mbtools:
I get your point but It’s a one line fix. Typing anything here is a waste of time and money imho.

jfilak:
Yeah, one line fix and I spent 15 minutes thinking "how comes the field is in remote file and not in local file". Taking it from this point, doing reviews is a waste of my time and someone's money because when applying this logic I get to the conclusion that only people who knows abapGit source code very well should do reviews.

PR #3884, opened 2020-09-14
Refactoring XML: zif_abapgit_object_enho

g-back:
ref #3531 

changed references in `object_enho` interface, also renamed variables


larshp (approved):
LGTM

PR #3910, opened 2020-09-20
Show repo list button in tutorial

flaiker:
Currently the button is only visible in the tutorial text while the others (new offline, new online) are also included in the main menu.


sbcgua (approved):
This works too.
image

sbcgua:
P.S. ... the #3594 is coming ... 
only in edge cases though ...

PR #3955, opened 2020-09-28
Fix syntax errors in 702 

mbtools:
Chaining method calls with CHANGING parameters is not supported in 702


sbcgua (approved):
Thanks for the fix!

PR #3960, opened 2020-09-29
Dialogs: add offline repo HTML form

mbtools:
image


sbcgua (changes requested):
I suggest adding a check that repo name is not empty. In the form. And somewhere deep in repo classes too (`ZCL_ABAPGIT_REPO_SRV->ZIF_ABAPGIT_REPO_SRV~NEW_OFFLINE` ???). Otherwise the below is possible :) The rest looks fine imho.

2020-09-29_23h31_48

image

mbtools:
I was expecting that the form validation takes care of this automatically. Turns out the "required" attribute wasn't set for the input field. Now it checks properly without round-trip for empty fields that are defined as required. So we don't need to make this check on the backend anymore.

src/ui/zcl_abapgit_html_form.clas.abap
```diff
@@ -484,6 +489,10 @@ CLASS ZCL_ABAPGIT_HTML_FORM IMPLEMENTATION.
         zcx_abapgit_exception=>raise( |Unexpected form field [{ <ls_entry>-k }]| ).
       ENDIF.
 
+      IF ls_field-required = abap_true AND <ls_entry>-v IS INITIAL.
+        zcx_abapgit_exception=>raise( |Unexpected empty form field [{ <ls_entry>-k }]| ).
+      ENDIF.
```
sbcgua, line 494: This is OK for the default mechanism, but I'd suggest make it a bit less technical. Just `|{ here-is-full-field-name } field cannot be empty|`

sbcgua (changes requested):
A bit overdone imho. See the comments. But the idea is good :)

src/ui/zcl_abapgit_html_form.clas.abap
```diff
@@ -359,12 +360,16 @@ CLASS ZCL_ABAPGIT_HTML_FORM IMPLEMENTATION.
...
+          lv_required = ' required'.
+        ENDIF.
+
         IF is_field-side_action IS NOT INITIAL.
           ii_html->add( '<div class="input-container">' ). " Ugly :(
         ENDIF.
 
         ii_html->add( |<input type="text" name="{ is_field-name }" id="{
-          is_field-name }"{ is_field-placeholder } value="{ lv_value }"{ is_field-dblclick }>| ).
+          is_field-name }"{ is_field-placeholder } value="{ lv_value }"{ is_field-dblclick }{ lv_required }>| ).
```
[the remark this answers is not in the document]
mbtools: Agree. Your way was nicer

src/ui/zcl_abapgit_html_form.clas.abap
mbtools: Exceptions show up in a banner. I removed this one.

mbtools:
1 step back, 2 steps forward. I added the check of the required fields to the form class. Works well now.

Settings pages are next

mbtools:
I'm not happy with "Enter" and "Escape" behaviour. Should be "Default button" and "Back" but it's not

sbcgua (approved):
Yeap, can go forward now :) Nice features ! :)

P.S. I have a strange feeling about 2 validate methods in form. Maybe they can be unified. No concrete suggestions yet, let's keep for now
P.P.S I have one kind request @mbtools : would be really great if you don't mix style changes and feature changes. This makes the review a lot more difficult. I know it's tempting ... I barely can withstand myself ... and sometimes I don't ... :( But would greatly simplify ... Also some of stylings are opinionated which also adds some confusion.

sbcgua:
> I'm not happy with "Enter" and "Escape" behaviour. Should be "Default button" and "Back" but it's not

"Escape" exits the whole transaction, very unfortunately. "Enter" - yeah, potentioan room for improvement. And I guess both should work better in web version in future :)

PR #3961, opened 2020-09-29
Add exit for integration of unit test with CI

mbtools:
Goal:

Run CI tests within the unit test of an object type class. The setup of a test class should be minimal. As a default, the unit test should not run any CI tests. However, it shall be possible to define locally a list of CI tests that should be executed.

Solution:

This PR provides a new exit that can be called from a unit test. In the exit, one or more abapGit test repos (https://github.com/abapGit-tests) are defined that will then be run via abapGit CI (https://github.com/abapGit/CI).

Requirements:

- Installation of https://github.com/abapGit/CI
- https://github.com/abapGit/CI/pull/80 (not merged yet)

Instructions:

Add the following to the test class of your object type - for example, `zcl_abapgit_object_tabl`. Side note: This should be mandatory for all object type classes since we require a test repo anyway. 

```abap
CLASS ltcl_ci DEFINITION FOR TESTING RISK LEVEL HARMLESS DURATION SHORT.

  PUBLIC SECTION.
    METHODS run FOR TESTING.

ENDCLASS.

CLASS ltcl_ci IMPLEMENTATION.
  METHOD run.
    zcl_abapgit_exit=>get_instance( )->run_ci_tests( 'TABL' ).
  ENDMETHOD.
ENDCLASS.
```

Implement the `run_ci_tests` method to call CI test run(s).  

Example implementation:

```abap
  METHOD zif_abapgit_exit~run_ci_tests.

    DATA:
      ls_result      TYPE zif_abapgit_ci_definitions=>ty_result,
      ls_repo_result TYPE zabapgit_ci_result,
      lr_repos       TYPE zcl_abapgit_ci_test_repos=>gty_repo_name_range,
      ls_repos       LIKE LINE OF lr_repos,
      lv_url         TYPE string,
      lx_error       TYPE REF TO zcx_abapgit_exception.

    ls_repos-sign = 'I'.
    ls_repos-option = 'EQ'.

    " >>> Setup: For each object type, list the repos that you want to test
    CASE iv_object.
      WHEN 'TABL'. " Name of object type
        ls_repos-low = 'TABL'.  " Name of repo at https://github.com/abapGit-tests
        APPEND ls_repos TO lr_repos.
        ls_repos-low = 'TABL_foreign_key'.
        APPEND ls_repos TO lr_repos.
    ENDCASE.
    " <<< End of Setup

    IF lines( lr_repos ) = 0.
      RETURN.
    ENDIF.

    TRY.
        ls_result = NEW zcl_abapgit_ci_controller(
          ii_repo_provider = NEW zcl_abapgit_ci_test_repos( CORRESPONDING #( lr_repos[] ) )
          is_options       = VALUE #(
            result_git_repo_url    = ''
            post_errors_to_slack   = abap_false
            slack_oauth_token      = ''
            exec_generic_checks    = abap_false
            exec_repository_checks = abap_true
            repo_check_options = VALUE #(
              check_local         = abap_true
              check_transportable = abap_false
              layer               = ''
            )
          )
        )->run( ).

        " No errors?
          exp = abap_false
          act = ls_result-ci_has_errors ).

        " Check individual repos
        LOOP AT ls_result-repo_result_list INTO ls_repo_result.
            exp = 'OK'
            act = ls_repo_result-create_package ).
            exp = 'OK'
            act = ls_repo_result-clone ).
            exp = 'OK'
            act = ls_repo_result-pull ).
            exp = 'OK'
            act = ls_repo_result-syntax_check ).
            exp = 'OK'
            act = ls_repo_result-object_check ).
            exp = 'OK'
            act = ls_repo_result-purge ).
            exp = 'OK'
            act = ls_repo_result-check_leftovers ).
        ENDLOOP.

        cl_abap_unit_assert=>fail( ).
    ENDTRY.

  ENDMETHOD.
```


src/objects/zcl_abapgit_object_tabl.clas.testclasses.abap
```diff
@@ -0,0 +1,12 @@
+CLASS ltcl_ci DEFINITION FOR TESTING RISK LEVEL HARMLESS DURATION SHORT.
```
flaiker, line 1: I don't think these tests are `HARMLESS` or `SHORT`.

mbtools:
Obviously depends on what you put in the exit. It runs nothing by default so I put harmless short. 

Since the intention is for more risky test, we certainly change it. It’s quite fast though.

pokrakam:
Curiosity question: Is there a reason for doing this in a user exit? Personally I'd implement this in a global test class.

larshp:
and then enable/disable via 

image

mbtools:
There are a few issues that need to be addressed. Correct me if I'm wrong:
1. We don't want a dependency on the CI repo
1. We don't want to impact a system in case someone runs all AG unit tests
1. As a developer/contributor, we want to be able to choose which CI tests to run

A global test class could solve 1. by calling the CI test code dynamically (which is still a dependency) and 2. by using the "dangerous" switch (but just partially since there's only one switch for all unit tests). But how would 3. work without an exit?

How about we use the exit to just ask for the range of repos and put the rest into a global class?

flaiker:
Won't 3 just work automatically? Since the test is registered in the test include of the serializer class the scope of the unit test classes to execute is determined by the tooling (right click on serializer class = just that one, right click on package = all). Or do you mean to specify the repos for the serializers with multiple test cases?

mbtools:
The later. For example, TABL has 10 different test repos. So I think it would be nice to specify which repos in https://github.com/abapGit-tests should be checked. There would also be the case of using your own repo(s) but that's not covered in this PR.

pokrakam:
I don't think a dependency is avoidable, and IMO it is not an issue since abapGit-CI is specifically for abapGit. If 10 developers each write their own exits to call CI then we still have 10 dependencies, the difference being that are "someone else's problem". 

I think a global TC could either call CI via the API or it can check it's existence. A global TC does not have to test all repos either, it just acts as a proxy. 
```
CLASS ZCL_ABAPGIT_CI_PROXY DEFINITON FOR TESTING
...
  METHODS run_ci_for IMPORTING iv_repo_name.
...
```

```
CLASS ltc_ci DEFINITON FOR TESTING. 
  PRIVATE SECTION. 
    METHODS run_ci FOR TESTING. 
...
  METHOD run_ci. 
    lo_ci_proxy->run_ci_for( 'TABL' ).
...
```

Now for the thorny bit: From my brief tests, this is still way too slow to use for TDD. Runtime ~5s via CI vs 0.5s local. Some kind of repo caching would be needed.

larshp:
change the Ctrl+Shift+F10 shortcut to do `CALL TRANSACTION 'ZABAPGIT_CI_RUN_TESTS_CONFIGURED_BY_USER'` ? 😜

mbtools:
I moved the logic to a global test class `ZCL_ABAPGIT_OBJECTS_CI_TESTS` that dynamically calls the CI test controller.

However, I can't figure out how to make the following work:

'Call of the method CONSTRUCTOR of the class ZCL_ABAPGIT_CI_CONTROLLER failed; wrong type for parameter II_REPO_PROVIDER'

        CREATE OBJECT li_ci_controller TYPE ('ZCL_ABAPGIT_CI_CONTROLLER')
          EXPORTING
            ii_repo_provider = li_repo_provider "<<< how to dynamically cast to ZIF_ABAPGIT_CI_CONTROLLER ???
            is_options       = <ls_options>.

Hmmm...

Unit test:

```abap
CLASS ltcl_ci DEFINITION FOR TESTING RISK LEVEL CRITICAL DURATION SHORT.

  PUBLIC SECTION.
    METHODS run FOR TESTING.

ENDCLASS.

CLASS ltcl_ci IMPLEMENTATION.
  METHOD run.
    zcl_abapgit_objects_ci_tests=>run( 'TABL' ).
  ENDMETHOD.
ENDCLASS.
```

The user exit is simplified to:

```abap

    " Return the name of repos at https://github.com/abapGit-tests
    CASE iv_object.
      WHEN 'TABL'.
        APPEND 'TABL' TO rt_repos.
        APPEND 'TABL_foreign_key' TO rt_repos.
    ENDCASE.

  ENDMETHOD.
```

The CI tests will do a complete end-to-end check (with a local package, not a transportable package). A few seconds seems reasonable for that. What's included in your 0.2s local test?

pokrakam:
Do you still need an exit class in this scenario? I would think now a dev can just call the proxy directly and be done with it. The proxy checks the critical setting and calls CI. 

Also, I don't think the CI controller is the most remote-friendly entry point. But technically it should still be possible to call it dymamically:
```
    DATA: li_ci_controller TYPE REF TO object,
          li_repo_provider TYPE REF TO object.

    DATA repos TYPE RANGE OF string.
    APPEND VALUE #( sign = 'I' option = 'EQ' low = 'TABL' ) TO repos.
    CREATE OBJECT li_repo_provider TYPE ('ZCL_ABAPGIT_CI_TEST_REPOS')
      EXPORTING it_repo_name_range = repos.

    CREATE OBJECT li_ci_controller TYPE ('ZCL_ABAPGIT_CI_CONTROLLER')
      EXPORTING ii_repo_provider = li_repo_provider.
```

For my test I copy-pasted the repo into code. See
https://github.com/abapGit/abapGit/blob/1e93c1d3d360efbf646ac190aa1d3341700ca12e/src/objects/zcl_abapgit_object_pdts.clas.testclasses.abap#L26
and the de/serialization test at
https://github.com/abapGit/abapGit/blob/1e93c1d3d360efbf646ac190aa1d3341700ca12e/src/objects/zcl_abapgit_object_pdts.clas.testclasses.abap#L265
It's been merged into `master` so you should be able to run the test yourself, zcl_abapgit_object_pdts

pokrakam:
I think either we should call the REST API or add a nicer background call to CI.

mbtools:
I already tried what you suggested. Unfortunately, the second create statement fails. 

I had a look at pdts test class. that is a lot of hardcoded XML. It works but is imho limited to simple cases. 

Have a look at the test cases for WDCA https://github.com/abapGit-tests/WDCA/tree/master/src or SFPF https://github.com/abapGit-tests/SFPF/tree/master/src. There are 120 object type classes. We can't possibly put all those test cases into abapGit itself.

mbtools:
Regarding exit see my comment above https://github.com/abapGit/abapGit/pull/3961#issuecomment-700777796

pokrakam:
Yes the XML is a copy-paste from GitHub. It was a quick hack to get a test working that would save me time. It's not ideal and that's why I raised #3776 because I think the whole thing could be streamlined. Your approach is tackling it form the other end. In the middle I think perhaps a local cache for the test repo or something like that could do the trick. 
The main point of my mentioning the PDTS test is that it runs fast enough to make it feasible to run de/serialization in a unit test, and from my own experience on it's been a huge timesaver.

An alternative to user exit could be a repo config option to list test repos that would be run for unit testing.

pokrakam:
BTW, I tested the dynamic call, and it seems ABAP doesn't like interfaces. Bug or design? Smells like bug to me, but nevertheless I learnt something today :)
```
INTERFACE lif_test.
ENDINTERFACE.

CLASS lcl_main DEFINITION.

  PUBLIC SECTION.
    INTERFACES lif_test.
    METHODS run.
    METHODS run_with_obj IMPORTING i_obj TYPE REF TO lcl_main.
    METHODS run_with_intf IMPORTING i_intf TYPE REF TO lif_test.

ENDCLASS.

CLASS lcl_main IMPLEMENTATION.

  METHOD run.
    DATA obj TYPE REF TO object.
    DATA intf TYPE REF TO object.

    CREATE OBJECT obj TYPE ('LCL_MAIN').
    CALL METHOD obj->('RUN_WITH_OBJ') EXPORTING i_obj = me.  "OK
    CALL METHOD obj->('RUN_WITH_INTF') EXPORTING i_obj = me. "Dump

  ENDMETHOD.
```

mbtools:
ok, why bother with the controller that has big overhead. I changed to `ZCL_ABAPGIT_CI_REPOS` and now it's working and pretty fast. For TABL it takes about 1.5s. I tried with PDTS and it's 1.2s. 

The exit needs to return repo name and clone URL, which makes it even more flexible since you can refer to any of your own repos, too:

```abap
   " Return the name of repos and their clone URL
    " The name is used to define the package "$___<name>" link to the abapGit repo

    " Examples:
    CASE iv_object.
      WHEN 'TABL'.
        " Generic insert which works for any object type
        " given we require such test repo for any new object type
        INSERT VALUE #(
          name      = iv_object
          clone_url = |https://github.com/abapGit-tests/{ iv_object }| )
          INTO TABLE rt_repos.

        " Add some other repo from https://github.com/abapGit-tests
        INSERT VALUE #(
          name      = 'TABL_foreign_key'
          clone_url = 'https://github.com/abapGit-tests/TABL_foreign_key' )
          INTO TABLE rt_repos.

        " Add your own repo works as well
        INSERT VALUE #(
          name      = 'MYTEST'
          clone_url = 'https://github.com/mbtools/abapGit-test-del' )
          INTO TABLE rt_repos.

      WHEN 'PDTS'.
        INSERT VALUE #(
          name      = iv_object
          clone_url = |https://github.com/abapGit-tests/{ iv_object }.git| )
          INTO TABLE rt_repos.
    ENDCASE.
```

flaiker:
> BTW, I tested the dynamic call, and it seems ABAP doesn't like interfaces. Bug or design? Smells like bug to me, but nevertheless I learnt something today :)

I think you just had a typo :P

```diff
 CREATE OBJECT obj TYPE ('LCL_MAIN').
 CALL METHOD obj->('RUN_WITH_OBJ') EXPORTING i_obj = me.  "<--- OK
-CALL METHOD obj->('RUN_WITH_INTF') EXPORTING i_obj = me. "<--- Dump
+CALL METHOD obj->('RUN_WITH_INTF') EXPORTING i_intf = me. "<--- OK
```

pokrakam:
> I think you just had a typo :P

I did indeed, thanks! It was late, that's my excuse and I'm sticking with it! :P

mbtools:
Closing in favor of https://github.com/abapGit/abapGit/pull/3993

PR #3975, opened 2020-10-02
html event abstraction, phase 4 - form_data

sbcgua:
to #3602

final part (at least for now).

`ii_event->form_data( )` attempts to parse post_data assuming it is key-value and returns a string_map. Optionally `iv_upper_cased`. The rest is the same as `query`.

Also refactored most of pages. Except `add_online` because waiting for #3960. Will adjust both of them after. And except of both settings pages - they are painful to refactor and maybe does not make sense if they will be switched to new form which will change the whole parsing completely.

### DIFF
--- a/docs/collections/_development/developing-ui.md
+++ b/docs/collections/_development/developing-ui.md
@@ -107,6 +107,8 @@ To process sapevents in abap the component (page) must implement `ZIF_ABAPGIT_GU
   - `ii_event->query( )->get( 'XXX' )`
   - or `ii_event->query( )->to_abap( changing cs_container = ls_struc_with_fields )`
   - query string_map is immutable (attempt to `set` will raise an exception)
+  - accepts optional `iv_upper_cased` param to unify param names (default = `true`)
+- `ii_event->form_data()` - attempts to parse post_data assuming it is set of key value pairs. Returns a string map. Otherwise behaves as `query()` above. `iv_upper_cased = false` by default.
 
 Events can be processed on 2 levels - in page/component **or** in the router. On new event:
 - the GUI goes through event handlers stack - list of components that registered themselves as event handlers during rendering via `gui_services`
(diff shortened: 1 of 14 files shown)


sbcgua:
... and docs

mbtools:
cool. this is much easier to use. 

there seem to be too many options. query, form, lower, upper... confusing. 

To the point...
```abap
lv_param = ii_event->query_param( 'key' ).
ls_params = ii_event->query_params( ls_params ).
lv_param = ii_event->form_param( 'key' ).
ls_params = ii_event->form_params( ls_params ).
```

The first should also work with ii_event->...( 'KEY' ). Just to_lower the input before looking up the string map which I would just keep in lower case. 

The second method is short for this:
```
    DATA lo_map TYPE REF TO zcl_abapgit_string_map.
    lo_map = ii_event->form/query_data( ).
    lo_map->to_abap( CHANGING cs_container = ls_param ).
```

Maybe single value and structure can be combined this:
```
lv_param = ii_event->query( 'key' ).
ls_param = ii_event->query( ls_param ).
lv_param = ii_event->form( 'key' ).
ls_param = ii_event->form( ls_param ).
```
It really could not get simpler than that :-)

PS: With that we could get rid of all the parsing in `ZCL_ABAPGIT_HTML_ACTION_UTILS`.

sbcgua:
> lower, upper... confusing

Yes, agree, I had the same feeling. Problem is that some of the code does use it case sensitive and some is not. I didn't refactor all of those. But, yeah, I think it is a good **next** step. It will probably be `case_insensitive` string_map, rather than something in the event class directly. P.S. ufff, another portion of refactoring 😢 😢 

> `lv_param = ii_event->query( 'key' )` and `ls_param = ii_event->query( ls_param )`

I have to confess I like how it looks and how it reads (ignoring the fact that `ls_param = ...` will not work because of unknown type). But I have a gut feeling<sup>tm</sup> that it would be less flexible. E.g. you can take string_map and pass it as a param somewhere isolation the whole event from access. Or the string_map `strict( false )` which is used just in one place and can be eliminated to be honest, but ... again gives some flexibility. Or ! What if you want to get the full list of params, like in stage. So ... tentative. I'd keep as is.

P.S. ZCL_ABAPGIT_HTML_ACTION_UTILS will probably stay as an util class. Easier to unit test and a quick workaround for something unpredicted.

PR #3991, opened 2020-10-07
Adjust name prefix for types

mbtools:
https://github.com/abapGit/abapGit/issues/3398


mbtools:
Done! New rule will be active with this PR

flaiker:
Some of these types are public making them relevant for possible external usages. Do we need to check other repositories or inform about a change in the API?

larshp:
merging, and then when https://abaplint.app/stats updates its possible to look through those with "custom config"

larshp:
tested ok, no syntax errors

sbcgua:
a minor concern about table type prefixes. It was quite convenient to see `tt` or `tty` to identify a table type. Of course, it also depends on plural names. `tt_thing` -> `ty_things`. But ... if prefix does not add anything then why it is needed at all ? Just imho. Though the work done for unification is very good :+1:.

larshp:
it adds to static analysis via abaplint, having a convention then it knows if to issue an error,
 
image

mbtools:
@flaiker I had tested abaplint-sci and abapGit-ci repos but none of the changed types were used there. of course, we don't control how other tools might use parts of abapGit.

@sbcgua I wasn't sure what's better, plural or some suffix for the table types. I tried to stick with the kind already used by the class/interface. Also didn't want to change existing types unless there was a conflict. 

Anyway, chapter closed. :-)

Issue #3675, opened 2020-07-29
Dump when importing repo with files in root


mbtools:
Example: https://github.com/larshp/Datamatrix

Starting folder is set to /src/ but file package.devc.xml exists in /  [ it should be in /src/ ].

Root cause:

ZCL_ABAPGIT_DOT_ABAPGIT->IS_IGNORED does not handle this case which later leads to a dump since there's no package assigned to this file. 

Question:

If the starting folder is not root (/), then shouldn't **all** files outside the starting folder be ignored (.abapgit and .apack being the exception)? And if yes, then what is the point of listing files in the root folder to be ignored in .abapgit.xml? 

Suggestion:

Simplify the ignore method as follows 

```
 METHOD is_ignored.

    (data...)

    lv_name = iv_path && iv_filename.

    CONCATENATE ms_data-starting_folder '*' INTO lv_starting.

    " Always allow .abapgit.xml and .apack-manifest.xml
    CONCATENATE '/' zif_abapgit_definitions=>c_dot_abapgit INTO lv_dot.
    IF lv_name = lv_dot.
      RETURN.
    ENDIF.
    CONCATENATE '/' zif_abapgit_apack_definitions=>c_dot_apack_manifest INTO lv_dot.
    IF lv_name = lv_dot.
      RETURN.
    ENDIF.

    " Ignore all files matching pattern in ignore list
    LOOP AT ms_data-ignore INTO lv_ignore.
      IF lv_name CP lv_ignore.
        rv_ignored = abap_true.
        RETURN.
      ENDIF.
    ENDLOOP.

    " Ignore all files outside of starting folder tree
    IF NOT lv_name CP lv_starting.
      rv_ignored = abap_true.
    ENDIF.

  ENDMETHOD.
```

Result:

If you use the default /src/ starting folder, you do not have to maintain the ignore list anymore and you can add any files to the root folder. We could remove the defaults for ignored files in .abapgit.xml as well. You will only have to maintain the ignore list if you use / as a starting folder or want to explicitly ignore files contained in the starting folder tree.

PR #3996, opened 2020-10-07
Simplify Ignore Logic

mbtools:
This greatly simplifies how files to be ignored have to be maintained:
- Default **starting folder** remains **/src/**
- Any **files in root /** or any other folder than the starting folder are **ignored automatically**
- Therefore it will not be necessary to list files of the root folder into the ignore list
- And as a consequence, the **default ignore list is empty**

Additional unit test and updated docs.
Closes https://github.com/abapGit/abapGit/issues/3675


larshp (approved):
looks good to me, thanks 👍

PR #4011, opened 2020-10-10
Bugfixes in html_form and Finetune add pages to new event

sbcgua:
after #4007, ~~**don't review yet as it contains changes from both PRs**~~

So main purpose is to remove unnecessary parse_form from the add pages which was a very straight forward fix. But !
It appeared that validation and normalization of html_form are done incorrectly - it loops through incoming data instead of fields definition which lead to incorrect processing if input fields are missing. so fixed, and covered with UTs.
On the way I had  to remove the check on extra input fields that were missing in the definition. But I think it is ok, actually it is a good default validation logic.


sbcgua:
rebased and ready now

PR #4024, opened 2020-10-14
Add commit attribute to online repo to be able to save checked-out commit hash

mariusraht2:
Related to #3040

Adds an attribute to repo_online that stores the selected/checked-out commit hash. Getter and setter methods are needed in order to be able to read or set the related attribute.


mariusraht2:
s. #4025

PR #4041, opened 2020-10-17
ZCL_ABAPGIT_GIT_PORCELAIN=>FULL_TREE: Rename iv_branch to iv_parent

mariusraht2:
Related #3040

Method `full_tree` is to be used for commit hashes too. As the import parameters type is already of type ty_sha1 and the functionality can be used without a change for commit hashes there is no need to implement a commit hash-related method.

**Change Log:**
- Renamed `iv_branch` of `full_tree` to `iv_parent`
- Updated related caller


larshp (approved):
thanks, looks good to me

PR #4047, opened 2020-10-19
ABAP Syntax Highligher: Re-Add New Keywords

g-back:
- add minimal unit test
- readd new keywords 
- removed duplicates from keywords
- removed deprecated pseudo comments
- explicitly remove duplicates from keywords to avoid needless dumps

References: #4042, #4044


mbtools:
I should have kept it alphabetical. Thanks!

PR #4054, opened 2020-10-22
DEVC - Clear obsolete fields

### LINKED ISSUE #4053: DEVC - remove obsolete fields
This is supposed to be a "suggested prefix for package interfaces (obsolete)". These values seem to appear inconsistently on systems, causing diffs.

Could also remove the errsevrty, it's also marked as obsolete.

FreHu:
Resolves #4053

*   intprefx removed
*   `errsevrty` is not a problem as it's commented out already in cl\_package, but i found `cli_check` which is also marked as obsolete


mbtools (approved):
Thanks. I had the same problem.

PR #4066, opened 2020-10-26
ZCL_ABAPGIT_REPO_ONLINE: Add getter and setter for branch

mariusraht2:
Related to #3040

We have to be able to get or set the current repo hash based on checked out commits: The user can choose a commit from a popup and then in the background the repo commit (`mv_branch`) is set to this selected commit hash. AG resets the repo based on the checked out commit hash after that.

@larshp I'll plan to rename `mv_branch` to `mv_sha1`. Is this ok with you or do you prefer another name (e. g. `mv_commit`) or a different approach?


larshp:
hmm, guess it wont break anything

but looking at
image
there is logic to `mv_branch`, so don't think this is an easy fix, take care

PR #4079, opened 2020-10-27
UI: Adjustments to Repo Palette

mbtools:
- Sort alphabetically
- Pad offline icon (using CSS)

Before:

image

After:

image


sbcgua (dismissed):
Looks good, but I'd suggest another change to CSS. I didn't like the fix, because it is focused on a specific icon and so I was going to suggest another one, with fixed width ... but it didn't work ! So after some minutes of confusion I found the below code at the very end of this file.

```css
.icon-cloud-upload-alt {
    padding-right: 5px;
}
```

Removing it completely solve the alignment issue. And this is what I suggest. Even if that code was used for something it must be more specific. So I'd propose removing it.

The rest looks fine :)

mbtools:
The CSS was added to align the icons on the top:

image

Without it, "New Online" was very close to the icon:

image

Maybe it should be

```css
#toolbar-main .icon-cloud-upload-alt {
    padding-right: 5px;
}
```

Not perfect either.

sbcgua:
Yeap, good catch, I didn't notice. This will do. Not perfect but a lot more specific, so will not interfere.

P.S. Hmm are these spaces hardcoded ?
image

sbcgua:
option 2 ... probably to do in parallel is to adjust that icon - it really is larger
image

sbcgua:
image

Hmm ?

sbcgua:
May take some more time ... :(
image

sbcgua:
Or use other source. svgrepo seems to give icons under public domain
image

mbtools:
smaller icon alt2 looks good

mbtools:
I'm not sure why our cloud icon has an arrow in it? Why not just
https://fontawesome.com/icons/cloud?style=solid

PS: If you don't mind, please add an icon for "user" to the font:
https://fontawesome.com/icons/user?style=solid

sbcgua:
@mbtools have a look at #4112 - submitted some icon and css changes
Also consider cloud-commit icon. solid cloud looks too bulky and heavy on a real page

image

mbtools:
I removed the CSS here since it's addressed in #4112 now. I think this is good to go now

sbcgua (approved):
works. Thanks !

PR #4082, opened 2020-10-27
ZCL_ABAPGIT_GUI_CHUNK_LIB: Determine commit hash to show current commit hash

mariusraht2:
Related to #3040

Adjust render_repo_top_commit_hash in ZCL_ABAPGIT_GUI_CHUNK_LIB to show the right commit hash: Either the user checked a commit out (show related commit hash) or not (show branch hash).

### DIFF
--- a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
+++ b/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
@@ -584,11 +584,16 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
               iv_act   = |{ zif_abapgit_definitions=>c_action-direction }?direction=DESCENDING|
               iv_title = <ls_col>-title ).
           ENDIF.
-        ELSE.
+        ELSEIF <ls_col>-allow_order_by = abap_true.
           lv_tmp = lv_tmp && ri_html->a(
             iv_txt   = lv_disp_name
             iv_act   = |{ zif_abapgit_definitions=>c_action-change_order_by }?orderBy={ <ls_col>-tech_name }|
             iv_title = <ls_col>-title ).
+        ELSE.
+          lv_tmp = lv_tmp && ri_html->a(
+              iv_txt   = lv_disp_name
+              iv_act   = ``
+              iv_title = <ls_col>-title ).
         ENDIF.
       ENDIF.
       IF <ls_col>-tech_name = iv_order_by
--- a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
+++ b/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
@@ -784,7 +784,6 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
     IF lv_commit_hash IS INITIAL.
       lv_commit_hash = io_repo_online->get_current_remote( ).
     ENDIF.
-    
     lv_commit_short_hash = lv_commit_hash(7).
 
     lv_icon_commit = ii_html->icon( iv_name  = 'code-commit'
--- a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
+++ b/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
@@ -785,10 +785,7 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
           lv_display_url       TYPE zif_abapgit_persistence=>ty_repo-url,
           lv_icon_commit       TYPE string.
 
-    lv_commit_hash = io_repo_online->get_selected_commit( ).
-    IF lv_commit_hash IS INITIAL.
-      lv_commit_hash = io_repo_online->get_current_remote( ).
-    ENDIF.
+    lv_commit_hash = io_repo_online->get_current_remote( ).
     lv_commit_short_hash = lv_commit_hash(7).
 
     lv_icon_commit = ii_html->icon( iv_name  = 'code-commit'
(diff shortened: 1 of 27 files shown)


sbcgua:
I didn't follow the #3040, but the code fix looks OK for me and the normal branch hash work

image

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
```diff
@@ -780,7 +780,11 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
           lv_display_url       TYPE zif_abapgit_persistence=>ty_repo-url,
           lv_icon_commit       TYPE string.
 
-    lv_commit_hash = io_repo_online->get_sha1_remote( ).
+    lv_commit_hash = io_repo_online->get_sha1( ).
+    IF lv_commit_hash IS INITIAL.
+      lv_commit_hash = io_repo_online->get_sha1_remote( ).
+    ENDIF.
```
larshp, line 786: what is the difference between sha1 and sha1 remote? the first is a remote commit hash, so is the second?

larshp (dismissed):

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
mariusraht2: SHA1 can be fetched and set by calling fetch_remote, that's what get_sha1_remote does. get_sha1 just gets the value of SHA1 without fetching it from remote. This is necessary if the user checked out a different commit and set thereby SHA1 to a different value (that of the checked out commit hash).

mariusraht2: In short get_sha1 gets the local value of mv_branch (which is to be renamed) and get_sha1_remote fetches the current repo hash (commit hash) from remote and sets mv_branch to it.

g-back: In that case, wouldn't `fetch_sha1_remote` make more sense?

larshp: can we do the updating/renaming first? I cannot tell what this code does when reading it

switching to a commit or a branch or a tag, I expect to work the same way seen from a user perspective, changing to a branch/tag will refresh the repo

mariusraht2: sure, will create a related PR - method should be renamed too?

larshp: SET_SHA1, is this where the commit sha1 is set if the user chooses a commit? Then I suggest it to be renamed to SELECT_COMMIT

SET_BRANCH_NAME is the similar one, but for branches and tags -> SELECT_BRANCH_NAME

GET_BRANCH_NAME -> GET_SELECTED_BRANCH

GET_SHA1 -> GET_SELECTED_COMMIT

And then MV_BRANCH would be MV_CURRENT_COMMIT ?

mariusraht2: Looks good to me 👍

mariusraht2: What's with get_sha1_remote?

larshp: GET_SHA1_REMOTE, it fetches the remote, it could be a branch or tag or commit, so it might not correspond to a branch

GET_SHA1_REMOTE -> GET_CURRENT_REMOTE, would be symmetric to MV_CURRENT_COMMIT, if you like to get rid of the "SHA1" part

mariusraht2:
@mariusraht2 TODO: Update source code based on PR #4086

larshp (changes requested):
selecting a branch the current remote would be the latest commit on that branch, selecting a commit the current remote would be the selected commit?

ie. is this needed? I guess the current_remote will contain the value we need?

mariusraht2:
No, we need it to determine the branch hash after switching to a different branch: SHA1 is being set to space while switching the branch and therefore needed to be redetermined from remote. That's also the reason why we called `get_sha1_remote` (now `get_current_remote`) before.

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
```diff
@@ -785,7 +785,10 @@ CLASS zcl_abapgit_gui_chunk_lib IMPLEMENTATION.
           lv_display_url       TYPE zif_abapgit_persistence=>ty_repo-url,
           lv_icon_commit       TYPE string.
 
-    lv_commit_hash = io_repo_online->get_current_remote( ).
+    lv_commit_hash = io_repo_online->get_selected_commit( ).
+    IF lv_commit_hash IS INITIAL.
+      lv_commit_hash = io_repo_online->get_current_remote( ).
+    ENDIF.
```
larshp, line 791:
```suggestion
    lv_commit_hash = io_repo_online->get_current_remote( ).
```

larshp:
`get_current_remote` will get the current remote? and take care of it to be correct, for branches, for tags, for selected commits? to get_current_remote it should not matter from the outside, what has been selected by the user?

src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
mariusraht2: > `get_current_remote` will get the current remote? and take care of it to be correct, for branches, for tags, for selected commits? to get_current_remote it should not matter from the outside, what has been selected by the user?

Ahh, yes, we differentiate between them already in fetch_remote..

mariusraht2:
Ahh, yes, we differentiate between them already in fetch_remote..

Issue #4030, opened 2020-10-15
Deletion of local objects should enable the staging button


mariusraht2:
If one deletes an object (e.g. a report) within a package, AG shows the option to pull from remote after refreshing. In this case it should show "Stage" to be able to push the changes (deleted object) to the remote repo without the need to use "Force Stage".

**Reproduction:**
1. Create a test report > Refresh AG > Stage the new object
2. Delete test report in system > Refresh AG

abal-illumiti:
I have the same issue with object deletions. 

1. commit objects to newly created repo from a package.
2. Delete a report. 
3. Object deletion is not recognized.

PR #4084, opened 2020-10-28
Fix state calculation for deleted objects

mbtools:
In case objects are deleted locally or remotely, abapGit now compares against the last known state (checksum). 

Before: Deleted objects are shown as "added" for the opposite side

image

image

After: Deleted objects are properly shown as "deleted" (see lines 3 + 4)
The screenshot also shows other known state combinations that are included in a new test in the test class. 

image

Closes https://github.com/abapGit/abapGit/issues/4030


larshp (approved):
looks good to me 👍

PR #4095, opened 2020-10-28
Add selected commit to repo persistence

mariusraht2:
Related to #3040

Add checked out commit to repo persistence. Changed also related getter and setter.


src/zcl_abapgit_repo_online.clas.abap
```diff
@@ -337,7 +337,7 @@ CLASS ZCL_ABAPGIT_REPO_ONLINE IMPLEMENTATION.
 
   METHOD select_commit.
     reset_remote( ).
-    mv_current_commit = iv_sha1.
+    set( iv_selected_commit = iv_selected_commit ).
```
larshp, line 340: this should also reset the branch name, so the code can know its now a commie SHA1 that the user has chosen?

mariusraht2:
Do we have to adjust `switch_origin` in `zcl_abapgit_repo_online` too?

src/zcl_abapgit_repo_online.clas.abap
mariusraht2: ...and vice versa.

larshp: yes

larshp:
switch_origin: probably, it should reset the selected_commit

mariusraht2:
Hmm...just seeing that it calls `select_branch` if origin has to be switched. That seems ok.

mariusraht2:
Shouldn't empty branch_name though:
image

mariusraht2:
I need it to determine the commits to be selectable by the user

Issue #4098, opened 2020-10-28
UI: Icons for online repo, user id, and transport


mbtools:
I'm not sure why our cloud icon has an arrow in it? Why not just
![image](https://user-images.githubusercontent.com/59966492/97500851-b4673580-1946-11eb-9a11-9160b81585b4.png)

Add icons for user id:
![image](https://user-images.githubusercontent.com/59966492/97500902-cb0d8c80-1946-11eb-8eb0-96c08e85ee03.png)

Change icon for transport from:
![image](https://user-images.githubusercontent.com/59966492/97500920-d365c780-1946-11eb-9d71-53dbe4c95487.png)
to
![image](https://user-images.githubusercontent.com/59966492/97500956-dc569900-1946-11eb-9aa6-9b435fc43385.png)

https://fontawesome.com/icons/cloud?style=solid
https://fontawesome.com/icons/user?style=solid
https://fontawesome.com/icons/truck?style=solid

flaiker:
The transport icon used to be the one you want to change it to and was changed after some discussion because it was apparently more intuitive to some people than the truck.

flaiker:
#2153 , remembered it wrong, there was no truck before ?

mbtools:
Understood. I will try to get SAP to change the truck ;-)

mbtools:
Fontawesome has grown significantly. There are several different trucks available now.

PR #4112, opened 2020-11-01
new icons and css adjustments

sbcgua:
some new icons and css cleanups, related to #4079 and to #4098 
@mbtools 

- added solid cloud, cloud commit (own redesign), truck and user
- adjusted font css, so that it is not fixed width - potentially could lead to some shifts in UI ... to be observed. But in terms of css it is more consistent
- common.css: removed that hard fix at the end, adjusted repo palette css to keep it aligned. Thus fixed palette alignment and Add online button

2020-11-01_11h27_08

image

image


mbtools (approved):
Thank you!

PR #4115, opened 2020-11-01
Remove self-references

mbtools:
Remove unnecessary self-references (me->)

https://github.com/SAP/styleguides/blob/master/clean-abap/CleanABAP.md#omit-the-self-reference-me-when-calling-an-instance-method


larshp:
one remaining in below, after this https://rules.abaplint.org/prefix_is_current_class/ can be enabled
image

larshp (approved):

PR #4165, opened 2020-11-11
Update icon legend

mbtools:


sbcgua:
hmmm, we need to make this automatic. In fact the abapGit font has a build process. One of it's artifacts is a html page showing the available icons and their names. Somehow it should be autobuild on commit and be accessible as a github page. And then cross link it from the abapGit docs. Will create an issue there not to forget

mbtools:
It’s more about the semantics, the “abapGit meaning” of each icon. If we can make that part of the icon repo generated HTML in all for it

PR #4192, opened 2020-11-16
UI: New pages for global and personal settings

mbtools:
Redesign of abapGit settings using HTML forms. Two functional changes to old settings page:
- "Development Internal Settings" are only visible when using the developer version (not in standalone program)
- "User-specific hotkey settings" are not available anymore. The default hotkeys continue to work as before.

Global Settings:

settings-global-1
settings-global-2

(in Dev Version):

settings-global-3

Personal Settings:

settings-personal-1
settings-personal-2


mbtools:
Note: Using <textarea> required some small adjustments to HTML forms and HTML rendering since the innerHTML representing the value of the <textarea> field must not be indented.

mbtools:
Changed from drop-down menu to using a top menu bar for switching between "global" and "personal" settings:

image

larshp:
thanks 👍

thedoginthewok:
https://github.com/abapGit/abapGit/blob/56bd2b1b133ae4b526cb21c56f72615f5fea3127/src/ui/zcl_abapgit_gui_page_sett_glob.clas.abap#L259

This line breaks wildcards in proxy bypass lines. What was wrong with the popup that I put in there? 
I think there should be a check for wildcards or the value should be enclosed in * and option CP should be used.

mbtools:
@thedoginthewok we are replacing all GUI popups step-by-step with HTML pages/forms in order to eventually support a pure web version.

thedoginthewok:
Ah, that makes sense! I guess this might have something to do with cloud-based systems?

mbtools:
yes, https://github.com/abapGit/web_edition

PR #4198, opened 2020-11-17
Improvements for uninstall

### LINKED ISSUE #4197: Improve error handling on deletion
https://github.com/abapGit/abapGit/blob/5e629f19a7243140e9d7a0eb039a411c26e28f83/src/objects/zcl_abapgit_objects.clas.abap#L521-L532

The "report the error later" approach doesn't really work all that well when multiple errors occur, especially follow up errors. I ran into this in this scenario:

- Delete SHI3 -> reports error, structure is still used
- Delete SHI5 -> reports error via popup, SHI3 structure still exists

image

image

You cannot see why the SHI3 structure failed to be deleted. I suggest collecting the errors in a log.

mbtools:
- Improve error handling on deletion collecting errors in log
Closes #4197

- Remove limit of 3 loops which allows deleting objects with deeper dependencies
(Follow-up https://github.com/abapGit/abapGit/issues/3228#issuecomment-728800249)

### DIFF
--- a/src/objects/zcl_abapgit_objects.clas.abap
+++ b/src/objects/zcl_abapgit_objects.clas.abap
@@ -541,7 +541,7 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
 
     zcl_abapgit_default_transport=>get_instance( )->reset( ).
 
-    IF lx_error IS BOUND AND lines( lt_tadir ) > 0 AND NOT ii_log IS BOUND.
+    IF lx_error IS BOUND AND lines( lt_tadir ) > 0.
       RAISE EXCEPTION lx_error.
     ENDIF.
 
--- a/src/objects/zcl_abapgit_objects.clas.abap
+++ b/src/objects/zcl_abapgit_objects.clas.abap
@@ -41,14 +41,15 @@ CLASS zcl_abapgit_objects DEFINITION
       IMPORTING
         !it_tadir  TYPE zif_abapgit_definitions=>ty_tadir_tt
         !is_checks TYPE zif_abapgit_definitions=>ty_delete_checks OPTIONAL
+        !ii_log    TYPE REF TO zif_abapgit_log OPTIONAL
       RAISING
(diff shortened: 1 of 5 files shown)


src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -513,21 +516,32 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
...
+      " Exit if done or nothing else was deleted
+      IF lines( lt_tadir ) = 0 OR lines( lt_deleted ) = 0.
+        EXIT.
+      ENDIF.
     ENDDO.
 
     zcl_abapgit_default_transport=>get_instance( )->reset( ).
 
-    IF lx_error IS BOUND AND lines( lt_tadir ) > 0.
+    IF lx_error IS BOUND AND lines( lt_tadir ) > 0 AND NOT ii_log IS BOUND.
```
flaiker, line 544: Without raising the exception in the case of using the log the uninstall process does not abort on error. So you get the error popup but when you close it the repo is still removed even though objects still exist.

mbtools: back to showing the last error. or should we make it a generic one "Error during uninstall. Check the log"?

after it showed the popup, did it show the "Log" menu item, so you could go back to it?

flaiker: Not sure.

After I closed the automatically opened log popup I was immediately redirected to the repository overview where the repo wasn't listed anymore.

flaiker: I think now it's mostly fine. You get the error at the bottom (a bit generic, yes, but probably caused by the error handling in the specific object type I guess), you get the multiple errors at the top and can also open the log for callstack / longtexts.

image

mbtools:
I changed the msg for the exception at the bottom. Good to go now.

flaiker (approved):

PR #4202, opened 2020-11-18
Implement JUMP for SHI5

flaiker:
Looks like this:

image

This is actually the value help as `STREE_EXTENSION_NAME_EDIT` doesn't allow for deletion or changing the package assignment.


mbtools (approved):
👍

PR #4215, opened 2020-11-22
Catch unhandled exception of zcl_abapgit_repo_srv during screen navigation

PilotFlying:
When repeatedly navigating "back" to the main screen via the "Repository List" [o] button:
If somewhere in this navigation sequence you Remove a repository,
then continue on via "Repository List",
but later you use the GUI's actual "BACK" button, until you have returned to the main screen where that repository still existed,
a short dump occurs as the exception raised by `zcl_abapgit_repo_srv=>get` in `zcl_abapgit_gui_page_repo_view->render_content` propagates unhandled:
[ABAPGIT_DUMP.pdf](https://github.com/abapGit/abapGit/files/5579212/ABAPGIT_DUMP.pdf)


larshp (approved):
looks good to me

PR #4216, opened 2020-11-22
Refactor: Version comparison

mbtools:
Move version comparison from `zcl_abapgit_news` to `zcl_abapgit_version` and make it public for use in other places


src/zcl_abapgit_version.clas.abap
```diff
@@ -4,68 +4,132 @@ CLASS zcl_abapgit_version DEFINITION
...
-    IF lv_prerelease_n IS NOT INITIAL.
-      CONCATENATE rv_version '-' lv_prerelease_n INTO rv_version.
+
+  METHOD compare.
+
+    DATA: ls_version_a TYPE zif_abapgit_definitions=>ty_version,
+          ls_version_b TYPE zif_abapgit_definitions=>ty_version.
+
+    TRY.
+        IF is_a IS SUPPLIED.
```
flaiker, line 106: Imo `IS SUPPLIED` shouldn't be used anymore. If you wrap the call to the method in another one and pass an initial or no value there this will evaluate to true anyways.

PR #4236, opened 2020-11-24
Rebuild local checksums

PeregrinTooc:
as discussed [here](https://github.com/abapGit/abapGit/issues/2127#issuecomment-732777161)

Should be merged after #4231 and #4233


PeregrinTooc:
abandoned.

PR #4237, opened 2020-11-24
unify rebuild_local_checksums method

PeregrinTooc:
redefinition of rebuild local checksums is reduced to redefinition of "get_remotes".

as discussed in #2127


PeregrinTooc:
@larshp please review : -)

PeregrinTooc:
@larshp: can we merge this? There are more and more conflicts to be resolved with every subsequent incoming change I need to integrate into this one...

larshp:
something happend, yea, but I dont think conflicts is a big problem, only due to the way the changes for file `src/zcl_abapgit_repo_online.clas.abap` looks right now, let me try out a few things

larshp:
the changes to repo_online should look somewhat like,

https://github.com/abapGit/abapGit/commit/800b80e2d129e699f1cb279a3b8c4d3bb0760de8

image

currently, it is 

image

I'll go ahead and overwrite the file in the source branch

PeregrinTooc:
@larshp OK. I'm not sure what happened there, I just rebased the branch on abapGit/master a few times. Should I not have done that? I confess that I'm a bit new to this way of working, unfortunately, doing mostly ABAP or solo work, it looks like I got confused.

larshp:
some of the changes might have reordered the methods, and git was successful pulling the upstream changes into the branch, which caused method duplication

Anyhow, I don't really understand why the remote status has something to do with the local checksums, but its outside of this PR scope => #3997

merging, thanks 👍

PR #4308, opened 2020-12-10
Fix master language only, duplicate entries for shi3 objects

jan-jezek:


FreHu (approved):
👍

jan-jezek:
For "Fix duplicates in shi3":
abap_git

FreHu (approved):
We found the second problem on our internal system, not sure why the duplicates appear. It was not caused by the change to language serialization.

larshp:
thanks,

PR #4310, opened 2020-12-10
WDYN: Improve error messages and fix exists check

mbtools:
- Pass exception text to UI/log
- Check for any instead of just active version in `exists`

Related to https://github.com/abapGit/abapGit/issues/4297


larshp (approved):
LGTM 👍

PR #4349, opened 2020-12-22
Checkout commit: additional user interface (#4120)

EduardoCopat:
Implements #4120

When checking out to a commit, changes the branch name to the short commit sha:

image

It is showing a grey box around it. I didn't do anything in particular to change it, but it looks good 🤷‍♂️?

When you try to stage in this scenario, the following pop-up will show:

image

This will create a new branch and change to it. Feel free to make changes to the wording.

Also tested:

- Regular stage - OK
- Going to repo again - OK, shows that it is currently in a commit
- Switching branch - OK
- Creating branch - OK 
- Deleting branch - OK 
- Cancelling the pop-ups - OK


src/ui/zcl_abapgit_gui_router.clas.abap
```diff
@@ -141,7 +141,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_gui_router IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_GUI_ROUTER IMPLEMENTATION.
```
EduardoCopat, line 144: Not sure why this is happening. It is lowercase in my branch: https://github.com/EduardoCopat/abapGit/blob/4120-dev/src/ui/zcl_abapgit_gui_router.clas.abap

mbtools: One of many issues https://github.com/abapGit/abapGit/issues/3741. Just ignore this one.

mbtools (changes requested):
On the surface, this looks great. However, you saved the commit in the branch field of the persistence. This can't be good and will have unintended side effects (if not now, then sometime later). 

For example, anything calling `get_selected_branch` is expecting a branch and not a commit:
https://github.com/abapGit/abapGit/blob/master/src/zcl_abapgit_repo_online.clas.abap#L187 

The proper place to change the display is `render_branch_name`:
https://github.com/abapGit/abapGit/blob/master/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap#L273

mbtools:
We have a similar check when a tag is checked out. There we simply raise an exception:
https://github.com/abapGit/abapGit/blob/master/src/zcl_abapgit_repo_online.clas.abap#L215
This is where I would put the logic for blocking push on commits.

EduardoCopat:
@mbtools Thank you very much for the review and pointers. I was worried about the side effects.

I moved the logic to the `render_branch_name` method. Also, I removed the pop-up and added a validation on push as suggested:

image

mbtools:
- Render look good. 
- The new exception is good too. It comes late (on commit page) but will prevent the bad call to git for example in case on background or API calls without UI. 
- Sorry if you misunderstood. The popup was fine and helpful since it stops going to the staging page. Let's put it back. Maybe also raise the popup for tags `branch_name CP zif_abapgit_definitions=>c_git_branch-tags`

EduardoCopat:
No problems, it is easy to add back. I also added the same validation for tags.

mbtools (approved):
Excellent. Thanks 👍

larshp:
thanks 👍

Issue #4265, opened 2020-11-30
Dialogs: Check for unsaved changes in forms


mbtools:
When navigating away from an HTML form, abapGit does not check for unsaved changes. 

Key challenge: 

When using the "Back" button on the form, the current form is submitted and can be evaluated for unsaved data. However, when selecting a menu item (HTML link) or clicking the GUI back button, the current form data is not submitted. This makes it currently impossible to find out if the user made changes.

mbtools:
This could be mitigated by using a different type of menu item: Instead of a `a-href`, we could use a `button`.

mbtools:
One of Alexander's PRs fixed this. Closing

PR #4384, opened 2021-01-05
Dialogs: HTML forms for repo settings

mbtools:
This moves the repo settings to HTML forms. The settings are split into three pages:
- Repository settings (ie abapgit.xml)
- Local settings
- Infos (some statistics about your repo)

I temporarily took out the "Remotes" section. It will come back in a separate PR.

#4171

image

image

image

### DIFF
--- a/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
+++ b/src/ui/zcl_abapgit_gui_chunk_lib.clas.abap
@@ -109,6 +109,12 @@ CLASS zcl_abapgit_gui_chunk_lib DEFINITION
         !iv_act        TYPE string
       RETURNING
         VALUE(ro_menu) TYPE REF TO zcl_abapgit_html_toolbar .
+    CLASS-METHODS settings_repo_toolbar
+      IMPORTING
+        !iv_key        TYPE zif_abapgit_persistence=>ty_repo-key
+        !iv_act        TYPE string
+      RETURNING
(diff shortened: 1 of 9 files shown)


sbcgua:
works nicely. A couple of comments:

1) "Back" returns to the repo page. While F3 returns to previous settings. e.g. I click Local, then infos, then local again - all of that goes to stack. I'd expect F3 always goes to repo page, so the subsettings behave like tabs. However a proper implementation may need proper solution for #3594 ... Also, what should happen if I change tab 1, and click to tab 2 without saving ? 
2) Design have a room for improvement. light header looks ... not ideal. Also I'd think about removing border around form - too many lines

Both can be solved separately imho and not blocking this PR. If there is a quick fix for 1, I'd vote for it. But the rest looks good. And cool stats page (it wasn't there before right?) :) 

P.S. BTW maybe "Stats" instead of "Infos" ?

mbtools:
1. I couldn't find a perfect way for navigation. The jump-off page is bookmarked and when you use the "back" button on the form, it does go to the repo view. However, F3-back doesn't do the same. 
When going from "tab" to "tab", it's an a-href and the form is not submitted. It might need so JS to check if form content was changed and give a warning to the user.  https://github.com/abapGit/abapGit/issues/4265 

2. Some tweaking, maybe, but it's so much better than the old pages already

sbcgua (approved):
looks fine for me. 

The comments above could be fixed later

PR #4437, opened 2021-01-22
Add refresh button to stage page

FreHu:


sbcgua (approved):
works, useful feature :+1:

PR #4508, opened 2021-02-11
TABL: Remove field position from serialization

### LINKED ISSUE #4506: Pull Repo: ERROR while importing Object
This issue references to [Pull Repo: ERROR while importing Object [...] from the DB #729](https://github.com/sapmentors/abap2xlsx/issues/729) in Repo `ABAP2XLSLX` 

**System**: SAP ERP, Netweaver 7.50
**Problem**: `.INCLUDE BAPIBUS1006002_CENTRAL ` in Structure `ZEXCEL_S_ORG_REL ` has more fields than expected

While initially Pulling the ABAP2XLSX repo an error is occurring:
image

Structure: 
image

Like @sandraros detected: 

> In your system, it seems that .INCLUDE BAPIBUS1006002_CENTRAL is at the 109th position, but in my S/4HANA 1709 system, it's at the 108th position. So it means that in your system one of the previous SAP structures has more fields.

mbtools:
It's not necessary to serialize field position (`DD03P-POSITION`). Fields can simply be numbered sequentially during deserialization.

This avoids issues with include structures that have a different number of fields in original and target systems.

Closes #4506

Note: This change will initially cause a lot of diffs for structures and tables. Push the updated objects to your online repos to get back to no diffs.


larshp:
heh, I did not test this yet, but previously wrong numbering created a bug, now we are doing what also seems like wrong numbering, and it will fix the bug?

mbtools:
There's a difference between DD03L on DB and DD03P in the AG xml
- DD03L has a field list with expanded includes. this is where you can get different position numbers if the includes are different
- DD03P does not expand the includes. however, the xml nevertheless contains the position corresponding to DD03L. For example https://github.com/sapmentors/abap2xlsx/blob/master/src/zexcel_s_org_rel.tabl.xml

By removing <position> from the xml and just numbering sequentially, `DDIF_TABL_PUT` will automatically expand the includes and update DD03L accordingly (at least with my test).

larshp:
thanks, needed that story for it to fit together 👍

larshp:
actually, this is an opportunity, can something like
```
    <DD03P>
     <TABNAME>ZEXCEL_S_ORG_REL</TABNAME>
```
also be cleared?

perhaps also  `<DDLANGUAGE>E</DDLANGUAGE>` as its the master language=

larshp:
@seniorFreshmen can you help testing if this change fixes the issue?

larshp:
the changes look good, but I'll wait merging it until tagging a real release, as its important to have these changes in the changelog

larshp:
@seniorFreshmen see comment above

yellappam:
ABAP: 75D
abapGit: 1.105.0
**Scenario1**: Include structure and table created with position parameter and number of fields in the include structure remain the same during PULL
**Expected**: Table must be created and diff should show position and table name fields

First we create the include structure in the system. This resembles the standard include and we will add additional fields in the next scenario.

Include Structure: ( This resembles standard structure where we will add additional fields in the next scenario )
image

Before PULL( Structure ):
image

After PULL( Structure ): The diff is expected which would show that position and table name parameters are not there in local
image

image

Before PULL( Table ):
image

After PULL( Table ): The diff is expected which would show that position and table name parameters are not there in local
image

image

image

The output is as expected hence, PASS.  

PS: In the next scenario, will add additional parameters to include structure and repeat the test. Coming soon...

mbtools:
At the end of this, you should commit the local version to the repo. In other words, what you tested is confirming the the change is compatible with existing repos. With the change tables will be serialized differently.

yellappam:
Yes, we need to commit the diffs to the repo. Intentionally, I haven't committed as I need to use same data for next scenario.

yellappam:
ABAP: 75D
abapGit: 1.105.0
**Scenario2**: Table created with position parameter and number of fields in the include structure changes in the destination system( Backward compatibility test )
**Expected**: Table must be created and number of fields under include structure, within the table, must match number of fields in the destination system's structure. It should also produce diff with position and table field parameters

1. Below is the screenshot of structure and table in the source system which is captured in the repo

image

image

2. The structure is modified in the destination system with additional field FIELD3

image

3. Before PULL( Table )

image

4. After PULL( Table ). Diff is expected to show for position and table name as per new logic

image
image

Table successfully created and number of fields within the include matches destination system. After PULL, don't forge to stage the changes. The test is PASS
image

In the next scenario, we will test STAGE and PULL process completely with new logic where we shouldn't have position and table parameters in the files.

yellappam:
ABAP: 75D
abapGit: 1.105.0
**Scenario3**: For object type TABL, <TABNAME>, <POSITION> and <DDLANGUAGE> should be removed 
**Expected**:  Below tag entries must not appear in the files during the serialization

    <DD03P>
     <TABNAME>YTABL_4506</TABNAME>
     <POSITION>0001</POSITION>
     <DDLANGUAGE>E</DDLANGUAGE>

    <DD08V>
     <TABNAME>YTABL_4506</TABNAME>

1.  Below are the structure and table

image
image

2. Before Staging

image

3. After staging

image

**DD03P**
image

**DD08V**
image

The excepted tags are not present under DD03P and DD08V tags. The test case is PASS

yellappam:
@mbtools - During the testing, I have observed <DDLANGUAGE>E</DDLANGUAGE> is present under <DD08V>. Can we remove this field?

image

yellappam:
Thank you for making the changes. It looks good.

larshp:
thanks 👍 merging, tagging new version and adding changelog in a bit

PR #4521, opened 2021-02-16
SUSH - raise error if not supported

### LINKED ISSUE #4520: SUSH: dump during import when system doesn't support SUSH
When I import the SUSH demo repo https://github.com/abapGit-tests/SUSH.git into my NW752 SP04 dev edition I get this dump as SUSH is not supported there I guess.
image

Would prefer an error message instead.

christianguenter2:
fixes #4520

image


mbtools (approved):
👍

PR #4541, opened 2021-02-20
Repo overview performance improvements

FreHu:
On our internal system there are now over 700 packages versioned and this is causing problems with startup time and the performance of the repo overview in general.

To avoid multiple loads of the  repo list, instead of 
`zcl_abapgit_persist_factory=>get_repo( )->list( )` I am using `zcl_abapgit_repo_srv=>get_instance( )->list( ).` which is cached.

For this reason I made `ms_data` on zcl_abapgit_repo public read-only instead of protected. People don't like public attributes, but abap returns by value and does not inline methods , so really they are great in anything performance intensive. You can also see calls to `repo->get_key( )` also add up to take over half a second. I will go over those as well.

Measurement of opening abapgit and waiting until repo list is loaded:
## Before
image
image
## After
image
### Current hitlist
image
### Selects over 100ms
image

Other ideas:
- the favorites logic should be server-side and not a client side toggle, then we can get away with only loading the favorite repositories on startup


larshp:
also note https://github.com/abapGit/abapGit/pull/4539 is currently open which touches the same object

larshp:
favorites: I'd personally like it to be server side

1½ million calls to get_key() seems off, probably some places can be fixed with caching into a local variable, something for a different pull request?

the call to function module SUSR_USER_ADDRESS_READ, try caching the name for each user in a static variable, guess multiple usernames are identical, something for a different pull request?

src/ui/zcl_abapgit_gui_router.clas.abap
```diff
@@ -234,16 +234,21 @@ CLASS ZCL_ABAPGIT_GUI_ROUTER IMPLEMENTATION.
...
     lv_key = ii_event->query( )->get( 'KEY' ).
 
     CASE ii_event->mv_action.
       WHEN zcl_abapgit_gui=>c_action-go_home.
         lv_last_repo_key = zcl_abapgit_persistence_user=>get_instance( )->get_repo_show( ).
-        lt_repo_list = zcl_abapgit_persist_factory=>get_repo( )->list( ).
-
+        lt_repo_obj_list = zcl_abapgit_repo_srv=>get_instance( )->list( ).
+        LOOP AT lt_repo_obj_list ASSIGNING <lr_repo_obj>.
+          APPEND <lr_repo_obj>->ms_data TO lt_repo_list.
```
larshp, line 250: `lt_repo_list` is only used to check if repos exists, the list `lt_repo_obj_list` can be used for this

FreHu:
I will keep this in draft until the smaller stuff is resolved and split separate prs from this if possible.

FreHu:
Testing with 20 favorites, only loading the extra info for favorites seems to do the trick.
image
image

Overall my startup time is currently down to 4.6s from yesterday's 11.

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -320,6 +319,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_OVER IMPLEMENTATION.
         lv_favorite_icon = 'star/blue'.
         lv_favorite_class = 'favorite'.
       ELSE.
+        lv_suppress_title = abap_true.
```
larshp, line 322: from a user perspective, I think this can easily be considered a bug, some places titles are shown and some not

Suggest buffering the usernames

for the package description TDEVCT is fully buffered on the application server, so it should be very fast

FreHu: really you open the non-favorites only if you want to favorite a repo, and then the alt text will be shown

FreHu: image
tdevct is actually the most problematic, there are not going to be many usernames

FreHu: also `susr_user_address_read` already contains buffering logic

larshp: okay, then what about always skipping the titles on the overview page for everything? that would also give a consistent/more transparent experience

larshp: alternatively, linking users and packages to 2 new pages, which can show the full information regarding the user and packages

^new pages, so it works in a possible future web edition

FreHu: Now that you mention it, the package is already a link to the package object and clicking a username shows a popup with the complete user information. I no longer see any point in the descriptions at all. @mbtools any strong feelings about this?

mbtools:
half a sec for 702 buffered select singles seems high. it's about the same speed as your unbuffered selects. 

Reading package and user names should not be a perf problem at all. I'm ok with merging this but would like to get back to it with a bit of code to buffer the selects (f.e. using a select for all entries for all involved repo packages and user id). 

A problem is actually the size of the XML in zabapgit. The design is not sustainable. Just for fun, can you look at the memory consumption for reading you 700 repos from the persistency? Keeping the checksum in the XML is blowing up the size and processing time tremendously. It's also making changes like #3997 very difficult. Imho, checksums should go into a separate transparent table (not as XML and covering the complete commit history with branches). But that's a different discussion...

FreHu:
I agree that this can be optimized enough not to matter, but let's do it piece by piece. I'll get back with some memory measurements.

PRs can be merged in the order
#4543, #3997
#4545 
this

FreHu:
The abapgit process sitting idly in the home page is consuming 243 518kb of memory accodring to SM04.

A memory trace never quite seems to finish, this is how far I've got:
[zipped html report from incomplete SAT trace.zip](https://github.com/abapGit/abapGit/files/6015750/zabapgit_memory.zip)

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -600,4 +599,4 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_OVER IMPLEMENTATION.
     <ls_col>-add_tz = iv_add_tz.
     <ls_col>-allow_order_by = iv_allow_order_by.
   ENDMETHOD.
-ENDCLASS.
+ENDCLASS.
```
larshp, line 602: I get a empty diff in abapGit after pulling, due to this whitespace change?

FreHu: fixed, tried pulling and there's no diff

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -301,6 +299,7 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_OVER IMPLEMENTATION.
       lv_zip_import_link TYPE string,
       lv_zip_export_link TYPE string,
       lv_check_link      TYPE string,
+      lv_suppress_title  TYPE abap_bool,
```
larshp, line 302: this variable, `lv_suppress_title` can be deleted?

larshp (approved):
thanks 👍

PR #4545, opened 2021-02-20
Main page - disable hover info for package/user links

FreHu:
part of #4541 

Disabled for performance reasons when there are a lot of repositories. 

Possible follow-ups:
- selecting entries at once and buffering
- most of the time is spent rendering things which are not visible, favorites/detail toggle should be server side logic


larshp (approved):
lgtm

PR #4548, opened 2021-02-20
Optimize performance of switching branches

FreHu:
When switching branches, if you have a checked out commit it is also reset. This reset is expensive and I think it can be skipped if no commit was selected previously.

image


mariusraht2 (approved):
Looks good to me!

larshp (approved):
lgtm

PR #4564, opened 2021-02-26
TABL: Skip objects in status new

yellappam:
Fixes [issue_4562](https://github.com/abapGit/abapGit/issues/4562)

### DIFF
--- a/src/objects/core/zcl_abapgit_skip_objects.clas.abap
+++ b/src/objects/core/zcl_abapgit_skip_objects.clas.abap
@@ -1,22 +1,33 @@
-CLASS zcl_abapgit_skip_objects DEFINITION PUBLIC FINAL CREATE PUBLIC.
+CLASS zcl_abapgit_skip_objects DEFINITION
+  PUBLIC
+  FINAL
+  CREATE PUBLIC .
 
   PUBLIC SECTION.
-    METHODS:
-      skip_sadl_generated_objects
-        IMPORTING
-          it_tadir        TYPE zif_abapgit_definitions=>ty_tadir_tt
-          ii_log          TYPE REF TO zif_abapgit_log OPTIONAL
-        RETURNING
-          VALUE(rt_tadir) TYPE zif_abapgit_definitions=>ty_tadir_tt.
+    METHODS skip_sadl_generated_objects
+      IMPORTING
+        !it_tadir       TYPE zif_abapgit_definitions=>ty_tadir_tt
+        !ii_log         TYPE REF TO zif_abapgit_log OPTIONAL
+      RETURNING
+        VALUE(rt_tadir) TYPE zif_abapgit_definitions=>ty_tadir_tt .
+    METHODS skip_tabl_status_new_objects
+      IMPORTING
+        !ii_log   TYPE REF TO zif_abapgit_log OPTIONAL
+      CHANGING
+        !ct_tadir TYPE zif_abapgit_definitions=>ty_tadir_tt .
   PROTECTED SECTION.
   PRIVATE SECTION.
-    METHODS:
-      has_sadl_superclass
-        IMPORTING
-          is_class         TYPE zif_abapgit_definitions=>ty_tadir
-        RETURNING
-          VALUE(rv_return) TYPE abap_bool.
 
+    METHODS has_sadl_superclass
+      IMPORTING
+        !is_class        TYPE zif_abapgit_definitions=>ty_tadir
+      RETURNING
+        VALUE(rv_return) TYPE abap_bool .
+    METHODS is_tabl_status_new
+      IMPORTING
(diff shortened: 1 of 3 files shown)


yellappam:
Unit Testing:

Table:
image

Structure:
image

abapGit:
image

mbtools:
this works, but we should avoid introducing more object type-specific logic (the skip_sadl should be replaced eventually, too. see #4187). 

tadir is already filtered using the `exists` method. so the logical place to avoid serialization would be `ZCL_ABAPGIT_OBJECT_TABL->EXISTS`. 

However, this goes back to #4105 where I added the following to fix uninstalling such tables:

```abap
      " Check for new, inactive, or modified versions that might not be in nametab
      SELECT SINGLE tabname FROM dd02l INTO lv_tabname
        WHERE tabname = lv_tabname.
```

I think it's wrong to completely ignore such objects. a "new" table "exists" and therefore should appear in the repo view. it should be marked as "inactive" (yellow bolt). This way it's clear the object exists but needs to be maintained/activated before you can stage it. so the question is really, was it marked inactive? no, fix that. yes, fix why staging was possible.

BTW, this is the same with other DDIC types which also can have a "new" version. 

PS: Even for fixes that seem obvious, it's good to briefly discuss in the issue how to implement it (I learned that myself here).

mbtools (changes requested):
see above

yellappam:
Agree with your comments. Currently, the new TABL objects are shown inactive (yellow bolt) hence, no changes in this space. Are you suggesting to stop the staging for such objects?

mbtools:
Yes. Only active objects are serialized. If an inactive version exists, it's quite confusing what would actually end up in the remote repo.  i would therefore exclude the object at staging: `ZCL_ABAPGIT_STAGE_LOGIC->GET` `rs_files-status` contains `inactive` flag already so it would be an easy change there.

yellappam:
Currently, the inactive objects are also serialized and this feature might be useful for certain object types. Example: If I am making changes to a complex functionality which includes programs, classes, function groups etc.. and in the middle of the logic, an urgent production issue will come and I need to stop my current changes and work on the production issue. Suppose, I couldn't activate my objects due to syntax errors or any other reason, I can't stage my changes unless my objects are in active state. This will delay production issue fix.

Probably, better solution would be to give an option to the user to decide whether to stage inactive objects or not though some global settings or repo settings. Any thoughts?

larshp:
abapGit only serializes active objects. Like this issue a TABL is in a inconsistent state with no rows if its in status New, more of these kinds of errors might occur if opening for serializing inactive objects.

Git is just files, so if the developer would like to do some changes outside the central development system, this is possible, just edit the files via the developers favorite git client and editor.

yellappam:
Little confused with the statement "abapGit only serializes active objects".  I thought, it serializes inactive objects as well based on the below example. Please correct if I am wrong.

Step 1: Activate program and commit changes

image

Step 2: Make changes to the program and save it( now in inactive state ).

image

Step 3: abapGit serialized the inactive program and below is the diff

image
image

Here, inactive program is serialized and shows latest changes in local. If we stop staging the inactive objects then latest changes can't be staged though they are visible in diff. In this case, are we still suggesting not to allow staging for this report?

larshp:
heh, okay, I would regard that as a bug 😀

yellappam:
Thank you for the clarification. Now, it is clear :). I will make the changes accordingly.

yellappam:
Stage logic change: Ignore any object with status new/inactive
Unit Testing:

New/Inactive objects:
image
image
image

Objects with new/inactive appears as "inactive" (yellow bolt) on main page
image

New/inactive objects not shown on stage screen:
image

PS: Only rs_files-local data deleted in staging logic as this is the main internal table used to display the objects on stage screen. Ideally, rs_files-status data should also be deleted however, the status data is being used in the below logic which dumps due to ASSERT sy-subrc = 0 logic. We can fix this as part of separate issue or combine with current change?

image

larshp:
this goes outside of the scope for the original issue, and might have side effects in a lot of places, eg. I expect to be able to stage the active version if an inactive version exits

as Marc hinted above, to skip the TABL objects in status new, changes are expected in ZCL_ABAPGIT_OBJECT_TABL->EXISTS

yellappam:
We might need to distinguish between New and Inactive version of an object. In addition to EXISTS method, we might need to check IS_ACTIVE method as well. Will analyze further...

mbtools:
you can stop looking. I have a general solution. it's changing the core XML output logic so it will need some good testing. this PR might not be necessary. let's put it into draft.

yellappam:
I am happy to help in testing after implementing general solution

mbtools:
Should be closed by #4584

larshp:
@yellappam & @mbtools thanks, after a bit back and forth I think we can move to #4584

the task seemed easy at first, but sometimes there is complexity to it 😃 

closing

Issue #4569, opened 2021-02-27
Empty serialized result leads to incomplete XML


mbtools:
If the serializer class does not return any result (only empty files), abapGit creates an "empty" XML. Example (#4564):

```xml
<?xml version="1.0" encoding="utf-8"?>
<abapGit version="v1.0.0" serializer="LCL_OBJECT_TABL" serializer_version="v1.0.0">
 <asx:abap version="1.0" xmlns:asx="http://www.sap.com/abapxml"/>
</abapGit>
```

Suggestion: 

Instead, the resulting file should be empty (to still allow diffs), but it should not be possible to stage the file. 

Required changes:

- [ ] Rendering of XML when result was empty
- [ ] Ignoring empty files when staging

mbtools:
Solved by #4584 for DDIC objects. Issue is still a possibility for other object types but closing this now

PR #4574, opened 2021-02-28
Ignore empty files when staging

mbtools:
Empty files should not be staged. Empty files can happen if objects exist but not in an active version (like a new table).

Related to #4569


larshp:
something like a SMIM object is a container for a different file, these might be empty

I think TABLs in New status should be considered non existing

mbtools:
If new TABL is "non-existing", it can't be uninstalled. It's in the system, so why hide it?

Empty mime-objects? Not sure if that's possible. I will check

larshp:
hmm

larshp:
a active ABAP include might be empty,
image

larshp (changes requested):
wip

mbtools:
Closing in favor of #4584

PR #4576, opened 2021-02-28
docs for i18n features

sbcgua:
docs for #4494. A bit drafty ... but better than nothing :)


mbtools (approved):
We can add details/screenshots later. Goed to merge 👍

PR #4584, opened 2021-03-01
DDIC: Error message in case of "new" versions

mbtools:
In case DDIC objects exist as "new" version only , the system will now raise an error message to alert the user that "No active version was found". 

image

Note: The log will disappear after some navigation and such objects will not be included in the repo view if they only exist locally (until they are activated).

The `, 1` at the end of the messages is the `sy-subrc` value (when serializing in parallel). Should be changed in separate PR.


larshp (approved):
looks good to me, will merge the change tomorrow

PR #4590, opened 2021-03-02
Log refactoring

sbcgua:
to #2821

This might need some reviewing and testing. I will yet adjust repo class itself to remove unnecesary methods.

@larshp But it already demonstrates what I was talking about in #2821. So in essence:

1) `zcx` may optionally receive a log instance
2) Example of usage is in `zcl_abapgit_objects=>delete` - all handling on it's way is removed and it bubbles up till ...
3) ... the GUI catches the exception and checks of log is there, and show is. (yet to do the `gui_page` error rendering feature - I didn't touch it)
4) the log class got 2 new features (not used in the code though yet): `merge_with` and `from_exception`. The first will be very useful to merge logs created in different parts of the code (e.g. mass serialization). The second is for converting one sception to a log, shell we decide to choose log as a main way to communicate with user
5) added some UTs for the log

UPD:
6) log can also `clone`, and `merge_with( min_level )`, where level is severity of message

The rest of changes are more or less unifications of log usage in various places (needs testing)

### DIFF
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -122,9 +122,6 @@ CLASS zcl_abapgit_repo DEFINITION
         !iv_offline TYPE abap_bool
       RAISING
         zcx_abapgit_exception .
-    METHODS get_log
-      RETURNING
-        VALUE(ri_log) TYPE REF TO zif_abapgit_log .
     METHODS refresh_local_object
       IMPORTING
         !iv_obj_type TYPE tadir-object
@@ -524,11 +521,6 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
 
     rs_settings = ms_data-local_settings.
 
-  ENDMETHOD.
-
-
-  METHOD get_log.
-    ri_log = mi_log.
   ENDMETHOD.
 
 
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -75,8 +75,6 @@ CLASS zcl_abapgit_repo DEFINITION
     METHODS refresh
       IMPORTING
         !iv_drop_cache TYPE abap_bool DEFAULT abap_false
-        !iv_drop_log   TYPE abap_bool DEFAULT abap_true
-          PREFERRED PARAMETER iv_drop_cache
       RAISING
         zcx_abapgit_exception .
     METHODS update_local_checksums
@@ -124,11 +122,6 @@ CLASS zcl_abapgit_repo DEFINITION
         !iv_offline TYPE abap_bool
       RAISING
         zcx_abapgit_exception .
-    METHODS create_new_log
-      IMPORTING
-        !iv_title     TYPE string OPTIONAL
-      RETURNING
-        VALUE(ri_log) TYPE REF TO zif_abapgit_log .
     METHODS get_log
       RETURNING
         VALUE(ri_log) TYPE REF TO zif_abapgit_log .
@@ -305,14 +298,6 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
 
     ms_data = is_data.
     mv_request_remote_refresh = abap_true.
-
-  ENDMETHOD.
-
-
-  METHOD create_new_log.
-
-    CREATE OBJECT mi_log TYPE zcl_abapgit_log EXPORTING iv_title = iv_title.
-    ri_log = mi_log.
 
   ENDMETHOD.
 
@@ -649,10 +634,6 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
 
     mv_request_local_refresh = abap_true.
     reset_remote( ).
-
-    IF iv_drop_log = abap_true.
-      CLEAR mi_log.
-    ENDIF.
 
     IF iv_drop_cache = abap_true.
       CLEAR mt_local.
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -68,7 +68,8 @@ CLASS zcl_abapgit_repo DEFINITION
     METHODS deserialize
       IMPORTING
         !is_checks TYPE zif_abapgit_definitions=>ty_deserialize_checks
-        !ii_log    TYPE REF TO zif_abapgit_log
+      RETURNING
+        VALUE(ri_log) TYPE REF TO zcl_abapgit_log
       RAISING
         zcx_abapgit_exception .
     METHODS refresh
@@ -331,6 +332,8 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
     DATA: lt_updated_files TYPE zif_abapgit_definitions=>ty_file_signatures_tt,
           lx_error         TYPE REF TO zcx_abapgit_exception.
 
+    CREATE OBJECT ri_log TYPE zcl_abapgit_log EXPORTING iv_title = 'Deserialization log'.
+
     find_remote_dot_abapgit( ).
     find_remote_dot_apack( ).
 
@@ -353,7 +356,7 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
         lt_updated_files = zcl_abapgit_objects=>deserialize(
             io_repo   = me
             is_checks = is_checks
-            ii_log    = ii_log ).
+            ii_log    = ri_log ).
       CATCH zcx_abapgit_exception INTO lx_error.
 * ensure to reset default transport request task
         zcl_abapgit_default_transport=>get_instance( )->reset( ).
@@ -362,7 +365,7 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
 
     APPEND get_dot_abapgit( )->get_signature( ) TO lt_updated_files.
 
-    CLEAR: mt_local.
+    CLEAR mt_local.
 
     update_local_checksums( lt_updated_files ).
     update_last_deserialize( ).
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -310,9 +310,7 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
 
   METHOD create_new_log.
 
-    CREATE OBJECT mi_log TYPE zcl_abapgit_log.
-    mi_log->set_title( iv_title ).
-
+    CREATE OBJECT mi_log TYPE zcl_abapgit_log EXPORTING iv_title = iv_title.
     ri_log = mi_log.
 
   ENDMETHOD.
--- a/src/utils/zcl_abapgit_log.clas.abap
+++ b/src/utils/zcl_abapgit_log.clas.abap
@@ -8,13 +8,8 @@ CLASS zcl_abapgit_log DEFINITION
   PROTECTED SECTION.
 
     TYPES:
-      BEGIN OF ty_msg,
-        text TYPE string,
-        type TYPE sy-msgty,
-      END OF ty_msg .
-    TYPES:
       BEGIN OF ty_log, "in order of occurrence
-        msg       TYPE ty_msg,
+        msg       TYPE zif_abapgit_log=>ty_msg,
         rc        TYPE sy-subrc,
         item      TYPE zif_abapgit_definitions=>ty_item,
         exception TYPE REF TO cx_root,
@@ -90,7 +85,7 @@ CLASS ZCL_ABAPGIT_LOG IMPLEMENTATION.
 
     DATA lx_exc TYPE REF TO cx_root.
     DATA lv_msg TYPE string.
-    lx_exc ?= ix_exc.
+    lx_exc = ix_exc.
     DO.
       lv_msg = lx_exc->get_text( ).
       zif_abapgit_log~add( iv_msg  = lv_msg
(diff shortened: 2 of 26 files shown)


sbcgua:
Hmm, looks like the standalone version moves the exception class to the top (makes sense) but it requires zif_abapgit_log. Which by the way require zif_abapgit_definitions ...

larshp:
hmm, and the some of the interfaces use the exception classes, perhaps adding more logic to abapmerge will fix this scenario

src/utils/zcl_abapgit_log.clas.abap
```diff
@@ -202,21 +226,22 @@ CLASS ZCL_ABAPGIT_LOG IMPLEMENTATION.
...
           CONTINUE.
         WHEN 'S' OR 'I'.
-          IF rv_status <> 'W'.
-            rv_status = 'S'. "okay
+          IF rv_status <> zif_abapgit_log=>c_status-warning.
+            rv_status = zif_abapgit_log=>c_status-ok.
           ENDIF.
           CONTINUE.
         WHEN OTHERS. "unknown
+          rv_status = zif_abapgit_log=>c_status-error. " it is a bug probably, so not OK
```
mbtools, line 244: `assert 0 = 1.` so we can find out what made the log entry.

src/repo/zcl_abapgit_repo.clas.abap
```diff
@@ -305,9 +306,7 @@ CLASS zcl_abapgit_repo IMPLEMENTATION.
 
   METHOD create_new_log.
```
mbtools, line 307: do we still need this?

mbtools:
thanks for the refactor. much better this way. 👍 

i don't know the abapmerge internals, but can't the exception classes be handled with the same dependency tree/ check like the other classes and interfaces? On a quick view, I don't see any circular deps, so it should work.

src/utils/zcl_abapgit_log.clas.abap
sbcgua: agree

src/repo/zcl_abapgit_repo.clas.abap
sbcgua: No, will fix it in the next commits (to this PR), just didn't have time to finish.

sbcgua:
abapmerge: it is not a circular dependency. It is just the fact that zcx_exception is defined (forcefully) at the very top of the file. I think either definitions&zif_log should also forced to the top. Or (crazy idea) define log as a `ref to object` in `zcx` and some simple way to convert it to log back e.g. `zcl_log=>cast_obj_to_log( )`. Not sure which is better. The latter approach is a crutch ... but it decouples things ... :)

src/repo/zcl_abapgit_repo.clas.abap
sbcgua: fixed

src/utils/zcl_abapgit_log.clas.abap
sbcgua: done

sbcgua:
Hehe, now this pops up at every deserialize :))))

image

Maybe still some room for fixes ... :)

mbtools:
We should also check if "Log" shows up in the menu (if it contains messages). It was one of the things that didn't work with uninstall.

mbtools:
the standalone works if you first put all `interface deferred`, then all `zcx* classes`, then `class deferred`, then all `zcl* classes`.

sbcgua:
> We should also check if "Log" shows up in the menu (if it contains messages). It was one of the things that didn't work with uninstall.

It doesn't. I mean, I didn't check but current does not suppose this really. Needs some more refactoring to get it back

sbcgua:
OK, so ready for review again. I wouldn't say this is ideal. But I got rid of the log in the repo. The log class gained some useful features. And the log handling os more or less unified in UI (though there is a room for improvement). Needs a review and testing.

P.S. abapmerge issues are fixed in relevant PRs.

mbtools:
Code looks good on my iPhone 😀 Many touched files. I will test tomorrow to get this merged quickly.

mbtools:
I tested deserialize and uninstall. took a long look and added some comments.

- deserialize: always log popup (easy fix provided)
- uninstall: no log popup and not error message althought exception was raised
- deserialize and uninstall: log not available in menu

the code is certainly better but we lost a core feature: being able to view the log after the process is done no matter the state. we should be able to view the log even if it's all green and also in case of any exception.  

the key problem I see is that the log will only be available (via `mi_last_log`) if an exception was raised *with* `ii_log` set. however, if any other exception occurs before that or if no exception is raised at all, then `mi_last_log` is not set and does not show in the menu. 

now there's a mixture between exceptions and the log which makes things more complicated. we basically have to think for every exception whether it's caught and added to a log or whether it needs the log object and set it itself. i think it's better to keep them separated. 

the only way i can see logging work consistently, is if the UI layer creates the log and passes it to the process. the process can then log whatever it wants. it can catch exceptions and log them or let other exceptions raise up to the UI (stopping the process). but the log would always be there in the UI. 

we have `mi_last_log` at `gui_page` level. that's good and it should be the only log object we need. here's how i think it should work for example in `gui_router->repository_services`:

```abap
WHEN zif_abapgit_definitions=>c_action-repo_purge.
  " set the log in event constructor or via gui_services to reference mi_last_log
  li_log = ii_event->get_log( )->clear( )->set_title( 'Uninstall log' ).
  zcl_abapgit_services_repo=>purge( iv_key = lv_key ii_log = li_log ).
  if li_log->li_log->get_log_level( ) >= zif_abapgit_log=>c_log_level-warning. 
     zcl_abapgit_log_viewer=>show_log( li_log ).
  endif.
```

this would work with any process that wants logging. it also remove the create log and show log logic from the process(es).

i know that's a lot and quite a bit different but it's a tricky subject that really needs work.

mbtools (commented):
see below

sbcgua:
Yeah ... I wanted to add a log entry to display unsupported LXE langs ... and didn't find a good place so decided to refactor the deserialization bit ... but it required some adjustments in log ... so I thought `why not to fix this also quickly, it does not seem complicated` ... hehe :)

Coming back to the matter, thanks for the review. Fair comments. I'll write a more detailed reply later. It is very controversial. Seems like proper log does not fit well into current code structure at all :(

sbcgua:
P.S. abapmerge fix worked

sbcgua:
Requirements:
- log is the result of operation, negative or also positive
- not all operations can produce log
- not all of operations should reset current log state
- we want to decouple log from repo
- we want a proper systematic place/approach to keep log
- log may have a title (questionable, see below)

So let's "helicopter view"

**Problem one: location of the log**

We have GUI class as a singleton controller for all UI stuff, it contains the stack of event handlers and the current page. The current page may consist of components (if we follow the POC I made with new-online page and imho it is a proper way - remove inheritance, introduce composition).

2021-03-13_12h36_53


Each component may be an event handler as well. + the router. A user may call a command which may land in **different event handlers** (including global router which does not know about current page). Also consumers of the log may appear on different levels potentially (if log is available - hide/show buttons)

So we need some common denominator - where the log will stay. And page is not the best place. Because of the router. And because of the fact that page may consist of decoupled components without common base (in current concept there is `mx_error` or `mi_last_log` in the page_base but it is bad imho).

The only common denominator seems to be the GUI instance (accessible via GUI services). Which makes log a global variable. But it is a global variable anyway. More on that below.

**Problem 2: where the log instantiates**

There are 2 approaches - from top or from bottom. Currently most of the code supposes "from top" - `ii_log` is passed as a param. So it must traverse all the layers. I think it is suboptimal. And less flexible. And demand additional `is bound` checks in the lower levels of code. I think it should be bottom to top. So, say, serializer creates a log instance as a deliverable. Then repo instance can merge serializer logs together, add own messages on top and pass above and so on.

But ! If we return log, how we return the actual result ? Options:
- via exception ! Exception is not coupled with the log logically - it is a transport. This is OK imho. But ! We cannot use it for info/warning. So does not fit for all cases
- via `return structure` - one piece is out result, another is a log instance. This is ugly.
- and, my favorite, via an interface. A class (e.g. repo) may implement `zif_loggable` which will expose `get_log` or `get_log_messages` (better imho) and which can be target for `log->merge`. So

``` abap
  create object global_log.
  create object x.
  some_stuff = x->run.

  global_log->merge_if_loggable( x ).
  " which checks if X is loggable and acquires the messages
```

clean and universal. Though will need to convert some static stuff into instances.
And thus the log will bubble up and aggregate.
And then it can be assigned to gui via `gui_service` and be accessible for all consumer who wants it.

And ! Exceptions ! They are still here. They are a transport of a log with errors. However, with `zif_loggable` it may give one potential option. It can be a log instance. Or ! It can be `me` of the caller. Which may implement `loggable`. Maybe `raise` can even get it from stack automatically ... (**can it ???**). 
Advantage: no direct coupling with `zif_log`, the reference can be to object. 
Disadvantage: we expose lower level instance to higher level, kind of insecure.
But potentially more elegant and automatic.

**Let's get back to Problem 1 and global var issue.** In this concept I think it is inavoidable to have log as a global var ... But i'm thinking how to reduce the scale of the disaster. And I think that maybe we should have not the log global var but some kind of `operation_context` - a bigger objects which would hold a log, an operation title (instead of the log which is a bad place in fact, especially if we consider merging), a repo (!) - the context is maybe a better place for the repo instead of passing instance between the components AND to make pages less dependent on creation parameters and make them more stateless (which could be good for web version eventually).
How this solves "log global var issue" - it does not directly, but it would be harder to forget to reinitialize the context properly that to reset a global log instance, there will be no "reset log", instead we will create new contexts.

Looking forward for comments and judgements
And sorry for the long read :)

mbtools:
Thanks for the detailed analysis. It is quite technical. I'm still kind of new here ;-) and we are discussing a complete re-design. So I'm asking myself... Why is AG doing logging differently than other ABAP applications? Why do we need a different logging framework? Why not standard SAP application logging? I have the feeling that we are reinventing something here.

Refined requirements:

- log is the result of *one or more operations*, negative or also positive
- not all operations can produce log, *but processes with many steps shall log operations*
- *no* operation should reset current log state. *the application (AG) shall be in control of opening, closing, and displaying the log*
- we want to decouple log from repo
- we want a proper systematic place/approach to keep log
- log may have a title (questionable, see below), *but start/end of operations must be identifiable*
- *it shall be possible to persist logs (via setting)* 
- *logging shall work for dialog and background processes*

No one likes `BAL_...` function calls but there are solutions hiding this complexity (like [ABAP-Logger](https://github.com/ABAP-Logger/ABAP-Logger/blob/main/docs/MISSION.md)). Everyone understand how it works and there's no need for "contexts", "interfaces", "passing log instances up or down", "is bound checks", "merging" etc. The process is simple: open log, any (!) operation can add messages, close log 

Let's use an example: "pull"

There are multiple operations: "serialize local files", "login remote", "get remote files", "calculate diff", "check deserialize", "user decisions", "deserialize objects", "activate objects"

Currently only "deserialize objects" is logged. Ideally however, the end-to-end process should be logged and optionally the log should be persisted (audit requirement of who did what when).

I will use abap-logger because it's so easy to understand and use. I'm not saying we should use this library but the pseudo code below should show how I see it work in principle. 

Problem 1: 

Processes might span multiple pages or components. So, yes, the global GUI class provides the logging service. 

```abap
CLASS-DATA: gi_log TYPE REF TO zif_logger.

class-method get_logger.
  ri_log = gi_log.
endmethod.
class-method set_logger.
  gi_log = ii_log.
endmethod.
```

Problem 2: 

When a process begins, it opens a new log. It's "from top" and it's in the UI layer (outside of "operations").

```abap
gui=>set_logger( zcl_logger_factory=>create_log( object = 'ABAPGIT' subobject = 'LOG' desc = 'Pull' ) ).
```

Operations just get the logger instance and add messages. No import or export of the log. This way we can add logging anywhere in AG. 

```abap
li_log = gui=>get_logger( ).

li_log->( 'Operation description...' ). 
loop at ...
  li_log->add( 'details' ). 
endloop.
```

Problem 3: 

Using an auto-save of the log (persist each message immediately) avoids having to close the log explicitly. So we can retrieve the current log anywhere: Directly at the end of the process, via a menu item later, etc. 

```abap
gui=>get_logger( )->show( ). 
```

As you wrote, GUI is already global. Therefore, I don't see the problem of a global log included with it.

Just my 3 cents.

sbcgua:
Reinventing - quite probably. But I would not compare to a regular abap app. 
First, a regular logging abap program is one operation, not an interactive tool. A global log for the whole execution would be fair for one-action-prog (for AG this scenarios is for background jobs only).
Secondly, I'm not sure abap follows "leading practices" in logging. E.g. what I described took some ideas from JS projects I know. I think it would be a better benchmark.

Having said this, I think we are talking about very close things. Just disagree on some details.

E.g. **top-to-bottom, bottom-to-top**. If there is a GUI service to provide log, then e.g. `object_deserializer` will know about GUI. Where is the GUI and where is the deserializer? This would be a very strong coupling. I think I have to commit some code to show what I mean with `loggable` idea.

**Context/no-context**: this is more subtle. The idea is that context is potentially less error prone, then global log. Intuitively. Maybe it should be in a separate PR ... because actually it is another subject.

```abap

  "@router
  "do_some_action

  context = gui_services->get_context( iv_create_new = abap_true ).
  context->set_title( 'do_some_action' ).
  context->set_repo( repo_srv=>get_repo( repo_id_extracted_from_the_event_params ) ).

  try.
    create object lo_deserializer.
    lo_deserializer->do_some_stuff( context->get_repo( ) ).
    context->get_log( )->merge_if_loggable( lo_deserializer ).
  catch zcx_error into lx_error.
  	" catching will be in GUI actually ... but just to demo the idea
  	context->get_log( )->merge_if_loggable( lx_error ).
  endtry.

  ...
  "@page
  context = gui_services->get_context( iv_create_new = abap_FALSE ).
  if context->get_log( )->count( ) > 0.
    ...
  endif.
  render_some_repo_stuff( context->get_repo( ) ).

```

**Autosave and persistence in general**. I think it is a good feature. At least for BG jobs for sure. Just for a separate PR. It should be easy to add `zcl_log_persistence` to convert the log class content to `BAL*`
framework.

larshp:
an operation can return a log, this is whats implemented right now IMHO

yea, some of the logs should maybe be returning instead of importing, but this is minor

the current approach I see as similarto BAPIs, which is an operation, and it returns a log, not really sure I understand what the design problem is with this?

larshp:
do programming to an interface, 

I don't like having global variables, where the caller needs to know the implementation of the action to know where to look for side effects. Eg. BAPIs have the log as part of the API

sbcgua:
Quick call summary:
- leave the `ii_log` params for now, make them mandatory
- remove `clear` method from the log class - to ensure log is not cleared occasionally at lower levels
- UI log goes to `gui_services` - `set/get_log` methods, but this is very error prone, so let's think about the `context` stuff in some future
- supposedly a new log will be created very early in pages or the router

One thing we forgot to discuss if we keep the exception as a transport for the log (the one with errors of course). If the log is mandatory param then maybe we don't need it ... on the other hand it may get handy at some point.

larshp:
I think its okay to have the logs possible to be transported via exceptions

larshp:
@sbcgua whats the status for this PR?

sbcgua:
Hi Lars, a bit busy at  the moment, cannot get my hands on it for some time. Does it block anything ?

larshp:
no, just wanted to figure out what the status for it is, can you convert it to draft, if its in progress?

sbcgua:
Closing for now. Cannot find time for this for last couple of month, sorry =( The next step was almost done in April - https://github.com/sbcgua/abapGit/tree/log2 but still needed some bug fixing.

PR #4599, opened 2021-03-03
Strip XML details user exit

sbcgua:
to #4540

Ready for review. Summary:
- Exit `PRE_CALCULATE_REPO_STATUS` was added
- Flag in local repo settings to enable/disable the behavior on repo basis (screen below in comments)
- The exit is called from `ZCL_ABAPGIT_FILE_STATUS->status` 
- ... and, **unfortunately**, from diff page, where it is called twice, because `status` is called there too. I think we need to separate status from the repo class. To solve this and to decouple code in general
- Some docs too & screenshot below
- user exit plugin - https://github.com/sbcgua/abapgit_xml_stripper_plugin - **I'd suggest including it into AG**, after review of course (there is a nice readme by the way - so the doc is ready)
- sample for the plugin - https://github.com/sbcgua/abapgit_xml_stripper_plugin_sample

Issues:
- appears that indentation done by sXML is different from how objects are serialized in AG (iXML??). And it gives a strange artifact. Which somehow does not impact the diffs ! (???)
2021-03-04_21h36_54

OFFTOPIC: @larshp: why did you use iXml in the frst place ? And not sXml ? The latter is faster. A room for improvement ? Or was there some specific reason ?


sbcgua:
https://github.com/sbcgua/abapgit_xml_stripper_plugin

sbcgua:
One more thing to do - adjust UI to be able to enable/disable stripping (business meaning: non need at origin system, but needed for deployment targets)

sbcgua:
Without stripper
image

With stripper
image

sbcgua:
Enabling the exit on repo basis
image

sbcgua:
ready for review, description updated

larshp:
suggest avoiding the new configuration option, pass `io_repo` to the exit, then it can have logic to determine if/when the logic is relevant to run

sbcgua:
This was my initial idea also. This logic exists. The exit searches a config file in remote files (`.xmlstrip.config` in the example) - if it doesn't find - it just returns. But there is one exceptional case (at least for my goals): I don't want the stripping to run on the original system because all the code there is supposed to be proper - stripping is for deployment only. So we need additional factor to switch it on and off. 
Alternative option is to save system id or hardware key to the config and analyze it in the exit. But it is rather indirect and also either not unique enough (sysid) or unprivate (hwkey).

larshp:
all other user exits are currently run always, the user exit should not make assumptions regarding the usecase IMHO, any logic can be implemented in the user exit

sbcgua:
Imho, we should ground this on functional requirement. Exit are additional functions after all. Could you suggest another solution for the issue I described ? Except not-installing the exit on the origin system

sbcgua:
Just thinking about alternatives and passing repo - this would allow exit to trigger repo methods which is "unsecure" ... it should be probably some other form ...

sbcgua:
After breakfast and conversation in salck: OK, I agree, user exists are flexible enough and system-based enough to implement it without settings. Will fix later today

sbcgua:
ready. Ok to review. 

P.S. some strange message from abaplint ... not related to my changes

larshp:
abaplint: due to a recent change in deps, let me fix it, sec

larshp (approved):
looks good to me

larshp:
thanks 👍

Issue #4622, opened 2021-03-10
Login language


nomssi:
While trying to upload an offline repository I got:

Current login language 'DE' does not match main language 'EN'. Select 'Advanced' > 'Open in Main Language' 

but I could not find "Open in Main Language" in the menu. I can vaguely recall it existing many iterations ago, or I am missing something.

best regards,
JNN

mbtools:
`ZABAPGIT` tcode will automatically exist for the dev version. If you are using the stand-alone version, you have to create a tcode for program `ZABAPGIT_STANDALONE` yourself.

nomssi:
Thank you all for the feedback.

PR #4627, opened 2021-03-11
Minor Fixes

### LINKED ISSUE #4625: TADIR - read_single, no exception thrown
https://github.com/abapGit/abapGit/blob/da43c34fc4f91ae8a9446f74d4df6ba7905085d5/src/objects/core/zcl_abapgit_tadir.clas.abap#L429-L438

This method has a `RAISING` in its signature  but does not throw an exception if the object is not found.
Either raise the exception when not found or remove exception from signature to make it clearer for the caller.

### LINKED ISSUE #4626: Search help popup - does not always return correct value
Example `TADIR-OBJECT`:
When selecting a value two values are returned:
image

Suggest to first try to read the exact fieldname, and if not found the first value instead of always selecting the first entry in the return table:
https://github.com/abapGit/abapGit/blob/d7f74bd49c28779fec09148851fffac064e14f82/src/ui/zcl_abapgit_popups.clas.abap#L891-L895

g-back:
closes #4626 - search help, first attempt field name
closes #4625 - remove exception from signature, remove obsolete comment

ref comment in #4622 - allow language switch when multiple tcodes are assigned to abapgit

### DIFF
--- a/src/objects/core/zcl_abapgit_tadir.clas.abap
+++ b/src/objects/core/zcl_abapgit_tadir.clas.abap
@@ -428,8 +428,6 @@ CLASS zcl_abapgit_tadir IMPLEMENTATION.
 
   METHOD zif_abapgit_tadir~read_single.
 
-* note that SICF has special handling, the obj_name is different in the system than serialized
-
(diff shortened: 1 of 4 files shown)


mbtools (approved):
👍

PR #4631, opened 2021-03-11
Update unit test to match changed message

pokrakam:


larshp:
thanks, I ran into this error just now,

image

PR #4646, opened 2021-03-16
Add measures to repo stats

mbtools:
- Lines in ABAP Files
- Lines of Code in ABAP Files (no empty or comment lines)

Method read_stats became to long and was split into smaller parts.

abapGit Example:

image

(Note: local counts are higher than remote because my fork has some new, uncommitted files)


larshp (approved):
👍

PR #4671, opened 2021-03-25
Fix namespace check (for DOCT objects)

### LINKED ISSUE #4670: Namespace does not exist message for non existing namespace
I get the warning "Namespace /ABC/#DEF/ does not exist. Create it in transaction SE03" (namespace anonymized). It's correct it doesn't exist but it also shouldn't as there are no objects in it nor is this a valid namespace.

The cause seems to be a DOCT object which is wrongly analyzed by the regular expression.

OBJ_TYPE: DOCT
OBJ_NAME: /ABC/DEF/DR_VERIFY
PATH: /src/#abc#something/
FILENAME: #abc#def#dr_verify.doct.xml

https://github.com/abapGit/abapGit/blob/8f0532d01576a46e93f2e1e7da7c062b56e417df/src/objects/core/zcl_abapgit_file_status.clas.abap#L497

mbtools:
Improve regex to avoid conflict with object names that contain multiple `/`

Closes #4670

### DIFF
--- a/src/objects/core/zcl_abapgit_file_status.clas.abap
+++ b/src/objects/core/zcl_abapgit_file_status.clas.abap
@@ -494,7 +494,7 @@ CLASS ZCL_ABAPGIT_FILE_STATUS IMPLEMENTATION.
 
     " Collect all namespaces based on name of xml-files
     LOOP AT it_results ASSIGNING <ls_result>.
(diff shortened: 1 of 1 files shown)


flaiker:
Thanks, that fixed my issue

PR #4676, opened 2021-03-26
Log refactoring, part 1

sbcgua:
cleaner part of #4590, ref #2821

this one contains just:
- new features in log
- excpetion as a log transport
- new gui service

the rest will follow separately


larshp (approved):
lgtm

PR #4677, opened 2021-03-27
Fix processing for files with `&nbsp;` 

christianguenter2:
#3073
`&nbsp;` (non-breaking space) in filenames is replaced in cl_http_utility=>unescape_url with `#`. Therefore we have to replace `&nbsp;` with space so abapGit is able to process file with spaces properly.


larshp:
does this PR also solve https://github.com/abapGit/abapGit/pull/4615 ?

christianguenter2:
Didn't see 4615, but I think so. 
@mkaesemann can you test whether this fix adresses your problem too. ?

christianguenter2:
#4613

mbtools:
we can merge this but should be clear that it's fixing the symptom, not the root cause. filenames should not contain any special characters (like nbsp) in the first place. when converting objects to filenames, such characters should be encoded (%XY).

christianguenter2:
nbsp is inserted here ZCL_ABAPGIT_GUI_PAGE_STAGE / render_file
image

mbtools:
this `replace` - and the reverse replace - are only necessary, since `is_file-filename` contains spaces. 

when the filename is created properly (with %20 instead of space), then the `replace` in `render_file` can be removed (and this PR is not necessary).

mbtools:
anyway, please merge this. we can get back to it another time

flaiker:
But the spaces would stay in the actual file name in the repo or would they also be replaced there with %20 ? I would expect them to stay there since space is a valid character in a filename unlike / .

christianguenter2:
Removing 
```
    REPLACE ALL OCCURRENCES OF ` ` IN lv_filename WITH '&nbsp;'.
```
doesn't work, cause multiple spaces are condensed into a single space automagically in the browser control. Leading to the same processing errors mentioned in the issues. Therefore it isn't that easy.
I try to come up with a follow PR separating visual representation a bit more from actual processing data solving the root cause.

mbtools:
also in the repo. 

the core problem is that we are not consistently escaping data that's included in HTML links and HTML forms. Therefore, we get problems when events (via URL parameters or POST) are returned and we try to parse the returned values. There have been several issues and attempts to correct this but it's just not done properly imho (see `ZCL_ABAPGIT_HTML_ACTION_UTILS->parse...`). 

so either we go through all of AG and make sure to add `escape` when ever we deal with filenames or we escape the filenames when they are created (which is done now - but not for spaces).

christianguenter2:
problem is that escaping spaces in file names as suggested here https://github.com/abapGit/abapGit/issues/3073#issuecomment-806977876 has visual implications like this:
image
Therefore my idea is to split the data a bit more between display data and properly escaped 'processing data' for HTML forms.

mbtools:
Suggest to
- Merge this
- Cancel #4615
- Close #4613

PR #4680, opened 2021-03-30
Add header for status column in repo view

mbtools:
image


sbcgua:
Nice feature but it seems to be non-interactive. Sorting doesn't work and sorting arrow does not appear on click

image

sbcgua:
Hmmm, the fact that it behaves differently is really confusing (judging by myself). Options:
- sorting by combination of status "AA", "DM" ...
- Making the header grey, so that it is clear that it is different

sbcgua (changes requested):
I suggest implementing one of the above.

sbcgua (approved):
It's good now. Nice feature by the way. I will use it, I have relevant scenarios :) Thanks !

PR #4686, opened 2021-04-07
ENHx: Improve error messages

mbtools:
Include reference to previous exception and exception texts in error messages for enhancement implementations (`ENHO`, `ENHC`)


larshp (approved):
LGTM 👍

PR #4705, opened 2021-04-15
New "Remote" tab in repo settings

mbtools:
In an effort to simplify the menu options, a new tab has been added to the repo settings which combines all options for switching the repository to a different source:

- Switch between "Online" and "Offline"
- Switch to a different Git URL
- Switch to a Branch, Tag, Commit, or Pull Request

You can switch multiple times within the dialog but the change only take effect when you select "Save Settings". 

You can use https://github.com/mbtools/TEST_SWITCH for testing, which has some branches, tags, and PRs. 

Note 1: Value help for URL history has not been implemented yet.
Note 2: Related menu items (and coding) have been removed yet.

image

#4171, #4163


larshp:
can the option make offline/make online be removed? and the corresponding popup dialog?

image

mbtools:
Yes. After merging this, the following can be removed
- Branch > Switch to PR (menu item)
- Advanced > Checkout commit (menu item)
- Advanced > Change Remote (menu item and popup)
- Advanced > Make Offline / Online (menu item and popup)

sbcgua:
Good idea and feature. :+1:

Did some clicking around (didn't review code deeply). Here are some thoughts.

1) Main. In terms of UX these functions, which are relatively frequent, are too deep now. I think there should be a dedicated command in menu to lead to this place. Also "Switch" does not seem very clear. Switch what? Maybe we can pick a better word? Imho "Remote" would be clearer (or Remotes). Or "Git remote" to be more precise.
2) repo url is preserved when switched offline and then online. Is it intended ? (I find it cool, but intuition also whispers that it is a source of bugs)
3) after switching to online the HEAD section options are not responsive. You must save first. This is UX contra intuitive - they are shown (moreover, they re-appear !) but not functional. 
4) That would make sense to have switch online/offline buttons somewhere near Type of repo field. But it is not supported by current form architecture. Hmm... certainly a separate concern though.

mbtools:
1. We have "Local" already, so "Remote" would work ("Git" doesn't apply to the "offline" option). Menu: "Advanced > Change Remote" to jump directly to the form? I would make the change together with the menu adjustments mentioned above.
2. Intention is to allow all changes within the form but the URL should be cleared on save for offline. 
3. Saving is necessary to change the repo instance. Otherwise, you won't be able to pick branches, etc. Either the head section appears and is functional or it should be hidden until you save. Will check
4. Yup

The code is in parts a bit clunky since abapgit.xml isn't structured well when it comes to these settings/fields, repo doesn't have an interface to handle such, and I didn't want to touch either. As always there's plenty room for refactoring...

sbcgua:
1. _""Git" doesn't apply to the "offline" option"_ - imho it does, we still have zip files. But I don't insist, "Remote" seems to be clear in the context. _"Menu: "Advanced > Change Remote""_ - yeap, good idea.
2. Mmm, no I meant different thing. But I cannot recreate it at the moment, maybe it was some occasion. One more point here, I'd say that "Make on/offline" should autosave. It looks and fills as a final command. But it's minor anyway.
3. _"Either the head section appears and is functional or it should be hidden until you save"_ - yeap. Exactly. If not possible to make it functional then ~~it means that the underlying logic is not decoupled enough and must be refactored and rewritten, :)~~ the form can show some kind of "save to choose the details" banner instead of those 4 fields
4. :+1: ... by the way this point correlates with (2). UX-wise the button at the bottom should act like a final command and if this button is within the form then extra save step is logical. Anyway this is a finetuning.

sbcgua:
One more thought: "Menu: "Advanced > Change Remote"" is OK, but maybe we can go straight to that place clicking the repo url. I don't think many people click it to go to repo source - it is a less frequent function. Switching repo is more frequent. Imho. Just an idea

mbtools:
- Changed "Switch" to "Remote"
- Added "autosave" for on/offline switch

larshp:
@sbcgua any additional comments?

sbcgua:
I would lightly suggest fixing also this point:

> after switching to online the HEAD section options are not responsive. You must save first. This is UX contra intuitive - they are shown (moreover, they re-appear !) but not functional ... Either the head section appears and is functional or it should be hidden until you save" ... If not possible to make it functional then the form can show some kind of "save to choose the details" banner instead of those 4 fields

both ways are OK, the essence is to not confuse people with reappread but not functional fields.

However not critical. We can also merge and improve this later, when adjusting menus

mbtools:
that was fixed already by adding the "autosave" for the on/offline switch

sbcgua:
Ah, ok, sorry. Yes, I see now. 

Then one more comment. I finally re catched (2) above. :)
So when I switch to Offline. The URL becomes Name.

image

image

However it is not immediately obvious. And if I just return to repo view I see the full URL as a repo name. Which is probably not ideal. What about converting the url to repo-name only ? ("abap-string-map" in the example) Should be one line to add.

mbtools:
it already had the "name" logic but it was overwritten by the save. I basically combined the "switch on/offline" and "save" events and improved the validation. 

- switch online > offline: 
changes URL to repo name and saves automatically 
- switch offline > online: 
shows message to "enter a URL and save"
- switch online > offline > online: 
goes back to the original/last valid URL and saves automatically

sbcgua (approved):
Cool :+1:
Nice feature

larshp:
@mbtools perhaps create a new issue for the cleanup?

PR #4710, opened 2021-04-18
Change some masters to main + remove unused parameter

pokrakam:


mbtools (approved):
👍 

changing c_git_branch might break some external dependencies but those would be easy to fix

#4043

pokrakam:
I did consider leaving `c_git_branch-master` alongside `c_git_branch-main` but in the end I thought lets just move forward, any breakages can benefit from updating.
Another argument was that "_we'll remove it later_" doesn't happen as often as we imagine
Anyhow, happy to put `master` back if anyone thinks it would be better.

PR #4717, opened 2021-04-24
Refactor PDTS into PDXX superclass

pokrakam:
In preparation for work on `PDWS`, shift common `PDTS`&`PDWS` elements into a `PDXX` superclass


mbtools (commented):
Compiles on 702. If experimental is OFF, just ignore the UT, no errors (same as #4683)

```abap
    IF is_experimental( ) = abap_false.
      "Alpha version, known issues:
      "- Container texts not de/serialized properly (functionally OK)
      "- Container handling still a bad hack, needs refactoring with lots of debugging time
      "- Probably has a few more bugs, more testing needed
      RETURN. "<<<
    ENDIF.
```

pokrakam:
> Compiles on 702. If experimental is OFF, just ignore the UT, no errors (same as #4683)
> 
Sure, though this is part of the original code, I tried to keep this a clean refactor of class hierarchy. Have added a quick fix to keep it green on all systems. 
I want to rework the unit testing part my plan is to do this under PDWS and retrofit it to PDTS.

mbtools (approved):
👍

Issue #4718, opened 2021-04-25
Clean-up after adding "Remote Settings"


mbtools:
The following items are replaced by the "Remote Settings" (#4705) and should be changed

- [x] Branch > Switch to PR
remove menu item
- [x] Advanced > Checkout commit
remove menu item
- [x] Advanced > Make Offline / Online
remove menu items and popups
- [x] Advanced > Change Remote
remove menu item and popup

- [x] To reach "Remote Settings" more easily, I suggest to add an "Edit" icon behind the URL

![image](https://user-images.githubusercontent.com/59966492/115986514-d1588a80-a5b0-11eb-9992-d89efb646fa6.png) ![image](https://user-images.githubusercontent.com/59966492/115986412-57280600-a5b0-11eb-960b-7002aeea4229.png)

sbcgua:
Good idea with edit icon

> Advanced > Change Remote: remove menu item and popup

Did you mean "Add item" ?

mbtools:
No. Do we still need the menu item if we have the one-click edit icon?

sbcgua:
That's about toolbar and menu duplication ...
But in this particular case, maybe it is OK to remove, because the remote setting **is** accessible from 2 places, so I think you're right, we can remove that item too.

g-back:
I can do the removal parts @mbtools - unless you've already started 🙂

PR #4723, opened 2021-04-26
Remote Settings: Toolbar icon

mbtools:
Continue #4718

- Add "Edit" icon behind URL to jump directly to "Remote Settings"
- Minor fix to remember original URL

image


mbtools:
Combined into #4724

PR #4729, opened 2021-04-28
docs: add missing point

schmelto:


FreHu:
I think the reason this can't be merged is because you didn't check the "allow edits from maintainers" option, so I don't have "update from main" available to sync it before merging. Please check the box or update from main yourself, I think you should see the button as the author.

FreHu (changes requested):
see comment above

schmelto:
I`ve opened a new PR for this #4769

PR #4731, opened 2021-04-29
Settings: Adjust background

mbtools:
Changes background to secondary color (same as repo top bar)

image


sbcgua (approved):
Looks good. I'd experiment with some shadow under the dialog, but that separate concern imho. Better than it was for sure

PR #4732, opened 2021-04-29
Add hover hightlight to repo toolbar

mbtools:
Highlights links in repo toolbar on hover

image

image

image


schmelto (approved):

sbcgua:
Hmmm, i'd look into changing default color. Hover is good, but the problem that these elements don't looks as active in the first place.

mbtools:
But why isn't the URL blue or the commit id? The favorite star is gray, too, and I don't think anyone has a problem using it. 

I think it's nice that these screen elements are gray and not "active": Good to have (for us) but generally not important or rarely used.

mbtools:
There's room for improvements (as always ;-). For PR, we should use the fork icon image. 

I do like to see the URL for the same reason you mentioned for the overview page.  But maybe we just mirror the GH design:
image
image

larshp:
merging

sbcgua:
good idea with fork/origin, just let's keep in mind the api request limit

Issue #4282, opened 2020-12-03
Combine "Pull" and "Reset local (forced pull)"


mbtools:
Current issues:
- "pull" automatically imports all objects that are changed only remotely (but not locally). There is no option to select which objects should be pulled.
- "pull" ignores all objects that exist only locally (but not remotely). There is no option to delete the local objects. However, "reset local" allows you to select if objects should be deleted. 
- "pull" and "reset local" let you select objects if they are changed on both sides or have been deleted locally. 
- "reset local" has a user confirmation popup, "pull" does not

Proposal:

Combine both features into "pull" and show **one** selection popup that lists **all** objects with a diff. The popup shall contain the following columns:
- checkbox
- object type
- object name
- current local state (same as in repo view)
- current remote state (same as in repo view)
- comment

The comment shall describe what happens if the object is selected and imported (plus a corresponding icon):
- "update local object" (selected by default)
- "delete local object" (not selected)
 
If the user selects objects to be deleted that exist only locally (so there's no remote backup), then the user shall confirm the deletion in an additional popup.

JanisB-AT:
Tagging #413.

I like the proposal very much, if for no other reason than because it provides rationale for doing the fairly extensive changes needed IMO to enable "selective pull" :) 

By "show one selection popup that lists all objects with a diff", did you mean that the popup should by displayed always, @mbtools? Another possibility were to let the "Pull" proceed with the default selection, and leave the selection popup to "Selective Pull", which could replace "force Pull" in the Advanced menu. The main reason I'm asking is: "selective Pull" would also need to include objects in the status "XA" (objects not existing locally) in the popup...

To enable optional deletion during pull were IMO fairly straightforward: starting point would be "moving" the deletion logic from ZCL_ABAPGIT_SERVICES_GIT->RESET( ) into  ZCL_ABAPGIT_SERVICES_REPO->GUI_DESERIALISE( ). There the popup could be merged with the two already there (overwrite and package overwrite). Because of "package overwrite" the local package would also need to appear on the popup.

Side question: is there any particular reason ZCL_ABAPGIT_POPUPS->POPUP_TO_SELECT_FROM_LIST( ) does not enable user to search and filter the list contents..? Asking because those would seem like obvious things to have, SALV has the functionality, but its not available on the UI (simply because of the choice of PF-STATUS, I suspect)...

It would IMO be nice to have two additional functions on the popup: select/deselect dependencies, but that may be taking it too far...

[...]

PR #4733, opened 2021-04-29
Enhance "Force Pull" to become "Selective Pull"

### LINKED ISSUE #413: Pull only selected files?
Hello,

Currently when I press PULL... all objects with differences are automatically overwritten.

Due to some issues probably related to SAP versions, when I stage my changes from source system A into target system B... even though all the code is properly sync'd, I still get eternal diffs, most times small irrelevant flags which I don't know how to change.

For example, most times the TOP include of a function group insists in showing this diff in the target system B:

Local: `<VARCL>X</VARCL>`
Remote: `<VARCL/>`

Sometimes a repository has several files which are ok but still show diffs. As a result, whenever I pull a repository in the target system... instead of just updating that small tweak... it unnecessarily rewrites function groups, classes, tables, etc.

Also, sometimes I have committed multiple changes in source system and in the target system just want to pull a specific one.

Given this, would it be feasible to implement a way to selectively pull only selected objects?

Thanks,
Nuno

mbtools:
With this PR you can now peform a selective pull. Yay!

- Changes "Force Pull" to show *all* changed objects in the usual "Overwrite Popup". This allows you to select individual objects to be pulled no matter if they were changed locally or remotely.
- Does *not* change the regular "Pull" command 

Closes #413, #4282


sbcgua:
Hmm, cool idea and look like a quick win for a low price. 

But doesn't work as intuitively expected ...
image

mbtools:
It remains a two step process. First popup about deletions. Second about overwrite (with all other changed objects). Combining both popups is a different story and bigger change.

src/objects/core/zcl_abapgit_objects_check.clas.abap
```diff
@@ -158,6 +177,33 @@ CLASS zcl_abapgit_objects_check IMPLEMENTATION.
...
 
 
+  METHOD warning_overwrite_find_all.
+
+    DATA: ls_overwrite LIKE LINE OF rt_overwrite.
+
+    FIELD-SYMBOLS: <ls_result> LIKE LINE OF it_results.
+
+    LOOP AT it_results ASSIGNING <ls_result> WHERE NOT obj_type IS INITIAL
+      AND NOT ( lstate IS INITIAL AND rstate IS INITIAL ).
```
larshp, line 187: I have trouble understanding this condition in relation to the IF statement below, suggest sacrificing the performance, which will make it easier to understand, and maintain as there will only be one condition instead of two that must match

```suggestion
    LOOP AT it_results ASSIGNING <ls_result> WHERE NOT obj_type IS INITIAL.
```

larshp: alternatively move everything to the loop if possible

src/objects/core/zcl_abapgit_objects_check.clas.abap
```diff
@@ -108,7 +123,11 @@ CLASS zcl_abapgit_objects_check IMPLEMENTATION.
 
 
 * make sure to get the current status, as something might have changed in the meanwhile
-    lt_overwrite = warning_overwrite_find( ct_results ).
+    IF iv_reset_all IS INITIAL.
+      lt_overwrite = warning_overwrite_find( ct_results ).
+    ELSE.
+      lt_overwrite = warning_overwrite_find_all( ct_results ).
+    ENDIF.
```
larshp, line 130: can we calculate a status column in `lt_overwrite` instead of having 2 sets of logic at this level?

and then in ZCL_ABAPGIT_SERVICES_REPO=>GUI_DESERIALIZE it can choose if some of the choices should be left out

larshp (changes requested):

src/objects/core/zcl_abapgit_objects_check.clas.abap
larshp: if `lt_overwrite` can contain everything, then perhaps it should be renamed

src/objects/core/zcl_abapgit_objects_check.clas.abap
mbtools: i will combine the warning methods and add field `local_change (abap_bool)`. if `local_change` is false, then the entry will not be relevant for a regular pull (because nothing was changed locally).

larshp: cool, I think that will look better, instead of the boolean input parameter

the boolean will still be part of the gui_deserialize I guess, but its okay IMHO

src/objects/core/zcl_abapgit_objects_check.clas.abap
mbtools: yes to status field (see below). renaming would quite a bit of work since `overwrite` is used a lot (vars, methods, ls_check, popups). i rather spend some time combining the `delete objects` popup with the `overwrite` so there's one selection of objects to be changed.

mbtools: maybe we add an "action" column to the popup that describes what happens to the object (delete, delete & recreate, overwrite local).

larshp: okay 👍

mbtools:
Idea for the combined popup when resetting (instead of "delete unnecessary" and "overwrite"). It shows all changed objects with action and description of what will happen if selected.

image

The problem comes when only part of an object is changed (like deleting a function module or testclasses). Then the "delete" action/description is not correct. Yes, the object will deleted but it will be recreated right after ("ugly" as mentioned in #1066). 

So is this really better than the separate popups?

larshp:
I think its better that the backend calculates the action this way, then the GUI can decide to display one or multiple popups, and to auto accept some of the actions

mbtools:
Ready... "force pull" is now combining all change in a single popup where individual objects can be selected. It also makes it more transparent what happens to objects and when objects are deleted and recreated (because an object part was removed remotely). 

image

Regular "pull" continues to work the same way (no deletions, no confirmation when adding objects or updating objects which have not been changed locally). Nevertheless, the popup also contains the new icon and description columns.

image

larshp:
two unit tests failing, which I guess is due to these changes,
image

src/zif_abapgit_definitions.intf.abap
```diff
@@ -354,6 +358,7 @@ INTERFACE zif_abapgit_definitions
            name   TYPE string,
            text   TYPE string,
            length TYPE lvc_outlen,
+           icon   TYPE abap_bool,
```
larshp, line 361: `icon TYPE abap_bool,`

this looks wrong, either the type is wrong, or change the field name, eg. "uses_icon" or something

larshp:
unit test failing, and comment above

overall: good, looks much better IMHO 👍

larshp:
@mbtools see comments above

FreHu:
What if there was only one pull button and the default pull behavior was a user setting?

larshp:
@FreHu I think we are heading there, consider this as one step on the way 😄

src/zif_abapgit_definitions.intf.abap
larshp: renamed to "show_icon" by @mbtools

larshp (approved):
LGTM 👍 unit tests passing

Issue #4734, opened 2021-04-29
Consider making abapgit use less than 100% of browser window width


FreHu:
I am trying to reduce visual clutter in the main page and came to the conclusion that there are only 3 _really_ important columns:
- name
- branch 
- actions
- (navigation to repo url can also be added to the online repo icon)

By hiding repo url and package in the detail section, I get this on 50% of a 1920px screen.
![image](https://user-images.githubusercontent.com/5097067/116566659-37be0f80-a907-11eb-930f-2d884743d1b7.png)

On fullscreen though, it looks strange
![image](https://user-images.githubusercontent.com/5097067/116567275-c29f0a00-a907-11eb-84cb-5814dd1add5e.png)

In general, abapgit puts way too much horizontal space between things and it becomes easier to find things if you shrink the window to 50%.

![image](https://user-images.githubusercontent.com/5097067/116567485-f4b06c00-a907-11eb-9c63-8668e3bc000e.png)
![image](https://user-images.githubusercontent.com/5097067/116567509-faa64d00-a907-11eb-990e-2968367c1e09.png)

![image](https://user-images.githubusercontent.com/5097067/116567609-13aefe00-a908-11eb-896d-d8ce0908e08f.png)
![image](https://user-images.githubusercontent.com/5097067/116567637-1ad60c00-a908-11eb-8941-7006d8b29780.png)

This makes me think that abapgit should be designed for some fixed width like 1024px instead of 100% of the screen width, except for some cases like side-by-side diff.

@sbcgua @mbtools  what do you think?

sbcgua:
- reduce visual clutter
  - I think package is useful
  - And URL is not, I'd get rid of it too ...
  - but it would be useful to see, if the repo is switched from the default url to PR. Needs experiments, because the straight forward solution (adding url part to the branch) would probaly look long and ugly. maybe. needs an experiment.

- abapgit should be designed for some fixed width like 1024px
  - hehe, I think we discussed it some time ago and I suggested the same ;) But getting back to constructive, in reality some dynamic way to switch thing could be useful. Ideally automatic (needs normal CSS support). Or maybe with user settings, hmm.
  - yeah, branch label looks lonely on the right, especially since commit is on the left. And the new edit button. And since it's color coded ... maybe it makes sense to move all to the left. maybe.
  - "Commit action" - agree, however it need some proper mechanism to attach menu items from lower level components. But conceptually I think it is the right direction. All commands are to the right, except stage and commit.

[...]

PR #4751, opened 2021-05-05
Change default width of abapgit

FreHu:
adresses part of #4734 
This changes the abapgit layout to a centered 1280px fixed width instead of taking up 100%, with some exceptions:

### Repo list will use full width with detail enabled
image
image

### Side-by-side Diff
image
image


mbtools:
i pulled the change and it looks good. my current screen is 1920px so it takes a bit to get used to. i will keep using it like this for a while before my final verdict.

mbtools:
The change looks and feels good! Some follow-ups:
- location of classic popups (see above)
- some wrapping on repo overview (I have quite long URLs, see discussion #4734)

those can be handled separately. approved from my side

mbtools:
yes, center. I think background settings is wider than the others. should all be same width.

FreHu:
All centered now
image
image

PR #4759, opened 2021-05-10
INTF: Fix check if interface exists

mbtools:
As mentioned in #4673, the existence check for interfaces is not correct and actually checking for classes. The PR will correct the check.


larshp:
CLAS subclass already have the check,
image

larshp (approved):
lgtm

PR #4760, opened 2021-05-10
Remove adjust_namespaces

mbtools:
Method is obsolete since namespace of object is already adjusted in `zcl_abapgit_filename_logic=>file_to_object`

https://github.com/abapGit/abapGit/blob/main/src/objects/core/zcl_abapgit_filename_logic.clas.abap#L55


larshp (approved):
lgtm

PR #4770, opened 2021-05-13
Bring supported object list up to date

pokrakam:
Set closed issues with object de/serialisers to `Yes`, update PDTS to remaining todo.


mbtools:
I was thinking about this update, too. thanks. 

I thought about adding a column with the reference to the merge that introduced the object type (or relevant issues discussing it) instead of removing it. This way you can find for example who "invented" it easily.

pokrakam:
Hmm, I see this as more effort than the value it adds. Just pop an object type into the GitHub search box and you get much more useful and relevant info.

mbtools:
The list is part of user docs. Agree, let's keep it simple.

mbtools (approved):
👍

PR #4771, opened 2021-05-13
Repo overview - rework action buttons + cosmetic changes

### LINKED ISSUE #4734: Consider making abapgit use less than 100% of browser window width

FreHu:
resolves #4734 

- Moved the action buttons to the table header menu to make more space in the table. 
- Repositories are now selectable by clicking, actions update based on the selected repo
- Selection is persisted when going back
- Keyboard navigation in the table through `<`, `>` and `Enter`. These shortcuts are js-only and currently a bit gimmicky as the focus is never where you need it for the keys to work. I'd like to extend this to other tables/diff in the future
- ellipsis for remote url if it is over 50 characters after removing `https` and `.git`

image

image


FreHu:
maybe even package before name

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -446,6 +443,29 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_OVER IMPLEMENTATION.
...
+    REPLACE FIRST OCCURRENCE OF 'https://' IN rv_shortened WITH ''.
+    REPLACE FIRST OCCURRENCE OF 'http://' IN rv_shortened WITH ''.
+    IF rv_shortened CP '*.git'.
+      lv_new_length = strlen( rv_shortened ) - 4.
+      rv_shortened  = rv_shortened(lv_new_length).
+    ENDIF.
+    FIND ALL OCCURRENCES OF '/' IN rv_shortened RESULTS lt_results.
+    IF sy-subrc = 0 AND lines( lt_results ) >= 2.
+      " host.domain/org_or_user/repo/...
+      READ TABLE lt_results INDEX 2 ASSIGNING <ls_match_result>.
```
pokrakam, line 462: Needs to cut at 3, otherwise we only see the gitHub user/organization name.
Several of my repos now only show URL of `github.com/pokrakam`

pokrakam:
Really like it. I think the column order is a more logical progression of information.

However it still doesn't solve the long URL, if we go for two levels we can still end up with 
image

Should we perhaps not just splice out the middle? Along the lines of: 
```ABAP
    lv_new_length = strlen( rv_shortened ).
    IF lv_new_length > 60.
      lv_new_length = lv_new_length - 5.
      rv_shortened = |{ rv_shortened(55) }...{ rv_shortened+lv_new_length }|.
    ENDIF.
```
This gives a more pleasing visual look:
image

I based 60 on the longest repos I could find in dotabap, I think two are over 60 chars.

Any opinions?

mbtools:
Why not the notation we are all used to?

image
(original)

image
(switched to a PR/fork)

This way you can jump to org and repo (and it's shorter)

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
FreHu: This was on purpose, so it shortens the column but you're still aware if it's switched to a fork, 

`github.com/abapgit/abapgit` and `github.com/frehu/abapgit` have the common part `abapgit`, so I wanted to cut that. Otherwise there is nothing to shorten as `https` and `.git` is already removed.

but I'm going to go with @mbtools' suggestion, I think this is the best variant yet

FreHu:
- Will all git hosts have the same org/user/repo url pattern though?

- I'm now even more convinced about package column first - usually the key columns are displayed first. The repo name is not a key, it's not even unique. Abapgit's key is just a sequential id, so it makes sense to not show that. But you can't have an abapgit repo without having a package first and you can only version a package once, so that is the most key column we have.

pokrakam:
I think the branch-styled notation is too platform-specific. Part of the issue is with URLs containing large hashes. If there's one thing I learnt on the DevOps course its that SAP looooves to use hashes all over the place in their Git tooling.

larshp:
https://github.com/abapGit/abapGit/blob/18c9b404f139266747313deb3249c4c8186e78d2/src/git_platform/zcl_abapgit_git_url.clas.abap#L58 tries to handle the different git hosts urls

mbtools:
Gitlab can have more than two levels. Still the last part is the repo, the part before that the org.

Random example: https://gitlab.com/eyeo/websites/web.adblockplus.org
image

FreHu:
Latest state:
image

Line break no longer happens on icon, but if you're unlucky it can break words on a single character which is not very nice
image

my attempts to get `overflow: hidden` or `text-overflow: ellipsis` to work on this table have been unsuccessful so far

FreHu:
Working on checkboxes
image

pokrakam:
Checkboxes so we can perform an action on multiple repos? 
Personally I find the last version much nicer and logical. Having lots of action links just adds visual clutter IMHO. If we want actions links then I think some kind of popup menu would be better.

mbtools:
Good progress. However, I'm for keeping the repo first, not the package. The primary column - unique or not - is the repo. It's the entry point to the core AG feature and what everyone clicks on most in this view. It's what everyone is used to. 

checkbox + actions as menu is a bit confusing if it doesn't work with multiple repos and doesn't differ between online and offline. maybe keep the actions at the end of each row but combine them into a dropdown.

FreHu:
Actions will be either disabled or hidden based on which repo is selected. Multiple selection will not be possible for now, but eventually I want to be able to syntax check multiple repositories at once.

FreHu:
It works!
https://user-images.githubusercontent.com/5097067/118395866-ec1c9d00-b64c-11eb-910f-1ecfd26247ad.MP4

Todo: 
- persistence when going back
- you can't currently unselect a repository, only change the selection
- themable background color for selected row

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -290,162 +308,206 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_REPO_OVER IMPLEMENTATION.
...
+      iv_act = |{ zif_abapgit_definitions=>c_action-repo_settings }{ lc_dummy_key }|
+      iv_class = |{ lc_action_class }| ).
+
+    ii_html->add( action_link( lv_settings_link ) ).
+
+    ii_html->add( |</div>| ).
   ENDMETHOD.
 
+  METHOD shorten_repo_url.
+    FIELD-SYMBOLS <ls_match_result> TYPE match_result.
```
mbtools, line 497: not used

mbtools:
need help testing?

FreHu:
sorry, i've been busy with other stuff. I'll get back to it soon

FreHu:
it's ready now, i updated the original pr description

mbtools:
I got around to test it. It looks nice and clean and for the most parts works well. Feedback:

- The write-protection icons are missing (https://github.com/abapGit/abapGit/pull/4745#issuecomment-831976065)

- Several of my URLs are cut off. Smaller package column, wider URL column (len 60), please

image

- If there are many repos, you scroll down to the one you want to work with. If you select it, the action links do change, but you have to scroll back to the top to use them. Sticky isn't supported for IE but it should work on Edge/Chrome. For some reason, it  wasn't working for me.

```css
table.db_tab th {
  text-align: left;
  padding: 0.5em;
  border-bottom: 1px solid;
  position: sticky; /* freeze table header */
  top: 0; /* required for the stickiness */
}
```

FreHu:
I fixed the display issues. Instead of the sticky header I plan to follow up with enabling keyboard shortcuts on the selected repository for action buttons as well.

mbtools (approved):
Thanks for the changes. Works well 👍

FreHu:
if you're okay with it i'll go ahead and merge

mbtools:
ok!

PR #4793, opened 2021-05-27
Checkout commit: Improve error message

christianguenter2:
Checking out commits is currently limited to branches with less than 99 commits. If this limit is reached a message appears saying 'No commits are available in this branch.'.
After this commit is applied the error message is 'Cannot find initial commit. Too many commits. Action not possible.'.

E.g. https://github.com/abapGit/CI.git

Before:
image

After:
image


mbtools (approved):
👍

PR #4794, opened 2021-05-27
Create branch: Add existence check to improve error message

christianguenter2:
If we try to create a new branch which already exists the following error appears:

Git protocol error: Failed to update ref
image

After this commit is applied, an existence check is performed before we try to create the branch. Hence the error message changes to:

Branch 'new_branch' already exists 
image


mbtools (approved):
👍

PR #4797, opened 2021-05-27
Started with BRF+ serialization and fixed some errors from the first attempt

SaschaWaechter:
#33 
Improved BRF+ serialization with some fixes. 
BRF+ system application is really difficult because there are many particularities which result in different behaviour. Standard interfaces for BRF+ (XML export and import) are really limited (e.g. handling of packages, locks, deletion,...)
My test case with a local test application worked. Unfortunately I have no bigger project with BRF+ at the moment.
If someone has more complex applications feel free to improve.


SaschaWaechter:
https://github.com/abapGit-tests/FDT0/pull/2

Improved test-repo for BRF+(FDT0)
Old test repo has some naming convention problems

mbtools:
Hi Sascha, thanks for working on this. I had a quick look and there are two bigger issues that need to be check/fix before this could go into abapGit:

1. Compatibility with release 702. Do all the tables and classes exist in 702? I can do that
2. Removing system/user/timestamp details from serialized object

Here's an example:

```xml
<FDTNS:CreationUser>DEVELOPER</FDTNS:CreationUser>
<FDTNS:CreationTimestamp>20210526163440</FDTNS:CreationTimestamp>
```

These fields must be removed/empty or any attempt to import the object somewhere else will lead to diffs. At deserialze, these fields should be filled again, either by your code or by the classes you call.

mbtools:
All of BRF+ classes you used exist also in 702. But there are a few method parameters that don't exist.

image

```abap
          cl_fdt_delete_handling=>delete_logical_via_job(
            EXPORTING
                its_application_selection         = lt_application_sel
                iv_retention_time = 0
                iv_background = abap_true
                iv_local_option = '1'
                iv_appl_transported_option = '2'
                iv_obj_transported_option = '2'
                iv_marked_option = '3'  "<<< does not exist in 702
                is_object_category_sel = ls_object_category_sel
            IMPORTING
                ev_failure = lv_failure ).
```

```abap
          lo_dexc->import_xml(
            EXPORTING
              io_dom_tree              = lo_dom_tree
              iv_create                = lv_create
              iv_activate              = abap_true
              iv_simulate              = abap_false
              iv_import_type           = if_fdt_data_exchange=>gc_xml_import_type_standard
              iv_import_scope          = if_fdt_data_exchange=>gc_xml_import_scope_normal  "<<< does not exist in 702
            IMPORTING
              et_message = lt_message   ).
```

```abap
        lo_dexc->export_xml_application(
          EXPORTING
            iv_application_id = lv_application_id
            iv_schema  = if_fdt_data_exchange=>gc_xml_schema_type_external
            iv_xml_version = if_fdt_data_exchange=>gc_xml_version
            iv_incl_deleted = abap_false  "<<< does not exist in 702
          IMPORTING
            ev_string         = lv_xml_fdt0_application ).
```

We typically resolve this with adding a try-catch and a dynamic call like here:

https://github.com/abapGit/abapGit/blob/main/src/objects/zcl_abapgit_object_devc.clas.abap#L194

SaschaWaechter:
Thanks for the fast feedback will look at this. BRF+-API was released fortunately with 702 as far as I know but it's relly static so I will try to filter the useless metadata via coding. Did so already to handle packages.
Also will look at the method calls.

mbtools:
It's missing the pretty print. View the class in source code mode and pp it.

larshp:
method `remove_child` method added in https://github.com/abaplint/deps/pull/67

src/objects/zcl_abapgit_object_fdt0.clas.abap
```diff
@@ -0,0 +1,425 @@
...
+
+          cl_fdt_delete_handling=>delete_logical_via_job(
+          EXPORTING
+                its_application_selection         = lt_application_sel
+                iv_retention_time = 0
+                iv_background = abap_true
+                iv_local_option = '2'
+                iv_appl_transported_option = '1'
+                iv_obj_transported_option = '1'
+                iv_marked_option = '3'
```
mbtools, line 208: iv_marked_option not in 702

mbtools (changes requested):
- one small change for 702
- maybe indent/align some of the parameters for method calls

mbtools:
I tried pulling https://github.com/abapGit-tests/FDT0 in my 750 but am getting the following error:

image

I see that the test case was serialized on 751 which has a newer BRF+ version. Can we get a test case that would work on lower releases (in addition to FDT0)?

src/objects/zcl_abapgit_object_fdt0.clas.abap
```diff
@@ -0,0 +1,425 @@
...
+    SELECT SINGLE application_id FROM fdt_admn_0000s INTO rv_application_id
+            WHERE object_type = 'AP'
+            AND name = ms_item-obj_name.
+
+  ENDMETHOD.
+
+
+  METHOD wb_request_choice.
+
+    "Transport Request
```
mbtools, line 119: AG sets the transport automatically. You can get it via

```abap
lv_transport = zcl_abapgit_default_transport=>get_instance( )->get( )-ordernum.
```

mbtools (changes requested):

SaschaWaechter:
I have no system lower than 7.40 for me but i changed XML-Version manually to 1.11. Also the test-repo had a problem with the application name (on lower release it didn't have to start with a namespace e.g. Z).
BRF+ is really strange sometimes....
=> https://github.com/abapGit-tests/FDT0/pull/2

pokrakam:
I think the exporting system's XML version should stay current. Otherwise exporting from 7.5x and re-importing may result in the loss of all components that are features of newer versions.

Therefore I would suggest that we stick to the current version, even if it means that 7.5 rules cannot be imported into 7.4. We can use 7.02 to test code compatibility. By all means hardcode 1.11 for creating the test repo but remove for productive use.

pokrakam:
@mbtools you misunderstood, we were talking about the BRFplus version. 
`<FDTNS:Fdt xmlns:FDTNS="http://sap.com/fdt/transport" AP="OFF" BRFplusInitialVersion="1.13 " BRFplusSchema="http://sap.corp/fdt/External.xsd" BRFplusVersion="1.11 "...`

But since you mention it, why should lower version CI tests not be necessary? IMHO if we support a version then CI must run on that version.

mbtools:
We're on the same page. I also meant the brf xml version. 

The CI test prg use new ABAP and won't work on lower releases (unless someone volunteers to downport it). But it's more important to get it to work properly, run it regularly, and publish results. I'm on that now.

pokrakam:
Yes have also been tinkering to try to get CI running again. Got as far as a partial push to my own clone, need one of them time machine thingies, pesky customers keep taking up all my time.

mbtools (approved):

mbtools:
Merging this one and following up with PR to complete the work

mbtools:
@SaschaWaechter see #4856

PR #4801, opened 2021-05-28
Improve secondary key access

christianguenter2:
Fixes theses syntax check warnings
image


larshp (approved):
Looks good to me, thanks 👍

PR #4802, opened 2021-05-30
Enable input fields (submit and text) for link hints

christianguenter2:
Before
image

After
image


mbtools:
Add `input[type=number]` should be easy:

image

I'm not sure if it should work for `textarea` too

christianguenter2:
That's right. Good suggestion. It's now possible to reach all input fields and textareas. Plus checkboxes are automatically checked/unchecked and submit buttons are clicked.
image

mbtools (approved):
Nice 👍

PR #4812, opened 2021-06-01
Styling - Part 2

mbtools:
#4810
$ABAPGIT_BACKGROUND


larshp (changes requested):
see discussion in #4810

Issue #4816, opened 2021-06-01
Focus html control on abapgit startup


FreHu:
When abapgit starts, the focus is here:
![image](https://user-images.githubusercontent.com/5097067/120357981-4e8cc300-c306-11eb-820c-954017ecd923.png)

I'd like to focus the html control on startup/page load so that you can press keyboard shortcuts immediately, is it possible?

mbtools:
+1 
I'm not sure. The problem is that there are no fields on the dynpro 1001. Therefore, it defaults to the command field. 
https://help.sap.com/doc/abapdocu_751_index_htm/7.51/en-us/abapset_cursor_dynpro.htm

FreHu:
I found the methods
```
cl_gui_control=>set_focus( ).
cl_gui_control=>get_focus( ).
```
but they don't seem to work here. I tried to focus the html viewer as well as the screen0 container, but nothing happens. 
Calling get_focus immediately after set_focus will dump on a `cntl_error`

mbtools:
Correct. I got it to work and will put it in a PR for testing.

mbtools:
Try #4817. Needs some testing with Web GUI, Java GUI so we don't crash those.

FreHu:
In windows sapgui it works. I noticed that on pages where there is a filter (repo list/stage), it is now focused automatically

mbtools:
doesn't focus with forms like settings but I think that's an HTML thing.

FreHu:
if the browser has focus now we should be able to focus the right element in javascript on load

mbtools:
no js necessary. what's missing is the `autofocus` attribute for the first form field. we can add this in `zcl_abapgit_html_form` after #4771 is merged

PR #4817, opened 2021-06-01
Focus HTML control on abapGit startup

mbtools:
#4816

`cl_gui_control=>set_focus` must be called during PBO. Maybe there's a nicer way to pass the HTML control to the PBO or the other way around: expose a method in the viewer that can be called from the PBO.


jfilak:
FYI: @FreHu pulled it to one of our decentralized systems and I tested it with JavaGU. The result: there is no widget with focus (all keyboard events are completely ignored). Without this PR, the focus is in the command/transaction field of JavaGUI.

FreHu:
In js, focus can be set on document load, for example `document.querySelector("a[href='sapevent:repo_newonline']").focus()`.
In javagui this has no effect either.

mbtools:
I tested also with Web GUI and it works well. Good to merge from my pov.

mbtools:
For WebGUI and Chromium Control, I think js is a feasible solution to move the initial focus (see @FreHu above).

christianguenter2:
We already have a logic in place which tries to focus the filter input field on startup. But it doesn't work either.
https://github.com/abapGit/abapGit/blob/0344f7e360a66ff16920f00b95d7ac31c0c6bace/src/ui/zcl_abapgit_gui_page_repo_over.clas.abap#L270
https://github.com/abapGit/abapGit/blob/a92059d059aec3637443fb8aad559d6f81ef5fd5/src/ui/zabapgit_js_common.w3mi.data.js#L141-L143

FreHu:
@christianguenter2 based on my testing the js should work now

christianguenter2:
Hm, no. Not for me with chromium based browser control.
I have to press tab once then 'New online' is focussed and I can use hotkeys.
abapGit

FreHu:
On IE control:
https://user-images.githubusercontent.com/5097067/121029208-b4ff5e80-c7a8-11eb-91ef-856134457c6d.MP4

christianguenter2:
Yes, it works for me on IE control too.

christianguenter2:
Can be merged from my point of view. Just want to let you know that there's different behaviour with Chromium.

mbtools:
Here are my tests to check if cursor is in "Filter" field after startup. 

GUI | Test
-----|-----
SAP GUI 7.60 patch 11 (IE control) | Works consistently
SAP GUI 7.70 patch 1 + IE control | Works consistently 1)
SAP GUI 7.70 patch 1 + Chrome control | Works sporadically 2)
SAP GUI 7.70 rev 1 (Java) | Does not work 3)
Web GUI (SAP Kernel 7.53 patch 819) + Chrome | Works consistently

1) Cursor is in filter field but strangely not blinking
2) Often is does not work. But sometimes it does: Put cursor into filter field, then exit, the start again. If you repeat it, the cursor will sometimes be in the field as expected. Imho, that's an SAP or Microsoft (WebView2) bug.
3) It seems, the HTML control does not gets the focus at all and isn't even part of the tab-sequence. AG only works properly, if you click into the control.

PS: I added a line of JS right after the focus statement to turn the field red to be sure it actually executed and the JS runs in all cases.

image

christianguenter2:
Good news. With SAPGUI 7.70 for Windows SP Patch 2 and Chromium Browser Control 'New Online' is focused on startup. It works consistently for me.
image

mbtools:
That's better but JS should have moved the focus to the filter field.

PR #4827, opened 2021-06-07
Adjust documentation for mobile devices

### LINKED ISSUE #4824: Docs: bad layout on mobile browsers
Documentation is rendered in a bad way on Safari and Chrome on iPhone 7 (portrait as well as landscape):

image

image

mbtools:
I tested the change with Google Dev Tools and docs.abapgit.org should work now also on devices with smaller screens.

Closes #4824


larshp:
https://docs.abapgit.org/development/docs.html, tested ok,

image

image

image

Issue #4841, opened 2021-06-12
SAP GUI for Windows: Issues with chromium-based browser control (WebView/Edge)


christianguenter2:
Collection of issues as discussed in the slack channel and PR  #4840
These errors only occur with SAPGUI 7.70 for windows PL2 and chromium based browser control. None of the occur with IE control or with previous patch level.

- [x] "Unknown action: SELECT" -> #4840
- [x] Back navigation doesn't work -> https://github.com/abapGit/abapGit/pull/4840#issuecomment-860035017
- [ ] Patch doesn't work - add file and click 'Patch'
![image](https://user-images.githubusercontent.com/17437789/121778138-ece61780-cb95-11eb-9af3-a2097200ea6c.png)
- [x] Diff doesn't work -> Shortdump
![image](https://user-images.githubusercontent.com/17437789/121778192-2c146880-cb96-11eb-8b79-69a98ef6d585.png)
![image](https://user-images.githubusercontent.com/17437789/121778195-2fa7ef80-cb96-11eb-9a1c-6624d8358431.png)
Filename is UPPERCASE
![image](https://user-images.githubusercontent.com/17437789/121778196-320a4980-cb96-11eb-8f8f-1c68c8e8a413.png)
![image](https://user-images.githubusercontent.com/17437789/121778200-346ca380-cb96-11eb-95af-c2a6527dbcb4.png)

- [ ] Stage page: 'Change by (me)' doesn't work
![image](https://user-images.githubusercontent.com/17437789/121778392-ff148580-cb96-11eb-88d4-dd884c018fe3.png)
![image](https://user-images.githubusercontent.com/17437789/121778389-fc199500-cb96-11eb-821f-ffba06952d2f.png)

- [x] Bad performance. Every navigation step/action takes ~2 seconds

jrodriguez-rc:
I just test the PL03 and it seems that all the problems (or at least almost) are fixed.

@christianguenter2 can you test it with the last PL03?

christianguenter2:
Thanks for the info. I just did a quick test and some issues are fixed (marked above). Few still exist. 
@jrodriguez-rc Is your OCC call still open? If yes can forward that to SAP? Many thanks for your effort.

jrodriguez-rc:
The issue reported to SAP was closed by inactivity, I was waiting for this patch to test and report feedback.

mkaesemann:
Small update: Staginging issue/filtering issue is still there in 7.70 PL5 as well.

[...]

PR #4842, opened 2021-06-12
SAPGUI 770 patch 2 chromium control - several fixes

christianguenter2:
Fixes several issues from #4841 which occur with SAPGUI 770 patch 2 and chromium browser control

- convert actions always to lower case
- back navigation fixed as URL in SAPGUI browser control has changed
- JS logic for determining SapEvent-prefix matching new SAPGUI browser control logic
- Diff view: Convert path and filename always to lower case. As chromium control sends them newly in uppercase.

Tested with SAPGUI 770 patch 2 chromium and IE. Needs thorough testing with different SAPGUI versions and configurations.


jrodriguez-rc:
There is the same problem with `getdata`, it is always uppercase
image

I think that the best solution is to report an incident to SAP, I was testing report `saphtml_events_demo`.

With IE control:
image

With Edge control:
image

mbtools:
Agree. These are bugs that SAP must fix:
- Changing the prefix from `saphtml:` to `http:`
- Uppercase of action and getdata responses

christianguenter2:
@jrodriguez-rc already any news from SAP?

jrodriguez-rc:
No news, still in "Sent to SAP" status

jrodriguez-rc:
Changed to "In processing by SAP"

jrodriguez-rc:
Update: SAP sent me a patched version of GUI, the event and getdata are now with the correct case, but I reported more issues that I haven't reported initially:
- URL changes
Internet Explorer (and Edge 7.70 PL1):
saphtmlp://htmlviewer.sap.com/0b3WPqjo7jkqrVigXcMsk0/HTML000002.htm
Edge 7.70 PL2
https://saphtmlp//htmlviewer.sap.com/0b3WPqjo7jkqrWTp6%7BmskW/HTML000002.htm
- Delay in each event

jrodriguez-rc:
Update: A temporary fix will be provided (if I understood correctly), and they are working with Microsoft for a permanent fix.

Note: Correction is not provided in PL02 hotfix 1

jrodriguez-rc:
Update: I tested a binary with the fix for actions/navigation and it seems to work.

The problem with the delay is still pending (I just tested a binary for Edge HTML control).

It is registered in note [2901278 - SAP GUI HTML Control based on Chromium Edge: Legacy HTML does not work (correctly) / present limitations](https://launchpad.support.sap.com/#/notes/2901278), under "Performance when invoking the HTML Control" section.

larshp:
guess we are waiting for feedback regarding the note?

jrodriguez-rc:
The problem is not from abapGit, so I think that this PR will be closed and continue with issue #4841

christianguenter2:
Closing. Patchlevel 3 is expected to ship on August 19. Let's see whether there is any progress.

PR #4846, opened 2021-06-14
Add more keyboard shortcuts for main page

### LINKED ISSUE #4844: Repo list - keyboard shortcuts for action buttons
after #4771 

Shortcuts can work based on selected repo, just have to make sure if the action is not allowed the key does not work.
image

FreHu:
resolves #4844 

Adds keyboard shortcuts for check, stage, and patch which work based on the selected repo. List navigation shortcuts are now documented.
image

Trying to stage/patch an offline repo will show an error message
image


mbtools:
Nice. Wouldn't "up" and "down" cursor keys make more sense for prev/next?

FreHu:
I tried them initially but they didn't work, I think they interfere with browser scrolling.

mbtools:
I see. Up/down is better left for scrolling. 

There is `KeyNavigation` in the JS for something else which seems to be using up/down when the vinum links are on but I didn't get this working.

The bigger issue is that it's actually the `,` and `.` keys, not `shift + ,` which is `<`. This makes sense for US but it becomes very confusing on other keyboards. In France, `.` is actually `shift + ;` and `<` is a separate key. Similarly in German.

image
 
image

FreHu:
You're saying the tooltips are misleading, not that the keys to press are inconsistent, right?

mbtools:
yes. using `.` and `,` works but might not be as simple/logical on non-US keyboards.

larshp:
whats the status for this PR?

src/ui/zcl_abapgit_gui_page_main.clas.abap
```diff
@@ -164,5 +164,36 @@ CLASS zcl_abapgit_gui_page_main IMPLEMENTATION.
...
+
+    ls_hotkey_action-description = |Patch|.
+    ls_hotkey_action-action = zif_abapgit_definitions=>c_action-go_patch.
+    ls_hotkey_action-hotkey = |p|.
+    INSERT ls_hotkey_action INTO TABLE rt_hotkey_actions.
+
+    " registered/handled in js
+    ls_hotkey_action-description = |Previous repository|.
+    ls_hotkey_action-action = `#`.
+    ls_hotkey_action-hotkey = |<|.
```
mbtools, line 185: either change this to `,` or the JS to work with `<`

src/ui/zcl_abapgit_gui_page_main.clas.abap
```diff
@@ -164,5 +164,36 @@ CLASS zcl_abapgit_gui_page_main IMPLEMENTATION.
...
+    INSERT ls_hotkey_action INTO TABLE rt_hotkey_actions.
+
+    ls_hotkey_action-description = |Next repository|.
+    ls_hotkey_action-action = `##`.
+    ls_hotkey_action-hotkey = |>|.
```
mbtools, line 190: as above

FreHu:
In the end I changed it to 4 and 6 (both numpad and number keys work).

2 and 8 would be better indication of up/down but without a numpad they are far away. Num2 is also not supported by my particular keyboard.

mbtools:
Works well. Thanks 👍

PR #4848, opened 2021-06-14
Refactor method in FM call for 702

### LINKED ISSUE #4847: Function call not compatible with 702
Cannot use functional methods in CALL FUNCTION:

https://github.com/abapGit/abapGit/blob/887e7dd3321dfd913127a2d7e035f18e9225e80f/src/ui/zcl_abapgit_gui_router.clas.abap#L733-L738

pokrakam:
Closes #4847


mbtools (approved):
👍

PR #4851, opened 2021-06-15
WDCA: Clear parameter descriptions

mbtools:
Fixes diffs because of parameters that are different between releases or languages (or because SAP change it)

Example:

image


larshp:
but I guess they are part of the development? the developer entered these?

mbtools:
these parameters and descriptions are shipped by SAP in `WDY_APP_PROP_DEF` / `WDY_APP_PROP_DFT`

image

we could probably remove more like the abap type and is_changable but the descriptions is really causing the diffs.

larshp:
ahh, okay 👍

PR #4914, opened 2021-08-27
Docs: Expand constructor description

mbtools:
Describe logic to be used to check requirements and determine if an object type is supported or not.


larshp (approved):
perfect 👍

PR #4921, opened 2021-08-29
Fix "Cannot switch origin twice" in remote settings

### LINKED ISSUE #4919: "Cannot switch origin twice" in remote settings is confusing
Note: this is unrelated to the PR in the screenshots, I can reproduce with any of the open PRs.

Switching to PR works just fine the first time:
image
If I click save again, I will get an error even though I didn't change anything
image
This also causes the repo page to not work
image

The trick in both cases is to switch the branch back to main.
image

This is confusing, I know you can only have one kind of remote checked out at once, but 
- the remote settings shouldn't do anything if the form didn't change
- I don't understand why the second interaction has different results

mbtools:
resolves #4919


FreHu (approved):
👍

PR #4944, opened 2021-09-10
Improvements for Objects Class

mbtools:
This introduces a new `is_type_supported` method for `zcl_abapgit_objects` which is about 30x faster than the current `is_supported` method.

`is_type_supported` is used in the other methods like `changed_by` and `exists` to avoid instanciating or calling the object class unnecessarily. It won't be necessary to check for support before using these methods which will alllow simplifying the calling programs (separate PR).

There are also some new test cases for `is_type_supported` and for serializing a commom interface.


larshp (approved):
👍

PR #4973, opened 2021-09-26
FUGR: Clear TEXT when referring to dictionary to avoid unnecessary diffs when data element text changes

yellappam:
In a Dynpro screen, when a field is referring to a data element and text is taken from dictionary, no need to serialize the text as it is always taken from dictionary. If the text is included then it will show unnecessary diffs when data element text changes. This is observed during S/4HANA migration. 

It is tested with custom data element  which is used in a table for which table maintenance is created. After table maintenance is generated, data element text is changed and table maintenance shows changed text automatically.

**Data Element before Change**:

image

**Table:**

image

Table Maintenance: For all values except 'F', text is always taken from Dictionary.

image

image

image

**Before changes:**

image

**After Changes:** 

Data element text changed from 'Material' to 'Article'

image

abapGit shows diffs for data element which is expected and no changes against the table maintenance function group. Also, SM30 shows the changed text i.e Article

image

image

The above is tested for all values of 1, 2, 3, 4 and V which shows correct text from dictionary.

### DIFF
--- a/src/objects/zcl_abapgit_objects_program.clas.abap
+++ b/src/objects/zcl_abapgit_objects_program.clas.abap
@@ -831,6 +831,11 @@ CLASS zcl_abapgit_objects_program IMPLEMENTATION.
             CLEAR <ls_field>-foreignkey.
           ENDIF.
         ENDIF.
+
+        IF <ls_field>-from_dict = abap_true AND
+           <ls_field>-modific   <> 'F'.
+          CLEAR <ls_field>-text.
+        ENDIF.
       ENDLOOP.
 
       LOOP AT lt_containers ASSIGNING <ls_container>.


mbtools:
I think value `X` is also non-DDIC but I'm not sure if this exists in combination with the ddic flag being true. 

Doesn't the FUGR diff go away after you generate the maintenance a screen and refreshing AG?

yellappam:
> on-DDIC but I'm not sure if this exists in combination with the ddic flag being true.

Currently, 'X' is disabled hence, I didn't consider it in the logic.
The scenario is, there is no change to the table maintenance and there is a change to one of the standard data element text. In this case AG shows diffs which needs to be staged due to data element text change. As the text is not used by the AG, in the case taken from dictionary, we can clear this field to avoid the diffs.

mbtools (approved):
👍

PR #4986, opened 2021-09-29
Diff-page: enable refresh functions

### LINKED ISSUE #4979: Refresh on Diff page
There are various "refresh" options on the patching page:

image

I suggest adding those options to the diff page as well. 

Since `ZCL_ABAPGIT_GUI_PAGE_PATCH` is a sub-class of `ZCL_ABAPGIT_GUI_PAGE_DIFF`, this looks to be an easy change.

christianguenter2:
fixes #4979 

First working version.
Needs more polishing + refactoring. 
How can we enable hotkeys for both Diff and Patch? Interface zif_abapgit_gui_hotkeys cannot be implemented in both classes nor redefined as its method is a class method.

image

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -1564,7 +1564,7 @@ function Hotkeys(oKeyMap){
       }
 
       // Or a global function
-      if (window[action]) {
+      if (window[action] && typeof(window[action]) === "function") {
         window[action].call(this);
         return;
       }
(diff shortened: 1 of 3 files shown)


mbtools:
You should be able to remove `ZIF_ABAPGIT_GUI_HOTKEYS` from `ZCL_ABAPGIT_GUI_PAGE_PATCH` (save the `GET_HOTKEY_ACTIONS` code), add it to `ZCL_ABAPGIT_GUI_PAGE_DIFF`, then redefine it in `PATCH`.

just a thought: maybe `Refresh` on the top should be a drop-down menu with `Local` and `Remote` as an option

christianguenter2:
Yeah, no I can't. Static methods cannot be redifined.

mbtools:
you're of course right. the hotkey logic seems to be a bit overengineered. shouldn't it be more like the asset manager? a page registers the hotkeys with the hotkey manager. and at render time, we get the registered keys from the manager.

mbtools:
Check #4990. `get_hotkey_actions` changed to instance method

christianguenter2:
Will reopen after #4990 is merged.

PR #4988, opened 2021-09-30
Delete child number from SICF node

jrodriguez-rc:
Clear the calculated field `ICFCHILDNO` to avoid differences between systems
image

This field just calculate all child nodes for the node
image


mbtools (approved):
👍

PR #4990, opened 2021-09-30
Refactor hotkey controller

mbtools:
- Removed "class analysis logic" from hotkey collection
- Hotkeys are now registered directly in the hotkey controller (similar to the asset manager)
- `zif_abapgit_hotkeys->get_hotkey_actions` was changed to instance method and can therefore be redefined in subclasses
- Renamed the controller class to match the interface name

Ref #3286


christianguenter2 (approved):
Did some tests. LGTM.

PR #4992, opened 2021-10-01
Diff-page: enable refresh functions

### LINKED ISSUE #4979: Refresh on Diff page

christianguenter2:
follow up #4986 
fixes #4979

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -1564,7 +1564,7 @@ function Hotkeys(oKeyMap){
       }
 
       // Or a global function
-      if (window[action]) {
+      if (window[action] && typeof(window[action]) === "function") {
         window[action].call(this);
         return;
       }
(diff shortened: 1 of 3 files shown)


sbcgua:
Nice feature. Is it finished ? I see that `METHOD modify_files_before_diff_calc` is empty ...
Also I'd probably change Refresh to Refresh All. Because it it became less obvious what it does. Or maybe make Refresh a dropdown ?

image

christianguenter2:
Yeah, from my pov it's more or less finished. But there's some potential for polishing and refactoring, no doubt about that.
`METHOD modify_files_before_diff_calc` is redefined in subclass `zcl_abapgit_gui_page_patch`.

FreHu:
Works for me. I noticed a minor issue that after the refresh, a newly created class will not be available in the jump menu or `F2` search. But it's not a huge deal.

christianguenter2:
@sbcgua did go for option one, renaming 'Refresh' to 'Refresh All' on both pages.
image

PR #4997, opened 2021-10-01
Only show favorites on startup

FreHu:
Followup to https://github.com/abapGit/abapGit/pull/4541 where I mention

> Other ideas:
> 
> the favorites logic should be server-side and not a client side toggle, then we can get away with only loading the favorite repositories on startup
> 

- Client-side toggle changed to server-side
- If there are no repositories favorited, everything is loaded
- If there are no repositories on the system, tutorial is opened


src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -539,70 +555,70 @@ CLASS zcl_abapgit_gui_page_repo_over IMPLEMENTATION.
...
+      iv_tech_name      = 'KEY'
+      iv_display_name   = 'Key'
+      iv_css_class      = 'ro-detail'
       iv_allow_order_by = abap_true ).
 
     _add_column(
-      iv_tech_name = 'GO'
-      iv_css_class = 'ro-go'
+      iv_tech_name      = 'GO'
+      iv_css_class      = 'ro-go'
```
FreHu, line 621: This is caused by a new option
image

I can undo the formatting if it's undesired, but as long as it is done automatically i think this is better

mbtools:
The current repo list takes a long time because all repo XML is being transformed to ABAP in the persistence layer... and this includes the checksums for all files/objects. So if you have big repos, then it will take a long time (and use a lot of memory). See #4555.

I'm planning to split the `local_checksums` into a separate DB key. Then getting the repo list should be much faster. 

Still your change will be an improvement 👍

src/persist/zcl_abapgit_persistence_db.clas.abap
```diff
@@ -160,6 +163,14 @@ CLASS zcl_abapgit_persistence_db IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD list_by_keys.
+    FIELD-SYMBOLS: <ls_key> LIKE LINE OF it_keys.
+    LOOP AT it_keys ASSIGNING <ls_key>.
+      SELECT * FROM (c_tabname) WHERE value = @<ls_key> APPENDING TABLE @rt_contents.
```
larshp, line 169: downport,
```suggestion
      SELECT * FROM (c_tabname) WHERE value = <ls_key> APPENDING TABLE rt_contents.
```

FreHu: This is maybe an interesting case for abaplint - `SELECT * FROM (c_tabname) WHERE value = <ls_key> APPENDING TABLE rt_contents.` is a syntax error if @ is omitted on higher releases.

The universal solution is
```abap
SELECT * FROM (c_tabname)
      APPENDING TABLE rt_contents
      WHERE value = <ls_key>.
```

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -714,4 +736,13 @@ CLASS zcl_abapgit_gui_page_repo_over IMPLEMENTATION.
     rv_html = |<span class="action_link">| && iv_content && |</span>|.
   ENDMETHOD.
 
+
+  METHOD set_only_favorites.
+    mv_only_favorites = iv_only_favorites.
+  ENDMETHOD.
+
+  METHOD get_only_favorites.
+    rv_result = me->mv_only_favorites.
```
mbtools, line 745: remote `me`. this should be a setting in abaplint.json

FreHu:
Measurement
image

src/persist/zcl_abapgit_persistence_db.clas.abap
```diff
@@ -160,6 +163,16 @@ CLASS zcl_abapgit_persistence_db IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD list_by_keys.
+    FIELD-SYMBOLS: <ls_key> LIKE LINE OF it_keys.
+    LOOP AT it_keys ASSIGNING <ls_key>.
+      SELECT * FROM (c_tabname)
+      APPENDING TABLE rt_contents
+      WHERE value = <ls_key>.
```
larshp, line 171: hmm, should `TYPE` be part of this SELECT? add as input parameter

image

FreHu: added

larshp:
I've changed the title, this PR changes the user experience(show), not the technical(load)

larshp:
perhaps add a text in the bottom of the list when only favorites are shown? after an update I'd see it as some of my repos suddenly disappeared

image

FreHu:
image

larshp (approved):
👍

PR #4998, opened 2021-10-02
Fix pull behavior

sbcgua:
Issue: objects with mixed status are ignored on pull

To recreate: 
- clone ajson - https://github.com/sbcgua/ajson
- switch to filtering branch
- pull
- switch back to master
- pull - nothing happens, although it is intuitively should just pull

2021-10-02_18h56_39

Adjusted the logic so that delete-like actions are not defaulted to "no". Popup is raised instead (existing logic)


sbcgua:
maybe merge ?

PR #4999, opened 2021-10-03
Pull on package reassign

sbcgua:
NB: Ready for review, but makes much more sense after #4998 and #4980 (**I'll rebase when they are merged**)

The problem: When files were moved to another packages Pull command is not active, which is wrong
image

Cause: L/R states do not reflect reassignment. The `packmove` flag does and it is not analysed or aggregated

Fix: (with some refactoring - couldn't pass by - that spagetty would become even worse after :)
- a new class `item_state` (+ UTs) was created to unify the logic of state aggregation - nothing particulary new, just a convenient helper
- the `state` class removed - I could do it there, but I though that the name is not clear `abapgit_state` can be used for something more global in future
- the new class take into account `packmove`
- the rest is rewiring
- also added a dedicated type for item state in definitions


mbtools:
nice. this makes it much clearer. 

`packmove` is similar to a "rename" on the remote side. if you see files, `add` and `delete` make sense. aggregated it could be `R` (rename). `R` would also make sense if other objects are renamed, for example, mime objects.

sbcgua:
> aggregated it could be R (rename)
Yeap, that's what I was also thinking of. But, in another PR :)

PR #5002, opened 2021-10-03
SCI result at stage and commit pages

sbcgua:
Wanted to do it for a while ... #3935 (6)

So if SCI passes, the SCI page is not displayed any more, instead "SCI: PASSED" is displayed at stage and commit pages. Similarly "Forced Stage" displays "SCI: FAILED"
2021-10-03_22h38_15
2021-10-03_22h38_30

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_code_insp.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_code_insp.clas.abap
@@ -12,6 +12,10 @@ CLASS zcl_abapgit_gui_page_code_insp DEFINITION PUBLIC FINAL CREATE PUBLIC
         RAISING
           zcx_abapgit_exception,
 
+      is_nothing_to_display
+        RETURNING
+          VALUE(rv_yes) TYPE abap_bool,
+
       zif_abapgit_gui_event_handler~on_event
         REDEFINITION,
 
@@ -60,7 +64,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_gui_page_code_insp IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_GUI_PAGE_CODE_INSP IMPLEMENTATION.
 
 
   METHOD ask_user_for_check_variant.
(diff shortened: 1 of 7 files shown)


larshp:
@christianguenter2 can you help to review this change?

mbtools:
Nice addition. I would render it a bit less bold

- Label "Checks:" (on background color)
- "Passed" and "Failed" in regular case (no caps)

If the check fails, you want to click on it to view issues, of course 😉

christianguenter2:
Did some tests, the feature is working as expected.

FreHu:
icons would be nice, the green one can be found here
image
and i suppose the one in the error banner would work for failure
image

src/ui/zcl_abapgit_gui_page_code_insp.clas.abap
```diff
@@ -231,9 +241,22 @@ CLASS zcl_abapgit_gui_page_code_insp IMPLEMENTATION.
           " we need to refresh as the source might have changed
           lo_repo_online->refresh( ).
 
+          READ TABLE mt_result TRANSPORTING NO FIELDS WITH KEY kind = 'E'.
```
FreHu, line 244: my suggestion is to keep the whole mt_result, then you can for example show the number of errors in the hover text

FreHu: as in export it to the page insstead of calculating the status here

sbcgua: Number of errors: yeah, good idea, will check
Calculating status ... actually I think it belongs to SCI class more then to stage ... what I wanted to do initially is to separate sci runner. Because now a service (controller) asks the sci page (view) about the result of sci (model) ... (?!?!?!) But I think it is a subject for another PR

sbcgua:
Original
image

Smaller font (+increased paddings) - my fav
image

With icon (maybe not bad)
image

... yeah, maybe not bad ...
image

No caps (no, really ... caps were good!)
image

Number of msgs - hmmm ... doesn't it look cluttered ? I actually like the idea ... but a bit cluttered ... is it worth ?
image

sbcgua:
Yeah ! This is it
image
image
And maybe
image

PR #5004, opened 2021-10-03
Don't show github note at password dialog on non-github repos

sbcgua:
Github
image

Non github
2021-10-03_23h19_13

Hmmm, there are still those ugly empty lines at the bottom ... is it possible to get rid of them too ? I'm not a big expert in sel screens :(


mbtools:
try these

```abap
IF lv_show = abap_true.
  screen-active    = '1'.
  screen-invisible = '0'.
ELSE.
  screen-active    = '0'.
  screen-invisible = '1'.
ENDIF.
```

sbcgua:
image

Cool, Thanks Marc :+1:
Weird are abap ways ... :)

mbtools (approved):
👍

PR #5021, opened 2021-10-06
Add pull and diff buttons to main page

### LINKED ISSUE #4996: Repo list - revise available commands
In the past these commands were columns in the repo list and took too much space, but this is no longer a problem. Commands are now also filterable per online/offline repo and have shortcuts.
image

There is also the minor inconsistency that `Check` actually runs code inspector because previously there was no space for two links. But I personally don't need the syntax check option as I prefer to define a check variant with just syntax check enabled and use that.

I would personally like to add `Diff` as well as `Selective Pull` to the menu. `Remove` would make sense (it's not unsafe, you can just add it again if you accidentally remove a repo), but `Uninstall` should remain an advanced option.

FreHu:
resolves #4996 
image

If there's no diff, an error banner will be shown:
image


mbtools (approved):
👍

PR #5022, opened 2021-10-06
Don't lowercase transport request id

### LINKED ISSUE #5020: Filtering with copied transport (uppercase) from SE09 in stage page (lowercase) not working
After the redesign of the pages with displaying transports in lower case the filtering of the list with transport requests copied from the standard transactions (SE09/SE10...) is not working anymore. 

The transport request in standard transactions are always shown in uppercase, so it would be very helpful to get this possibility back to filter in the stage page in uppercase.

FreHu:
resolves #5020


FreHu:
alright then

PR #5026, opened 2021-10-06
Change abapgit font to 72

FreHu:
This changes the default abapgit font to the SAP-approved 72 font which is probably optimized for uppercase letters.

https://experience.sap.com/72/

image
image
image
image


FreHu:
i believe sapgui uses this font too in belize themes

PR #5032, opened 2021-10-12
Add all Objects to Transport - fix log display

christianguenter2:
When there is an error in 'Add all Objects to Transport', currently a write list appears which is not compatible with abapGit UI.
image
image

The error list is replaced with custom abapGit error log popup.
image


christianguenter2:
Well, it already contains gui popup
https://github.com/abapGit/abapGit/blob/b5d3ac5282e85d94300813db1a0a2366b85713f4/src/cts/zcl_abapgit_transport.clas.abap#L80-L83
TRINT_REQUEST_CHOICE also fires a popup asking for transport request. Imho it's ok for now. But you're right it would be nice if we could get rid of the gui dependencies.

PR #5039, opened 2021-10-16
Diff/Patch: Split view - enable line hover

christianguenter2:
Makes it a bit easier to track line changes.

abapgit_patch_hover


christianguenter2:
Wow, that's very nice indeed. Thank you very. I updated the CSS and enabled it for unified view also. Though did go for a bit lighter grey.

PR #5042, opened 2021-10-18
XSLT: Fix activation error

### LINKED ISSUE #5001: Error from XSLT activate, 0

mbtools:
Remove activation from `deserialize` and use `zcl_abapgit_objects_activation` instead.

Closes #5001

### DIFF
--- a/src/ui/zabapgit_css_common.w3mi.data.css
+++ b/src/ui/zabapgit_css_common.w3mi.data.css
@@ -501,6 +501,11 @@ table.diff_tab td.patch, th.patch {
   user-select: none;
   cursor: pointer;
 }
+
+table.diff_tab tr.diff_line:hover td {
+  background-image: linear-gradient(rgba(0, 0, 0, 0.075), rgba(0, 0, 0, 0.075));
+}
+
 table.diff_tab td.num::before {
   content: attr(line-num);
 }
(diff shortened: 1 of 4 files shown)


mbtools:
Tested with https://github.com/abapGit-tests/XSLT and https://github.com/abapGit-tests/XSLT_ref

PR #5053, opened 2021-10-23
Serialization performance: pre-filter by transport

Fmueller17:
Hello,
my name is Frank. I‘m working in a bigger SAP implementation project where we have several 1000 Z-Objects in a GIT-Repository.
The staging of objects takes a long time (also in parallel mode), because Transaction ZABAPGIT always serializes all objects of the repository.
That’s why I developed a pre-filter where the user can optionally select SAP transport requests / tasks with the objects he wants to stage.
I implemented a popup in the stage functions where the user can select the relevant transport requests.
The pre-filter selects all objects of the transports from table E071 and determines the R3TR objects of it (TADIR entry). For example, if there is just a LIMU Object like a method in the transport then the filter determines with the SAP Standard function module GET_R3TR_OBJECT_FROM_LIMU_OBJ the class object of the method. This function module is also in the Mini ABAP Trail System available. So I assume it is in all SAP Systems available.
Unfortunately, I don’t know how to get the TADIR entry for other object types like LANGU. So I implemented a workaround, that add this object type to the filter, I add the R3TR object type with the same object name to the filter and I try to find with LIMU object type and original object name to get the R3TR object (see method ZCL_ABAPGIT_REPO_PRE_FILTER->GENERATE_LOCAL_FILTER for details).
The filter creates based on these objects a range table that I pass to the serialization class zcl_abapgit_serialize (existing parameter). I redefined the method ZCL_ABAPGIT_REPO_ONLINE->GET_FILES_LOCAL for it, because the filter works only with the online repositories.
The new filter class ZCL_ABAPGIT_REPO_PRE_FILTER creates also based on these objects a pattern with filenames. These filenames contain the object name, the object type and a * at the end. I filter the received files from the GIT-Repository with this pattern in the method ZCL_ABAPGIT_REPO_ONLINE->GET_FILES_REMOTE.
I would be very happy, if you add this feature implementation to the standard ABAPGIT.

Best regards
Frank

PR #5060, opened 2021-10-26
Serializer Cache (experimental)

mbtools:
The serialize process can take quite a while, especially for larger repos. This is an attempt to introduce a cache that improves the runtime significantly. 

The cache was implemented based on the following requirements:

- No persistence (cache is only valid when working with a repo)
- Single user (no shared memory)
- Only some object types are relevant (most types are serialized pretty fast)
- Logic to determine if an object can be or should not be cached must be fast

The current implementation supports `prog`, `clas`, and `intf`. I considered `fugr` but the logic would be too complex/slow.

Here's how it works:

1. Opening repo

When opening the repo, all objects are serialized as usual and the results (file content) are saved to the cache. There is no performance improvement here. 

2. Working with repo

When using features like diff or stage, objects need to be serialized again (to make sure we get the latest version). The cache will first determine, which objects were created or changed "today" (for the object types mentioned above). This logic is based on DB selects using primary key and for-all-entries and is quite fast (it doesn't determine the "last changed date" but simply compares dates to `sy-datum`). For objects that are older than "today", the files are returned from the cache. Other objects are serialized as usual. 

The assumption is that the vast majority of repo objects are not being worked on and therefore you get a high cache hit ratio. The performance of these subsequent serialize tasks should be much better. 

3. Switching repos

Currently, objects will accumulate in the cache (until you exit abapGit). It might make sense to reset the cache when opening a different repo to keep memory consumption low.

My observations:

I have been using this for some time now and it seems to work well. My repos contain mostly OO code so the improvement is very obvious: diff and stage come up much faster (most time is now fetching the remote repo). For abapGit, there are currently about 420 objects in the cache and it uses about 250KB of memory. 

Results will be different for repos that contain a lot of other object types.

The impact on parallel processing is not clear to me yet (I actually disabled parallel processing). Maybe it should be a shared memory class but it would be limited to the developer version of abapGit.


sbcgua:
Hehe, the history seems to go in circles :))) #366, #370
P.S. I'm actually +1 for caching. I just think that the lesson learned from the previous approach is that the cache should be a easily pluggable (or unpluggable) layer.
P.P.S "today" may not be ideal for those who work at night and/or for those whos server is in a different time zone. I think it should be today and yesterday, or last 24h (when possible). Looking at fixed date range could be good idea. Simpler than timestamp approach from the former cache and yet effective enough.

mbtools:
Closing in favor of https://github.com/abapGit/abapGit/issues/5137

sbcgua:
Imho some feature to narrow down the list of serialized object would still be useful. We might get back to this concept in a while :)

mbtools:
That's what the pre-filter should do #5053

PR #5074, opened 2021-10-29
Better multi-lang class tpool handling

sbcgua:
Problem: in the original system there were class texts with translations. Later the texts were deleted. But the translations remained ! Somehow. Maybe because of the fact that this is a standalone system without further transports but not sure. And now the serialized XML contains translation but no texts in the master lang which is a bug.

2021-10-29_19h59_51
2021-10-29_20h00_23

This PR fixes that.

P.S. Sorry for ZCX_ABAPGIT_EXCEPTION - auto capitalization in 731 ...


mbtools:
👍 

just edit ZCX_ABAPGIT_EXCEPTION in github ;-)

sbcgua:
> just edit ZCX_ABAPGIT_EXCEPTION in github ;-)

Mmm, should I do it for every PR ? :)

mbtools:
isn't it lower case after a pull? anyway, it's on the agenda to automate this annoyance away https://github.com/abapGit/abapGit/issues/5018#issuecomment-951038221

mbtools (approved):
lgtm 👍

PR #5076, opened 2021-10-30
Strip FUGR generator comments

sbcgua:
to #5073

OK, presenting for your judgement. Don't judge too strictly :)
Works quite nicely imho :)

image


mbtools (approved):
👍

mbtools:
Note: This is a fix/workaround for a bad practice of SAP to include dates/versions in source code

PR #5097, opened 2021-11-03
Performance Optimization with Namespace

Fmueller17:
This solves the Issue #5095

### DIFF
--- a/src/objects/core/zcl_abapgit_tadir.clas.abap
+++ b/src/objects/core/zcl_abapgit_tadir.clas.abap
@@ -81,7 +81,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_tadir IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_TADIR IMPLEMENTATION.
 
 
   METHOD add_local_packages.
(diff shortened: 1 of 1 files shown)


mkaesemann:
Maybe Im missing something, but depending on the order in CT_TADIR this solution is not actually better.
I would have expected something along the lines of collecting looping over CT_TADIR once, collecting the namespaces and namespace entries in sorted tables, filtering the existing namespace entries from the sorted collected table and then creating  NSPC entries based on the remaining (missing) namespaces.
That way the runtime should be down to O(n+x), or am I mistaken?

Fmueller17:
I'm collecting the namespaces in a separated table lt_tadir_nspc, that contains only namespaces entries. There should be normally only one entry in exception scenarios there could be may be 10 entries. So a sorted table is therefore from my point of view not required. I added additional the field lv_prev_namespace. I avoid with it to read this small table, if we have the same namespace as on the previous object with the a namespace.
grafik

mbtools:
I implemented it originally but now that I think about it....

If I'm not mistaken, there can be only one namespace in a repo (or none). The package will be in a namespace and therefore all objects must be in this namespace, too. 

No looping over all objects required. Just find the first one.

mkaesemann:
@mbtools Sorry to disagree, but that assumption is not true. We have multiple namespaces in our add-on repository (lots of history involved there :-))
@Fmueller17 I did miss the insert on my first read-through. You are correct. Personally I usually try to avoid assumptions about the scenarios as much as possible no matter how unlikely they are, so I would have gone for a hashed or sorted table, but for most cases your solution will still yield an improvement this way, so I retract my comment.

Fmueller17:
@mkaesemann thank you very much for you comment. It is the first time that I heard from a project that has more than 10 namespaces. How many do you have? Do you think that an adjustment to a sortet table is required? The previous logic takes about 3 seconds with a table that has about 7000 entries.

mkaesemann:
@Fmueller17: Actually we also only have 4-5 namespaces (albeit far more than 10000 objects), so even for us your change would be sufficient. My note about not making assumptions was more a general remark, that I generally try to avoid making assumptions (who knows if there actually are projects with more namespaces ;-)).
So 'required' is definetly too strong of a word. If you do want to improve it further, I certainly don't mind, but the additional performance gain will probably be minimal.

mbtools:
I will stop assuming and celebrate that my original code was functionally correct... albeit slow 😉 

Reminder to self: https://en.wikipedia.org/wiki/Murphy%27s_law

mbtools (approved):
👍

larshp:

Fmueller17:
@larshp thank you very much :grinning:

PR #5100, opened 2021-11-03
Commit page: limit comment input length

christianguenter2:
Currently it's possible to input more characters than allowed.
grafik

+ As comment length is limited, we should only allow inputs up to that limit to increase UX
+ placeholder message 
grafik


larshp (dismissed):
The 50 character is a recommendation, https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

not a limitation in the git protocol, I think typical command line git also allows any commit message length?

christianguenter2:
Yeah, it isn't static. Can be configured in settings.
grafik

This restores the behavior of the old commit page.

mbtools (approved):
Aha, you found the missing part of my PR. thanks!

PR #5121, opened 2021-11-08
bug: When Imoprt ZIP does not display error log.

boy0korea:
In case of offline + Import ZIP, error log is not showed.


boy0korea:
[test_check_namespace.zip](https://github.com/abapGit/abapGit/files/7493598/test_check_namespace.zip)

mbtools:
Tested. works

PR #5122, opened 2021-11-09
Hotkey overview: Show link hint activation key

christianguenter2:
If link hints are activated in user settings
grafik
then they are shown in hotkey overview
grafik


mbtools (approved):
👍

PR #5123, opened 2021-11-09
Add Feature: Create New Branch for Commit

### LINKED ISSUE #5091: Creating new branch at commit page
Follows #5087 @mbtools 

> I see the commit as the final step, where you confirm what will actually happen. Switching things will confuse and the file statuses will be incorrect (we would have to refresh from remote and calc diff again). Switching or creating branches should happen on the staging page.

- I agree that **switching** at commit page is a bad idea
- but **creating new branch** is actually good, and it is a proper place in terms of flow

How it happens now: I stage, choose the files, maybe validating the diffs also, (so spending some time), then "Ah, I had to switch the branch!", and I do create a branch directly from the stage page, which also triggers re-fetch and re-serialization so it is slow and annoying. It would be nice to specify a branch name and then AG would create and push in one step.

Well, maybe this option should be at stage page ... but intuitively I would expect it at the latest moment before the commit. Also the commit page has an appropriate form, while at stage there is no good place for it. And also "new branch" would not create conflicts.

imho.

g-back:
- Allow to create a new branch before committing
- No validation of branch name in form itself
- create follow up issue in new docs repo (if this PR is accepted/merged) to adjust screen shot, maybe add a small section for this functionality

image

closes #5091

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_commit.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_commit.clas.abap
@@ -108,9 +108,14 @@ CLASS zcl_abapgit_gui_page_commit DEFINITION
         VALUE(rv_text) TYPE string.
     METHODS is_valid_email
       IMPORTING
-        !iv_email       TYPE string
+        iv_email        TYPE string
       RETURNING
         VALUE(rv_valid) TYPE abap_bool.
+    METHODS branch_name_to_internal
+      IMPORTING
+        iv_branch_name            TYPE string
+      RETURNING
+        VALUE(rv_new_branch_name) TYPE string.
 ENDCLASS.
 
 
@@ -442,6 +447,9 @@ CLASS zcl_abapgit_gui_page_commit IMPLEMENTATION.
 
   METHOD validate_form.
 
+    DATA: lt_branches        TYPE zif_abapgit_definitions=>ty_git_branch_list_tt,
+          lv_new_branch_name TYPE string.
+
     ro_validation_log = mo_form_util->validate( io_form_data ).
 
     IF is_valid_email( io_form_data->get( c_id-committer_email ) ) = abap_false.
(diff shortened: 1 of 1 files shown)


src/ui/zcl_abapgit_gui_page_commit.clas.abap
```diff
@@ -322,7 +324,15 @@ CLASS zcl_abapgit_gui_page_commit IMPLEMENTATION.
...
       iv_name        = c_id-author_email
-      iv_label       = 'Author Email' ).
+      iv_label       = 'Author Email'
+    )->checkbox(
+      iv_name        = c_id-new_branch_checkbox
+      iv_label       = 'Create New Branch for this Commit'
+    )->text(
+      iv_name        = c_id-new_branch_name
+      iv_label       = 'New Branch Name'
+      iv_placeholder = 'new_branch_name' ).
```
mbtools, line 334: Something like:
```
iv_placeholder = 'Optionally, enter a new branch name for this commit'
```

src/ui/zcl_abapgit_gui_page_commit.clas.abap
```diff
@@ -322,7 +324,15 @@ CLASS zcl_abapgit_gui_page_commit IMPLEMENTATION.
       iv_placeholder = 'Optionally, specify an author (same as committer by default)'
     )->text(
       iv_name        = c_id-author_email
-      iv_label       = 'Author Email' ).
+      iv_label       = 'Author Email'
+    )->checkbox(
```
mbtools, line 328: Remove?

mbtools (commented):
I tested it and it works beautifully. I think we can remove the checkbox. If you enter a branch, it will be created. If you leave it empty, it works as before.

mbtools:
For `validate_form`:

```abap
    DATA lt_branches TYPE zif_abapgit_definitions=>ty_git_branch_list_tt.

    lt_branches = zcl_abapgit_git_transport=>branches( mo_repo->get_url( ) )->get_branches_only( ).

    READ TABLE lt_branches TRANSPORTING NO FIELDS WITH TABLE KEY name_key
      COMPONENTS name = io_form_data->get( c_id-new_branch_name ).
    IF sy-subrc = 0.
      ro_validation_log->set(
        iv_key = c_id-new_branch_name
        iv_val = |Branch already exists| ).
    ENDIF.
```

g-back:
Hmm, this logic is already implemented in `raise_error_if_branch_exists` of the repo_online class so this would create duplicate code.
Could make the method public (or add to repo_online interface)?
But then again, this is pretty minor... what do you think?

mbtools:
using the validation shows the message next to the input field. that's nicer than having an exception at the bottom. maybe, it should be a new `check_branch` method in `ZCL_ABAPGIT_GIT_BRANCH_LIST` that can be reused.

mbtools:
nice 

image

mbtools (approved):
👍

christianguenter2:
Very useful feature. Many thanks 👍

sbcgua:
:+1:

Issue #5099, opened 2021-11-03
Serialization of SAP Standard Includes in a Z-function group


Fmueller17:
We have an enhancement of the material master in a Z-function group. This Z-function group contains a lot of SAP Standard Includes. AbapGIT generates during the serialization also files that contains these SAP Standard Includes. Is this a bug? Will the deserialization overwrite the SAP Standard Includes?

mbtools:
The repository must include only objects that are part of the associated SAP package (or sub-packages). Therefore, none of the SAP standard includes should be serialized.

https://github.com/abapGit/abapGit/blob/main/src/objects/zcl_abapgit_object_fugr.clas.abap#L594

The select above is not sufficient to filter out includes that are part of standard function groups (since they don't have a `tadir` entry). 

Determining the function group and then package will take too long. Try adding a `select for all entries` on `trdir` for the includes and eliminate all that have `cnam` and `unam = 'SAP'` (or `like 'SAP%'`).

Fmueller17:
@mbtools 
I made a test and created a new function group with the include LSEWBSYM.
![grafik](https://user-images.githubusercontent.com/59128840/140183579-c1de0995-6f7d-4c91-9ef9-619009c1eba9.png)

It will be staged
![grafik](https://user-images.githubusercontent.com/59128840/140183810-1ffb06b6-bae9-40f1-90ed-246423f4a79a.png)

I tried to find out how SE80 is determining the package and the original function group
![grafik](https://user-images.githubusercontent.com/59128840/140184121-ce567470-65ff-4462-9578-385a3213eb05.png)

I found this coding, it determines the function group of the include
![grafik](https://user-images.githubusercontent.com/59128840/140184543-ce2bcbb6-df65-4d82-95ec-3200379d3973.png)

![grafik](https://user-images.githubusercontent.com/59128840/140184686-599f1837-e311-4e98-b35c-788019fbcbab.png)

I think we should add it to the method.

What do you think?

mbtools:
Let me see if I understand: For each include, determine the function group based on the logic in your screenshot and compare it to the function group we are serializing. If it's different, then skip this include. 

Sound good to me.

Fmueller17:
Thanks @mbtools I will take care on it, when I find some time.

Fmueller17:
@mbtools I solved it with the pull request #5134 

the include LSEWBSYM will be excluded from staging and the repo view

[...]

PR #5134, opened 2021-11-13
Check that include belongs to the same fugr

Fmueller17:
This pull request solves the issue #5099.

### DIFF
--- a/src/objects/zcl_abapgit_object_fugr.clas.abap
+++ b/src/objects/zcl_abapgit_object_fugr.clas.abap
@@ -645,13 +645,13 @@ CLASS zcl_abapgit_object_fugr IMPLEMENTATION.
         WITH KEY progname = <lv_include> BINARY SEARCH.
       IF sy-subrc <> 0.
         DELETE rt_includes INDEX lv_tabix.
-      ELSE.
-        "Make sure that the include does not belong to another function group
-        IF belongs_incl_to_other_fugr( iv_include = <lv_include> ) = abap_true.
-          DELETE rt_includes.
-        ENDIF.
+        CONTINUE.
       ENDIF.
 
+      "Make sure that the include does not belong to another function group
+      IF belongs_incl_to_other_fugr( iv_include = <lv_include> ) = abap_true.
+        DELETE rt_includes.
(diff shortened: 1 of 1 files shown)


src/objects/zcl_abapgit_object_fugr.clas.abap
```diff
@@ -1202,4 +1212,38 @@ CLASS zcl_abapgit_object_fugr IMPLEMENTATION.
...
+  METHOD belongs_incl_to_other_fugr.
+    " make sure that the include belongs to the function group
+    " like in LSEAPFAP Form TADIR_MAINTENANCE
+    DATA ls_tadir TYPE tadir.
+    DATA lv_namespace TYPE rs38l-namespace.
+    DATA lv_area TYPE rs38l-area.
+    DATA lv_include TYPE rs38l-include.
+
+    rv_belongs_to_other_fugr = abap_false.
+    IF iv_include(1) = 'L' OR iv_include+1 CS '/L'.
```
mbtools, line 1226: Namespace split/check is not correct. Includes are /.../L...

mbtools (changes requested):
Namespaces

src/objects/zcl_abapgit_object_fugr.clas.abap
Fmueller17: @mbtools Hi Marc, why do think it is not ok? I implemented like the SAP Standard. iv_include+1 CS '/L' means search  '/L' after the first '/' (iv_include+0).

grafik

Fmueller17: It is not possible to have more than 2 / in a program name.
grafik

It is also not possible to have 2 / and the first one is not the first character.
grafik

mbtools: that's not it. in your code, `lv_namespace` isn't set. (sorry, my marker wasn't precise. )

mbtools: you are right (i read it as `=`)

mbtools (approved):
👍

PR #5136, opened 2021-11-15
add more toolbars to Command Palette

christianguenter2:
e.g. on various settings pages
grafik

and others which haven't been supported because they're part of high order component (hoc)
grafik

now all toolbars which ids start with `toolbar` are considered.


mbtools (approved):
👍

PR #5146, opened 2021-11-19
Diff: Add options to ignore indent, comments, case

mbtools:
Continuing on #2804

Enjoy!

image


larshp (approved):

christianguenter2:
Really useful feature 👍 Thx

christianguenter2:
Would be useful on patch page too?

albertmink:
`Ignore Indentation` suggest that it does ignores only leading white-space. Indeed, it seams it ignores white space changes at all. This diff is `no diff` for `Ignore Indentation`.
image

Might align to git wording, e.g. `space change`
image

mbtools:
You're right about the indentation/whitespace changes. We are reusing the standard SAP diff, which describes it incorrectly:

image

I was just using the same terminology 😀 .  

How about "Ignore Whitespace"?

albertmink:
Ignore Whitespace is definitely better than the indentation 😅

PR #5153, opened 2021-11-20
Expand TABL includes before PUT

### LINKED ISSUE #5152: TABL - wrong deserialization of foreign keys
Issue. **In the original system** there is a strcture with check tables. In an include !

image

The DD08L contain noinherit flag = N (so "inherit=yes")

image

Thus TABL object removes it from the XML

image

In the target system, the same object produces a diff. Force pull does not help. Note the noinherit = Y. Apparently, the records are autoadded 

image

By the way DD08L contains the records. And the SE11 does not show them which is strange (or I'm just mixing the tables)

image

Apparently records are auto added in `DDIF_TABL_PUT` but not 100% sure. Also not sure why this happened in the first place. As I understand stripping inherited records added relatively recently. But this looks like a bug because force pull does not help, it should  fix things. I'm see it in at least 2 target systems.

sbcgua:
Closes #5152

maybe some DD gurus should validate... but looks like a simple fix and follows what actually SE11 does.

And worked on my target systems. (no more irrelevant diffs !)

### DIFF
--- a/src/objects/zcl_abapgit_object_tabl.clas.abap
+++ b/src/objects/zcl_abapgit_object_tabl.clas.abap
@@ -801,6 +801,22 @@ CLASS zcl_abapgit_object_tabl IMPLEMENTATION.
       corr_insert( iv_package = iv_package
                    ig_object_class = 'DICT' ).
 
+      CALL FUNCTION 'DD_TABL_EXPAND'
+        EXPORTING
+          dd02v_wa          = ls_dd02v
+        TABLES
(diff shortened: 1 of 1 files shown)


mbtools:
`DD_TABL_EXPAND` has a `MODE` parameter but I think the default (1 = activation) is ok here since we want to activate the object right after the PUT.

sbcgua:
Yes mode = 1 by default. This was also the value SE11 passed there, so I didn't pass it explicitly.

PR #5158, opened 2021-11-22
Refactor: Jump to Object Logic

mbtools:
Continue #5114

- Removes jump_adt and jump_se11 from zcl_abapgit_objects_super
- Remove almost all object-specific jump handlers (zif_abapgit_object~jump) since they are covered by GUI Jumper now


christianguenter2 (approved):
LGTM 👍

PR #5169, opened 2021-11-29
Repo view: Shorten path

mbtools:
If files are shown and the starting folder is at least two levels deep (for example `/exercises/practice/`), then the display of the path will be shortened by replacing the starting folder with `/.../`. If you hover over the `/.../`, it will show the original path.

Before:

image

After:

image

### DIFF
--- a/src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
+++ b/src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
@@ -39,6 +39,7 @@ CLASS zcl_abapgit_gui_page_repo_view DEFINITION
     DATA mv_max_setting TYPE i .
     DATA mv_show_folders TYPE abap_bool .
     DATA mv_changes_only TYPE abap_bool .
+    DATA mv_starting_folder TYPE string .
     DATA mv_order_by TYPE string .
     DATA mv_order_descending TYPE abap_bool .
     DATA mv_diff_first TYPE abap_bool .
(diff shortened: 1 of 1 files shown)


larshp:
hmm

mbtools:
It's only if the starting folder has several levels, which is most of the time, not the case (pretty much everyone is using `src`) and it's still there if you want to see it:

image

mbtools:
an alternative would be to remove the starting part completely:

`clock/package.devc.xml` instead of `/exercises/practice/clock/package.devc.xml`

flaiker:
Another idea would be to replace the starting path with `~`. Then some people are somewhat used to the concept of `/` starting at the top level and some special directory, like the home directory of the current user / source code directory in abapGit, being replaced for easier access?

mbtools:
`~` makes sense. I changed it.

sbcgua:
Hmmm, I also think this is a bit controversial feature. The path represents a subpackage, and I'd prefer seeing and double checking if the package assignment is correct.
I do have real cases of incorrect package assignment in my team and I even though about **adding** subpackage field at the main repo screen (but that's another story of course).
If we are to merge it, can it be done switchable ?

mbtools:
Just to clarify, it replaces only the starting folder in git which is not part of the package/sub-package structure. It would actually remove confusion. 

I kept it limited to starting folders with at least 2 levels, so few will see any difference.

sbcgua:
Hmmm, but then "..." is confusing. what you've wrote is not obvious :) (telling that as a unaware user :)
maybe I'd then replace it with either just ".", or something like "$". A) it is shorter, B) it supposes that "$" is something specific (the starting folder) and not just arbitrary shortening of long path.
Imho

sbcgua:
And yes, it should be then replaced always

sbcgua:
if it should ... :)

mbtools:
No consensus and the Exercism repo isn't happening like this. Closing.

PR #5170, opened 2021-11-29
Support dot-folders in repositories

mbtools:
Some repositories use (hidden) dot-folders, for example `https://github.com/exercism/abap`, which were not supported by abapGit. Such folders are now mapped to SAP packages replacing `.` with `$`.

Example:

Starting folder `/src/` 
Path `/src/clock/.meta` 
SAP Package `$MAIN_CLOCK_$META`


larshp (changes requested):
hmm, "$" has special meaning in package names, perhaps its only the first character? Suggest using "_" instead. Alternatively just not support this scenario

PR #5171, opened 2021-11-29
Commit page: option to hide author fields

mbtools:
The vast majority of the time the committer is the same as the author (I assume). This adds an option to hide the author name and email fields on the commit form.

image

It streamlines the commit page and avoids scrolling:

image


christianguenter2 (approved):
👍

PR #5172, opened 2021-11-29
Refactor: Consolidate front-end classes

mbtools:
There's no need for two "front-end classes". Moves the remaining methods of `ZCL_ABAPGIT_GUI_FUNCTIONS` to `ZCL_ABAPGIT_FRONTEND_SERVICES`. Now all front-end calls are nicely together and you need only one factory/injector.

No functional change, some method reordering. 

Ref #2510, #5145


christianguenter2 (approved):
LGTM 👍

PR #5175, opened 2021-12-01
Fix vimium links

mbtools:
- No Vimium links for hidden fields
- Fix Vimium link for radio selections 
- Move link box to end of labels (fixes the unwanted line-breaks between input control and label)

Closes #5038 (for @christianguenter2 😉)
 
Example:

image

PR #5176, opened 2021-12-03
Patch: Fix link hints

christianguenter2:
Regression from #5175 - link hints don't work anymore on Patch page

before
grafik

grafik

grafik

after
grafik


mbtools (approved):
The mix of ABAP and JS is killing me 🥴. Thanks for fixing it.

PR #5184, opened 2021-12-06
WebGUI: Disable jump

### LINKED ISSUE #5183: WebGUI: Jump not possible
In WebGUI it's not possible to jump to objects because most (if not all?) jumps open new windows or are opened in eclipse. This is not possible in browser.

GUI
grafik

ADT
grafik

Therefore I'd suggest to disable jump in WebGUI completely.

christianguenter2:
fixes #5183

### DIFF
--- a/src/objects/core/zcl_abapgit_gui_jumper.clas.abap
+++ b/src/objects/core/zcl_abapgit_gui_jumper.clas.abap
@@ -112,6 +112,11 @@ CLASS zcl_abapgit_gui_jumper IMPLEMENTATION.
 
   METHOD zif_abapgit_gui_jumper~jump.
 
+    " WebGUI cannot open windows or ADT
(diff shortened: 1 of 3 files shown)


mbtools:
it could still jump to the object if we stay in the same window. so for WebGUI it should skip ADT and then use IV_NEW_WINDOW = false.

christianguenter2:
> it could still jump to the object if we stay in the same window. so for WebGUI it should skip ADT and then use IV_NEW_WINDOW = false.

Yeah, opening in same window caused some trouble with navigation in the past. Theferore AFAIR most of old jump logic opened new windows. So I'm not sure if that's an improvement here.

mbtools:
ok, let's disable it. when webgui becomes a feasible alternative (no more gui popups) we can revisit this.

mbtools (approved):

PR #5187, opened 2021-12-07
performance optimization parallel mode

manhofmann:
in parallel mode a new session is created for each object to be serialized. method IS_TYPE_SUPPORTED is called in each new session with only one object type. however all objects are loaded via method SUPPORTED_LIST. this change implements a specific cache for method IS_TYPE_SUPPORTED.


larshp:
quick runtime measurements, before ~6 sec, after ~2 secs, serializing abapGit,

image

larshp (approved):
thanks, looks good to me

will leave the PR open for a bit, to allow others to also review, and/or provide feedback

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -1017,30 +1048,45 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
...
-    IF gt_supported_obj_types IS NOT INITIAL.
-      rt_types = gt_supported_obj_types.
+    IF gv_supported_obj_types_loaded = abap_true.
+      LOOP AT gt_supported_obj_types ASSIGNING <ls_supported_obj_type> WHERE supported = abap_true.
+        INSERT <ls_supported_obj_type>-obj_type INTO TABLE rt_types.
+      ENDLOOP.
       RETURN.
     ENDIF.
 
+    " delete content because it might ne filled already by method IS_TYPE_SUPPORTED
```
christianguenter2, line 1065: small typo
```suggestion
    " delete content because it might be filled already by method IS_TYPE_SUPPORTED
```

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -4,6 +4,14 @@ CLASS zcl_abapgit_objects DEFINITION
...
   PUBLIC SECTION.
 
+    TYPES:
+      BEGIN OF ty_supported_types,
+        obj_type  TYPE tadir-object,
+        supported TYPE abap_bool,
+      END OF ty_supported_types.
+
+    TYPES: ty_supported_types_tt TYPE SORTED TABLE OF ty_supported_types WITH UNIQUE KEY obj_type.
+
```
christianguenter2, line 14: I'd suggest moving these types to `PRIVATE SECTION`

christianguenter2 (commented):
Two small remarks. See comments.

mbtools:
Nice. Big improvement for parallel setting.

Sidenote: The list of supported object types is actually an ideal candidate for a shared memory class (minimal memory, usable across users, sessions, and servers). Would work for the dev. version, only, though.

src/objects/zcl_abapgit_objects.clas.abap
manhofmann: thanks. good hint. done

larshp:

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -888,13 +897,35 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
...
+    FIELD-SYMBOLS <ls_supported_obj_type> TYPE ty_supported_types.
+
+    IF iv_obj_type IS INITIAL.
+      " empty object type should never exist
+      RETURN.
+    ENDIF.
+
+    READ TABLE gt_supported_obj_types
+      ASSIGNING <ls_supported_obj_type>
+      WITH KEY obj_type = iv_obj_type.
```
Fmueller17, line 912: @manhofmann You should use at this read table with with **table** key instead of with key, because gt_supported_obj_types is a sorted table with the unique key obj_type. So with table key is faster.

PR #5188, opened 2021-12-08
fixing short dump

lucasborin:
image

image


larshp:
hmm, where exactly does the dump occur?

we have spent quite a lot of time removing the aliases, https://github.com/abapGit/abapGit/pulls?q=is%3Apr+remove+alias+is%3Aclosed

lucasborin:
> hmm, where exactly does the dump occur?

image

larshp:
heh, did not think it was in use, anyhow, thats a fix to https://github.com/abapGit/abapGit-Plugins

larshp (changes requested):
see above ^

lucasborin:
Oh, I see. It is deprecated. Let me remove it from the system.

PR #5223, opened 2021-12-15
Requirements dependency check

jrodriguez-rc:
- If the requirement patch registered has less than 4 characters (08 instead of 0008) check fails
- Added tests for requirement validation


mbtools (approved):
SAP is not storing the patch (extrelease) consistently. I've also seen '00000003' for example. 

If it's not numeric, it should be a string comparison but it didn't support that before either. 

Anyway, thanks for improving it.

PR #5226, opened 2021-12-18
Pre Filter Backend Logic

Fmueller17:
@mbtools , @larshp This request PR contains only the backend logic of the pre filter. I refactored it and split complete the gui logic and backend logic in different classes. I also implemented your suggestions and do not longer use the pre filter in the status classes.


larshp:
please adjust the description of the PR

larshp:
As a developer you decide which changes to include in a PR, we have discussed the criticality and complexity of this change. The changes below are not strictly related to this PR, and not caused by pretty printing or the specific editor. So its a choice to include it in this PR, they might as well be in a different PR, which will be easily reviewable, also making this change easier to review.

image

image

larshp:
For the `zif_abapgit_repo_pre_filter` interface, it is currently designed to work for transports only? We need a solution that works more broadly, as mentioned in https://github.com/abapGit/abapGit/issues/5054#issuecomment-951079434.

```abap
INTERFACE zif_abapgit_repo_pre_filter PUBLIC.

...

  METHODS set_filter_values
    IMPORTING
      iv_package  TYPE tadir-devclass
      it_r_trkorr TYPE ty_trrngtrkor_tt OPTIONAL
    RAISING
      zcx_abapgit_exception .
```

Fmueller17:
> 
@larshp the pre filter can currently only handle transports. The idea is, that when you later add the functionality to filter by other criterias like user then you can adjust the method set_filter_values and add other optional parameters. That's the reason why it_r_trkorr is also optional. I find this approach more flexible than a method set_transport_filter_values.

larshp:
well, if we need to change the interface to add the scenarios, then the abstraction is wrong

larshp:
looking at the interface, it is only these two methods that are called from the central repo logic, ie. all other parts of the interface can be moved to `zcl_abapgit_repo_pre_filter`

```abap
  METHODS filter_files
    CHANGING
      ct_files TYPE zif_abapgit_definitions=>ty_files_tt .

  METHODS get_local_filter
    RETURNING
      VALUE(rt_filter) TYPE zif_abapgit_definitions=>ty_tadir_tt .
```

larshp:
status: interface adjusted by @Fmueller17

larshp:
can `string` be used instead of data element `bapi_msg` ?

Fmueller17:
> can `string` be used instead of data element `bapi_msg` ?
@larshp I replaced it

larshp:
perhaps rename to object filter? as it is an object filter?

zif_abapgit_repo_pre_filter -> zif_abapgit_object_filter
apply_file_filter -> apply_object_filter

Fmueller17:
> perhaps rename to object filter? as it is an object filter?
> 
> zif_abapgit_repo_pre_filter -> zif_abapgit_object_filter apply_file_filter -> apply_object_filter

@larshp  Happy new year. I renamed it.

larshp:
cool, lets go 🏃 , thanks

PR #5227, opened 2021-12-18
Pre Filter Backend and GUI Logic

Fmueller17:
@larshp , @mbtools this PR contains the refactored pre selection backend and gui logic.


larshp (changes requested):
after https://github.com/abapGit/abapGit/pull/5226

src/ui/zcl_abapgit_gui_obj_filter_tr.clas.abap
```diff
@@ -0,0 +1,65 @@
+CLASS zcl_abapgit_gui_obj_filter_tr DEFINITION
```
mbtools, line 1: Let's move this popup to `zcl_abapgit_popups`. Ideally, it should be able to replace the same popup in `zcl_abapgit_transport_mass / select_tr_requests` so we will have only one central place asking for multiple transport requests.

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -477,83 +477,84 @@ CLASS zcl_abapgit_gui_page_repo_over IMPLEMENTATION.
...
-      lv_patch_link    TYPE string,
-      lv_diff_link     TYPE string,
-      lv_pull_link     TYPE string.
-
-    DATA:
-      lv_zip_import_link TYPE string,
-      lv_zip_export_link TYPE string.
-
-    ii_html->add( |<div class="float-right">| ).
+    DATA lo_toolbar TYPE REF TO zcl_abapgit_html_toolbar.
```
mbtools, line 485: Changing the actions to a proper toolbar is a good idea. However, the actions should be only the frequent used commands. So we should not add the filtered options here unless there's consensus.

src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
```diff
@@ -404,11 +417,17 @@ CLASS zcl_abapgit_gui_page_repo_view IMPLEMENTATION.
 
     lo_tb_advanced = build_advanced_dropdown( iv_wp_opt = lv_wp_opt ).
 
+    lo_tb_stage = build_stage_dropdown( ).
```
mbtools, line 420: Similarly, here, "Stage" is used by everyone. It should not be a dropdown. I would put "Stage with Filter" into the "Advanced" menu.

src/ui/zcl_abapgit_html_toolbar.clas.abap
```diff
@@ -212,29 +219,37 @@ CLASS ZCL_ABAPGIT_HTML_TOOLBAR IMPLEMENTATION.
         ENDIF.
       ENDIF.
 
+
       IF <ls_item>-cur = abap_true.
-        lv_class = ' class="current-menu-item"'.
+        lv_class_value =  |current-menu-item { <ls_item>-li_class }|.
```
mbtools, line 224: if `li_class` is initial, then the class value will contain an extra space.

mbtools (changes requested):
See above. (As Lars mentioned, please keep whitespace changes to a minimum since it makes it quite hard to focus on what really changed).

src/ui/zcl_abapgit_gui_obj_filter_tr.clas.abap
Fmueller17: Ok I will move it there and add some parameters to make it flexible.

src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
Fmueller17: see above.

src/ui/zcl_abapgit_html_toolbar.clas.abap
Fmueller17: I will add an IF

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
Fmueller17: @mbtools perhaps not everybody uses this button, but I need it on this screen. If I have to open the repository first then it will take a long time until AG has checked everything. Perhaps I could add there also an "Advanced" menu but then it will have just one entry for now...
Do you have a better idea?

mbtools: you could add a keyboard shortcut for it without adding it to the toolbar. or we add a "more" or "..." icon/menu to the end of the toolbar to collect less used options.

src/ui/zcl_abapgit_html_toolbar.clas.abap
Fmueller17: @mbtools I added the IF. So I assume that the other PR #5230 which only changes the buttons to the toolbar (without the new one) can be merged now?
image

larshp:
does it work if `TRWBO_TITLE` is replaced with `string`?

larshp:
and the type definition of `ty_trrngtrkor_tt` must be moved to some interface, as the build(into the copy paste report) fails

src/ui/zcl_abapgit_gui_obj_filter_tr.clas.abap
Fmueller17: I added the methods popup_select_tr_requests and popup_select_wb_tc_tr_and_tsk to the interface zif_abapgit_popups (class zcl_abapgit_popups).
Popup_select_tr_requests selects the transports with imported parameters. Popup_select_wb_tc_tr_and_tsk calls popup_select_tr_requests with the parameters to select workbench and transport of copies transports and tasks.

So the method zcl_abapgit_transport_mass / select_tr_requests can be adjusted to call one of this methods later.

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
Fmueller17: I moved the new functions to the new "more" sub menu.
image

image

src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
Fmueller17: I move it to the "Advanced" menu.
image

image

Fmueller17:
> does it work if `TRWBO_TITLE` is replaced with `string`?

It was not possible to replace it by string, because it is a parameter of the FM TRINT_SELECT_REQUESTS, but the commit was successful.

Fmueller17:
> and the type definition of `ty_trrngtrkor_tt` must be moved to some interface, as the build(into the copy paste report) fails

I moved the type to the interface zif_abapgit_definitions and then it works.

src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
```diff
@@ -313,22 +314,29 @@ CLASS zcl_abapgit_gui_page_repo_view IMPLEMENTATION.
...
+    ro_advanced_dropdown->add( iv_txt   = 'Uninstall'
                                iv_title = `Delete all development objects belonging to this package `
-                                       && `(and subpackages) from the system`
-                               iv_act = |{ zif_abapgit_definitions=>c_action-repo_purge }?key={ mv_key }|
-                               iv_opt = lv_crossout ).
+                                          && `(and subpackages) from the system`
+                               iv_act   = |{ zif_abapgit_definitions=>c_action-repo_purge }?key={ mv_key }|
+                               iv_opt   = lv_crossout ).
 
+    IF mo_repo->is_offline( ) = abap_true.
```
mbtools, line 333: Let's move this menu item up. It should be after "Add All Objects to Transport". 

Also rename to "Export by Transport" and "Stage by Transport".

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
```diff
@@ -477,83 +477,77 @@ CLASS zcl_abapgit_gui_page_repo_over IMPLEMENTATION.
...
+                     iv_li_class = |{ lc_action_class }| ).
 
-    ii_html->add( action_link( lv_check_link && lc_separator ) ).
+    CREATE OBJECT lo_toolbar_more_sub EXPORTING iv_id = 'toolbar-ovp-more_sub'.
 
-    lv_zip_import_link = ii_html->a(
-      iv_txt   = |Import|
-      iv_act   = |{ zif_abapgit_definitions=>c_action-zip_import }{ lc_dummy_key }|
-      iv_class = |{ lc_action_class } { lc_offline_class }| ).
+    lo_toolbar_more_sub->add( iv_txt      = |Stage, filtered by Transport/Task|
```
mbtools, line 533: Just make the description shorter: Rename to "Export by Transport" and "Stage by Transport".

mbtools (changes requested):
Small changes. See above.

I have items for a follow-up PR:
- Change to transport selection to a centered popup (#5275)
- Fix hotkey display (#5288)

src/ui/zcl_abapgit_gui_page_repo_view.clas.abap
Fmueller17: @mbtools I moved and renamed it.
image

image

src/ui/zcl_abapgit_gui_page_repo_over.clas.abap
Fmueller17: @mbtools I renamed it.
image

image

mbtools (approved):

mbtools:
@larshp this looks good to me now

larshp:
cool, lets go 🚢

PR #5247, opened 2022-01-04
TOBJ - fix xslt serialization dump due to empty numc value

FreHu:
Colleague encountered an error during serialization.

image
image
image

The serialization code looks like it should work, but `initial_components = suppress` which is there does not help in this case.
```abap
      CALL TRANSFORMATION id
        SOURCE (lt_stab)
        RESULT XML li_doc.
```

note that this is likely data corruption and not a real situation, i can't figure out how to even set numc to an empty value


mbtools (approved):
Lgtm

Issue #4395, opened 2021-01-09
Incorrect display of diffs


mbtools:
AG shows the first `ENDMETHOD.` as part of the diff:

![image](https://user-images.githubusercontent.com/59966492/104105641-8757cd00-527d-11eb-8d77-20c93ff6cabb.png)

Github shows correctly the second `ENDMETHOD.` as part of the diff:

![image](https://user-images.githubusercontent.com/59966492/104105688-dbfb4800-527d-11eb-822e-ba922c9f8450.png)

mbtools:
turns out it's a bug in function `SVRS_COMPUTE_DELTA_REPS` more specifically in 

```abap
  CALL 'C_MAKE_TEXT_DELTA'                         "#EC CI_CCALL
    ID 'TI_TEXT_PRI' FIELD ti_text_pri-*sys*
    ID 'TI_TEXT_SEC' FIELD ti_text_sec-*sys*
    ID 'PI_DELTA_OPTIONS' FIELD pi_delta_options
    ID 'TO_TEXT_DELTA' FIELD to_text_delta-*sys*.
```

![image](https://user-images.githubusercontent.com/59966492/104107140-34374780-5288-11eb-9e14-ecac3012585b.png)

SAP kernel... hmmm

mbtools:
I gave it a good shot with SAP. Here's the answer:

```
Dear customer,

while the majority of diffing tools will probably produce your desired result in this case,
the diff result produced by the diff algorithm in the ABAP kernel is still a technically 
correct interpretation of what the diff is depending on the direction in which you traverse the file.

Aside from the technical correctness of the result, we unfortunately cannot change the 
principal algorithm due to compatibility reasons. For example the delivery of ABAP
correction instructions using SNOTE is built upon this algorithm, so a change in its behavior
would break this functionality.

Kind regards, Felix
SAP Development Support
```

I had a similar issue with WinMerge but this tool has an option to change the diff algo 🤷 
https://github.com/WinMerge/winmerge/issues/1053

PR #5251, opened 2022-01-07
Adjust diff algorithm

mbtools:
This implements a workaround for the SAP Kernel issue when determining diffs (see https://github.com/abapGit/abapGit/issues/4395 for background). 

Here's an example of the change. Method `render_field_image` was inserted but SAP Kernel shows a diff related to the previous method. 

Before:

image

After:

image


larshp (changes requested):
need a unit test for this?

mbtools:
Took me a while to reduce the tests I had to something short for a UT but now they work.

larshp (approved):

PR #5255, opened 2022-01-07
Support GitHub dark mode in readme icon

flaiker:
Before:

image

After:

image

Colors according to https://git-scm.com/downloads/logos .

Note the tag switches based on the theme setting on GitHub. You could also query the OS dark/light mode setting using CSS in the svg but the theme setting might be different. Other than on GitHub both icons will appear when the markdown is rendered though instead of there being a more generic solution support for GitHub's solution seems to be added to other editors instead (see bottom of the feature request discussion for links to issues in vscode for example.).

https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/
https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981


larshp:
nice 👓

flaiker:
Hmm, doesn't seem to work in their own app, oh well

flaiker:
Seems like relative links in combination with the light/dark mode tags do not work in the app. I sent an email to the app's feedback address.

PR #5256, opened 2022-01-07
Change zabapgit download

mbtools:
Browsers will download source instead of displaying it


flaiker:
Doesn't seem to change anything in edge, still displayed.

mbtools:
hmm, that's odd:
https://css-tricks.com/building-good-download-button/ 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes

flaiker:
Should work: https://caniuse.com/download

I tested here: https://github.com/abapGit/abapGit/blob/4f3fbd5c161360e49d2043a95185e6e4b91dd6ad/README.md

Seems like GitHub strips it?

image

mbtools:
It's also not "same origin". So I don't think the change will work

flaiker:
Yes that's probably it. Even GitHub doesn't do it when clicking their download button on https://github.com/abapGit/build/blob/main/zabapgit.abap

image

flaiker:
Or wait, does `<a href="https://github.com/abapGit/build/blob/main/zabapgit.abap?raw=true" download></a>` work?

mbtools:
probably striped by the markdown parser. maybe it need to be a button instead of ahref.

larshp:
I changed this PR to draft mode, as it sounds like its not working yet?

mbtools:
dr. google doesn't know an answer either. closing

Issue #5164, opened 2021-11-25
repo overview ordering similar to native git


jfilak:
I often need to deal with huge packages where I care only about few objects.
I version only the objects I am interested in and I leave the other objects as untracked (some super simple "add to ignore would be nice").

In the repo overview, I would love to see  the following order:
- Modified objects remote (or both sides)
- Only remote objects (new remote objects)
- Modified objects local
- Only local objects (new local objects)
- Unmodified objects

mbtools:
Did you try clicking on the "Status" column header? Sorting up or down gets you pretty close to what you describe. 

![image](https://user-images.githubusercontent.com/59966492/143486083-3c07d7ab-d46c-46f4-ad61-ddf485b039f3.png)

jfilak:
Nice tip!
Can we make the "descending" sorting default at least?
However, I still believe the native like grouping would be usefull.
Even more, I don't need to see the list of all unmodified objects, what's the added value of having them listed?

christianguenter2:
I think 'View' > 'Changes only' could be your friend. I there are no modified objects it looks like this.
![grafik](https://user-images.githubusercontent.com/17437789/143488444-7b2ab9bf-f7d2-455e-9a08-9af95c35e762.png)

jfilak:
Good tip again. Thank you!
How can I make that default?

sbcgua:
as far as I remember it should autosave to user profile (or maybe per repo - I don't remember)

mbtools:
View settings are indeed user-specific. Sort order is currently not saved.

flaiker:
@jfilak #5257 should help, can this be closed?

jfilak:
Thank you!

PR #5257, opened 2022-01-08
Save additional repo view settings

flaiker:
#5164

I made order-by 3 states instead of 2 when clicking on a column so you can return to the default sort order.


christianguenter2 (approved):
LGTM 👍

PR #5261, opened 2022-01-10
Repo View: 'Order by path' considers object name

### LINKED ISSUE #5259: Repo View: Should order by path also consider object name?
If we sort by path object name isn't considered. I think working like that it doesn't add much value and it's a bit confusing, at least for me. What about considering path + object name? And also sorting file names accordingly?

grafik
https://github.com/abapGit/abapGit/blob/79961a1f20c92eac456886ab5d48bddb86be0c1d/src/ui/zcl_abapgit_gui_page_repo_view.clas.abap#L236-L237
grafik
grafik

christianguenter2:
fixes #5259 

before
grafik

after
grafik


mbtools (approved):
works nicely. thanks! 👍

PR #5269, opened 2022-01-13
Correct task name parallel processing

### LINKED ISSUE #5224: WebGUI: Runtime Error ASSERTION_FAILED in zcl_abapgit_serialize
Hi all,

When trying to connect to an online repository via WebGUI, I get an `ASSERTION_FAILED` runtime error in `zcl_abapgit_serialize` (line 482, after the `Z_ABAPGIT_SERIALIZE_PARALLEL` function call in method `run_parallel`). `sy-subrc` has value 2 and `lv_msg` says "RFC task 2 already open".

Details:

- Happens only with the non-standalone abapGit via WebGUI. For the standalone version (no matter if SAPGUI or WebGUI) or in SAPGUI (no matter if standalone or not) the issue is not present.
- SAP_BASIS 757
- abapGit version 1.113.0

Any ideas?

Best,
Jonathan

mbtools:
When serializing objects in parallel threads, it can happen that there are not sufficient system resources available (free dialog processes). In this case, abapGit can run into error "RFC task <task name> already open". This is caused by reusing the same task name in subsequent tries to serialize an object.

The correction appends a sequential number to the previous task name so each try will result in a unique task name.

Closes #5224.

PS: This does not fix any resource issues but should prevent the error. It will lead to wait situations and longer processing, however.

### DIFF
--- a/src/objects/core/zcl_abapgit_serialize.clas.abap
+++ b/src/objects/core/zcl_abapgit_serialize.clas.abap
@@ -451,14 +451,16 @@ CLASS zcl_abapgit_serialize IMPLEMENTATION.
   METHOD run_parallel.
 
     DATA: lv_msg  TYPE c LENGTH 100,
+          lv_task TYPE c LENGTH 32,
           lv_free LIKE mv_free.
 
 
     ASSERT mv_free > 0.
 
     DO.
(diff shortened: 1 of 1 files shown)


christianguenter2 (approved):
LGTM 👍

PR #5277, opened 2022-01-18
Fix lost remote on favorite refresh

sbcgua:
There was a subtle bug which actually caused me and my team some troubles recently.
If I **don't have favorites** and try to update an offile repo (via zip), then the diff is displayed but pull does not work (in fact it starts deleting all objects ... and tables ... with data ...).

This happens because `zcl_abapgit_repo_srv->list_favorites` (and `refresh_favorites`) - which are called in particular for repo palette, at the end of rendering, **after** diff was rendered - so this `refresh_favorites` **re-instantiates** repo instances ... consequently it looses `mt_remote` before executing the next command - pull in this case.

The issue is quite easily reproduced. Offline repo -> Import from zip -> see diffs (at object list) -> do any next command e.g. drill down to diff page or pull.

So, the change is as follow:
- don't reinstantiate repos, just reconcile them with the actual fav list

Doubts:
- reconciliation is done not by repo_id or url - which do not guarantee uniqueness (?!) - but by package. Which seems to be double checked when new repo added. But it seems strange to me that URL is not checked in the same manner ...
- some code duplication was introduced due to the current code structure. I think in the long run we should remove the `list_favorites` at all and load all repos. It is the page who should filter the repos to display. (I know there is an issue with checksums. Maybe I'll try to find time to straighten it a bit.)

In addition I renamed the `zif_abapgit_persist_repo~list_favorites` to `list_by_keys`. Because the method receives the list of repos externally and does not know anything about favs inside, so semantics was incorrect. Couldn't pass by this refactoring, sorry ;P

and some SE80 reordrings (no changes to `get_repo_from_content`) ...

### DIFF
--- a/src/persist/zcl_abapgit_persistence_repo.clas.abap
+++ b/src/persist/zcl_abapgit_persistence_repo.clas.abap
@@ -43,7 +43,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_persistence_repo IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_PERSISTENCE_REPO IMPLEMENTATION.
 
 
   METHOD constructor.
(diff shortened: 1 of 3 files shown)


flaiker:
Does this by chance also fix the issue of pressing DIFF on a single file causing of full refresh of the repo if it is not a favorite (https://github.com/abapGit/abapGit/issues/5137#issuecomment-991948028)?

sbcgua:
@flaiker I think so, though better to recheck.

P.S. I also thought that `zif_abapgit_persist_repo~list` might also be prone to similar issues as it also clears `mt_list` ... though there were no reports on that ... maybe in next PR ...

flaiker:
> @flaiker I think so, though better to recheck.

Just tested it, it's fixed

sbcgua:
side effects :)

mbtools (approved):
👍

PR #5285, opened 2022-01-22
APACK manifest not refreshed after class update

### LINKED ISSUE #5284: new APACK version considered only by restarting abapGit
Steps to reproduce:
- Start abapGit, open your repository having an existing APACK class
- Using your favorite ABAP editor, edit the APACK class and change the version
- In abapGit, in your repository, click "Refresh"
- You can see that `.apack-manifest.xml` has not been recalculated, it doesn't contain the new version defined in the APACK class

Workaround: restart abapGit before clicking "Refresh".

Request: is it possible to not restart abapGit, by executing the APACK class in an RFC session so that to always consider the latest APACK class?

sandraros:
Fix #5284

### DIFF
--- a/src/apack/zcl_abapgit_apack_helper.clas.abap
+++ b/src/apack/zcl_abapgit_apack_helper.clas.abap
@@ -28,7 +28,7 @@ CLASS zcl_abapgit_apack_helper DEFINITION
       IMPORTING
         !iv_clsname        TYPE seometarel-clsname
       RETURNING
-        value(ro_provider) TYPE REF TO object.
+        VALUE(ro_provider) TYPE REF TO object.
   PROTECTED SECTION.
   PRIVATE SECTION.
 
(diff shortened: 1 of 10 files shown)


mbtools:
Hi @sandraros, function modules don't work with the standalone version (since #4814). Closing

mbtools:
Hi @sandraros, function modules don't work with the standalone version (since #4814). 

PS: APACK is a dying feature... I'm getting serious replacing it with the real deal.

sandraros:
So, I guess we won't have the possibility anymore to show the latest modifications of any repository in `ZABAPGIT` (except the ones of abapGit itself), Is there a workaround that is used by other abapGit projects? (like users can simply display the diff of `changelog.txt`)

mbtools:
APACK is there and can be used. Any replacement will allow for the same features (and more).

sandraros:
Issue closed and PR abandoned.

Issue #5294, opened 2022-01-27
Pulling structure (TABL) with missing dependency


FreHu:
Repo setup:
- structure with field referring to data element
- data element

Initialize the repo and pull just the structure (I am simulating a missing dependency during a larger pull)

![image](https://user-images.githubusercontent.com/5097067/151344619-39fa6000-b2c0-4c6f-80d6-dfc8189590d3.png)

Then it says "no active version found", which is true but  the inactive object icon is missing. 
![image](https://user-images.githubusercontent.com/5097067/151344659-2c063131-e9e5-47db-8f25-25b58495dd90.png)

The object exists on the system in a partially active state:
![image](https://user-images.githubusercontent.com/5097067/151344816-d1d616b3-941d-43e7-98ae-2a694183bf9b.png)

Abapgit makes it look like the object will be created if you just pull, but if you keep repeating the pull, the DTEL will be pulled but you will not get out of the situation without activating the TABL manually.
![image](https://user-images.githubusercontent.com/5097067/151345066-3bc37bac-aeb4-46c5-8668-a4c8f73a298d.png)

There are a few improvements which can be made to this flow:
- show partly active objects like inactive objects
- Get rid of the duplicate message
- "no active version found" should point out that an inactive version was found and you need to activate it. Normally abapgit would activate or show an activation popup automatically
- the activation errors have the best information about the problem, but you will only see this popup once

mbtools:
That's a bit messy indeed. 

1. Inactive check/icon

For partially active table, what status is returned in `ZCL_ABAPGIT_OBJECTS_SUPER->IS_ACTIVE`?

2. Duplicate message

TABL is in DDIC and LATE step. So it's trying twice to activate (since there might be missing dependencies in the first step). That's why we see the error twice. Solution might be a simple remove duplicates.

3. Activation errors

Those should be merged into the AG log. Needs to pass `ii_log` to `zcl_abapgit_objects_activation=>activate`.

FreHu:
the is_active check is not reached because serialize throws due to missing `dd02v`
![image](https://user-images.githubusercontent.com/5097067/151356611-fec94700-23c1-45de-af59-f9ec883ee6ac.png)

FreHu:
Based on the code in `zcl_abapgit_objects=>serialize`, this should be a general issue, the `is_active` logic will always be skipped if serialize throws.
```abap
 CREATE OBJECT li_xml TYPE zcl_abapgit_xml_output.

[...]

PR #5302, opened 2022-01-31
Refactor: Activation log

mbtools:
- Merges the activation log into the regular pull log
- Avoid popup in case of activation errors (good for batch runs)
- Handle errors when updating where used list
- Removes duplicate messages

image

Ref #5294


FreHu (approved):
tested, the errors are appended and the gui popup is gone

PR #5303, opened 2022-01-31
TABL: Properly show inactive objects

### LINKED ISSUE #5294: Pulling structure (TABL) with missing dependency

mbtools:
Current `is_active` check is insufficient for DDIC objects and can't detect "new" or "partically active" objects.

Introduces a new `is_active_ddic` method in `zcl_abapgit_objects_super` which should be used by all DDIC types (follow-up). 

image

Closes #5294

### DIFF
--- a/src/objects/zcl_abapgit_object_tabl.clas.abap
+++ b/src/objects/zcl_abapgit_object_tabl.clas.abap
@@ -915,7 +915,7 @@ CLASS zcl_abapgit_object_tabl IMPLEMENTATION.
 
 
   METHOD zif_abapgit_object~is_active.
-    rv_active = is_active( ).
+    rv_active = is_active_ddic( ).
   ENDMETHOD.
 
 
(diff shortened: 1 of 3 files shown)


FreHu:
After the initial pull, I don't see the icon yet, but after refresh it is there.
image
image

I think this is acceptable, the result is much better already.

mbtools:
I will fix the refresh in another PR

Issue #4861, opened 2021-07-01
Enablement of JSON based file format processing for object types in abapGit


ThomasPloski:
**Motivation**
ABAP repository objects are stored on development systems and not as regular text files as many other programing languages. Since there are a lot of development tools dealing with ABAP development objects, the ABAP file formats initiative is about to provide a common and transparent definition of how to store ABAP repository objects in files.

The design of the ABAP file formats covers the following principles: 

- The file content is complete and can be used for the sharing and distribution of ABAP repository objects. 
- The file content is human-readable and the structure is as close as possible to what users know from their IDE. 
- The files don't contain metadata like dates, usernames, last editor, system specific information, plus no obsolete or redundant information.

Further information: https://github.com/sap/abap-file-formats 
Technical Specification: https://github.com/SAP/abap-file-formats/blob/main/doc/file_names.md

Since abapGit is one of such development tools handling ABAP development objects, I would like to discuss with you how to enable the current abapGit core for the JSON based file formats. 

For that reason, I did a small PoC for the abapGit version in SAP BTP ABAP Environment (aka Steampunk) to get a first feeling about what is needed to enable JSON based file processing in abapGit. Out of this the following aspects I would like to point out:

**Exchange file abstraction** 
The abstraction for the current exchange format file is realized by the following two classes: 
- CL_ABAPGIT_XML_INPUT implements IF_ABAPGIT_XML_INPUT
- CL_ABAPGIT_XML_OUTPUT implements IF_ABAPGIT_XML_OUTPUT 

There is no consistent usage of the above-mentioned interface types. In particular, all methods in the Interface IF_ABAPGIT_OBJECT use the class reference types as part of the method signature. This means, that every object type handler operates on the class reference type(s). In order to introduce a new JSON based file format abstraction also being compatible with the current one, a new common abstraction layer is required to handle both.

![image](https://user-images.githubusercontent.com/39330834/124167046-d7cd1c00-daa3-11eb-84f8-7f96aa4e4120.png)

As a first step, the new interface IF_ABAPGIT_EXCHANGE_FILE should be defined with needed public methods and be implemented by the current XML change file representation class. Subsequently, all reference type information should be changed to interface type in all method signatures as well as in the processing part. Having this new abstraction level in place, additional file formats can be added easily without breaking the current implementation. Only the file import processing part of the abapGit framework needs to decide which implementing class needs to be called to create an object for a given change file.

**JSON file transformation**
_JSON Parser_ 
[...]

PR #5313, opened 2022-02-01
Detect JSON files during pull

mbtools:
Fixes JSON file detection (#5305)


albertmink:
Pull results in
image
It is not quite there yet.

albertmink:
Might be somewhere around here where there is a hard-coded reference to file ending `XML`.
https://github.com/abapGit/abapGit/blob/main/src/objects/zcl_abapgit_objects.clas.abap#L676

The method `read_xml` does what it is supposed to do 😄 
image

mbtools:
this will take several steps to refactor since i can't test it myself...

albertmink:
Is it necessary to run an ABAP System with support for CHKO object? It felt like this part of deserialization in AbapGit is not related to specific CHKO API.

mbtools:
it isn't related. but on lower releases AG will filter object types that are not supported before you even get to this point.

you can see this on "Utilities > Debug Info":

image

The new AFF object types are not shown on my system.

Anyway, I think I got it with the next commit

larshp:
there is also relevant metadata in the json files, so this needs refactoring sometime, if it works

mbtools:
absolutely. it needs to be properly modeled (see #4861)

mbtools:
@albertmink give it a try again

albertmink:
Thanks for your immediate support.
Now, it is the missing devclass
image

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -656,9 +657,10 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
             lv_path = <ls_result>-path.
           ENDIF.
 
+          ls_item-devclass = lv_package.
```
larshp, line 660: hmm, moving this statement seems like a hack, I guess its not the way the xml gets the package

mbtools:
It's kind of a mistake in your deserialize method where it uses ms_item-devclass instead of iv_package. I made a small change so ms_item-devclass is set to the same package. 

FYI, I see you check the object type in several places. That's not necessary since the class will only be called if the obj_type matches the classname.  You can remove all of these and probably won't need to redefine the methods much:

```abap
    IF ms_item-obj_type <> 'CHKO'.
      RETURN.
    ENDIF.
```

src/objects/zcl_abapgit_objects.clas.abap
mbtools: the ls_item is passed to the constructor (ms_item) which includes devclass. it's confusing if this does not match the package passed to deserialize method. 🤷

mbtools: https://github.com/abapGit/abapGit/blob/main/src/objects/aff/zcl_abapgit_object_common_aff.clas.abap#L153

albertmink:
👋 No error message after pulling, so this might work for the JSON files.
Unlucky, that we didn't tested this behavior when working on #5163. Relaying on unit test was not enough.

albertmink:
It is working as expected for CHKC and CHKO.
I could deserialize them into the ABAP system, requires #5320 #5319

PR #5314, opened 2022-02-01
Update object descriptions - Part 1

mbtools:
Standardize with "abapGit - " prefix


FreHu:
not sure how worthwhile, but a description regex would not be a difficult abaplint rule

mbtools:
didn't take that long. yes, a lint rule is the goal.

FreHu (approved):
for future reference,
https://github.com/abaplint/abaplint/blob/main/packages/core/src/utils/name_validator.ts
is a good starting point

Issue #5263, opened 2022-01-10
Refactor: Replace `delete_tadir` flag


mbtools:
Deleting `tadir` entries should be handled the same way as inserting them.

Therefore, the `delete_tadir` metadata flag (`zif_abapgit_definitions=>ty_metadata`) should be removed and replaced with method `zcl_abapgit_objects_super=>tadir_delete` (to be called from `zcl_abapgit_object_<type>->delete`).

PR #5326, opened 2022-02-04
Enable AFF to delete TADIR entries

albertmink:
We need to discuss the error handling here and in #5320


schneidermic0 (approved):
Looks good to me. Thanks

mbtools:
As mentioned in #5263, we would like to get rid of the "tadir delete" flag and replace with a `tadir_delete` method (same as `tadir_insert`). 

With this PR, the mapping between item and trkey is hidden in SAP code. Is this necessary? or is it just a move-corresponding?

albertmink:
Ahh, it was naive coding - I just called the methods most familiar to me.
The exporting parameter of
https://github.com/abapGit/abapGit/blob/114a202812c8938ba1a0891d447cbfc59d802642/src/objects/aff/zcl_abapgit_object_common_aff.clas.abap#L54
should be sufficient and more transparent.

PR #5343, opened 2022-02-09
Persist "Folders" option in Repo View

### LINKED ISSUE #5341: Repo view: Save View > Folders in user persistence
Other View settings are already stored. So I think it would be reasonable to save this too.
grafik

Here
https://github.com/abapGit/abapGit/blob/baa09083e6f6ccb5e898eba44a16286d3563dc06/src/ui/zcl_abapgit_gui_page_repo_view.clas.abap#L1284

something like this:
https://github.com/abapGit/abapGit/blob/baa09083e6f6ccb5e898eba44a16286d3563dc06/src/ui/zcl_abapgit_gui_page_repo_view.clas.abap#L1273

g-back:
closes #5341


christianguenter2 (approved):
👍

PR #5357, opened 2022-02-16
JUMP: Switch execution order - first specific then generic

### LINKED ISSUE #5335: SSFO: Jump doesn't work
https://github.com/abapGit-tests/SSFO.git

christianguenter2:
With this change execution order of the jump logic is switched. At first we execute the object specific `JUMP` and if that's not successful we execute the generic handler. Therefore we introduce a new returning parameter `RV_EXIT` which controls whether the generic handler should be skipped or not.

fixes #5335

### DIFF
--- a/src/objects/zcl_abapgit_object_shi3.clas.abap
+++ b/src/objects/zcl_abapgit_object_shi3.clas.abap
@@ -323,9 +323,9 @@ CLASS zcl_abapgit_object_shi3 IMPLEMENTATION.
         jump_se43( ).
       WHEN 'GHIER'.
         jump_sbach04( ).
-      WHEN OTHERS.
-        zcx_abapgit_exception=>raise( |Jump for type { ls_head-type } not implemented| ).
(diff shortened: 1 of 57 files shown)


src/objects/zcl_abapgit_object_shi3.clas.abap
```diff
@@ -323,10 +323,10 @@ CLASS zcl_abapgit_object_shi3 IMPLEMENTATION.
         jump_se43( ).
```
mbtools, line 323: `rv_exit` should go into the two `when` branches.

mbtools (changes requested):
just a small change. looks good otherwise

src/objects/zcl_abapgit_object_shi3.clas.abap
christianguenter2: good catch. Fixed.

mbtools (approved):

Issue #3382, opened 2020-05-21
New Local and Remote Views


mbtools:
I always wanted to see the exact serialization of local and remote objects, of course nicely rendered in abapGit. I have enhance the "Diff page" to now be able to view just the local or just the remote side of files. It's a bit of a bigger change, but it works quite well... and I have added some other features as you can see below. 

- New items in Repo > Advanced memu "Show Local" and "Show Remote"

These will display all files included in the local or remote repositories (Note: If you do that with abapGit, it will take some time to render since the repo is so large!)

- If files are shown in Repo, then the item state icons on the right side become links to view the local or remote version of this file

![Image6](https://user-images.githubusercontent.com/59966492/82516795-f8dee180-9b1b-11ea-9d43-74fc23fadbdc.jpg)

- The new local and remote views also include statistics for number of lines, sloc, and size (same as in Github)

- At the top also the totals are displayed

- Also includes minor correction to map blank "last changed by" to "unknown"

- Since you can display a complete repo in these new abapGit views, you can quickly search for anything using Ctrl+F, or cut&paste any code easily. And it's all there without having to launch any ABAP tools.  

![Image10](https://user-images.githubusercontent.com/59966492/82516851-11e79280-9b1c-11ea-8047-f8c0deb1c1d2.jpg)

If you like to give it a try, use https://github.com/mbtools/abapGit/tree/local-remote-view and **turn on experimental features.** in your settings.

Let me know what you think!

PS: It's worth repeating since I know many will try to view the complete abapGit repo this way... It will take some time to display. The AG repo is ~85k sloc and the generated HTML page becomes huge (32MB due to syntax highlighting). My GUI went into "not responding" for about 2 min, but it did eventually show everything correct and navigating on the page was seamless.

FreHu:
Hi,

first of all, thanks for taking the time to make these extensive contributions!

My notes:

*   The item state icons are great, if a file is new or deleted I don't really need to view it all green or all red.
*   Totals are great to know. I have a personal gripe with _sloc_ though_. "_Source lines of code" sounds like a case of the [RAS\_syndrome.](https://en.wikipedia.org/wiki/RAS_syndrome) I'd go with _20 lines (15 code)_ but you're trying to match github and they use _sloc_ so this is just personal preference.
*   The last changed by improvement is nice, different values for "unknown" clutter the filters
[...]

PR #5358, opened 2022-02-16
more improvements to repo view

### LINKED ISSUE #5355: Repo view path name layout
I must say I really do not like the path layout introduced in #5323 (sorry @FreHu, no offence, really appreciate all the other great stuff you've done!). Is it a huge effort to make this optional?

Personally I prefer higher information density and it is now way too spread out and I can see less. I now quite literally have half the info I had before on my screen, my view of 12 objects has been reduced to 6:
image

Onto the subjective opinion part: visually I also find it awkward with everything crammed to the left with oodles of space on the right, I am almost always primarily interested in the objects and much prefer them on their own without the fluff in between. I find it makes the object list harder to scan through.

FreHu:
resolves #5355 
- status is now on the left
- status is now clickable to show diff
- to unite different display modes, show parent status for object with number of changes
- object type now contains the icon instead of a separate column
imageimage
imageimage


mbtools:
maybe try to put the status icons right in front of the file name. 

at one point, I had turned these icons into links to the file (see https://github.com/abapGit/abapGit/issues/3382). maybe, they should link to the diff.

FreHu:
image
image

FreHu:
- updated screenshots in pr description
- the icon has to be the link - or for grouped changes there is nowhere to place the diff link

FreHu:
image
fixed

pokrakam:
Having the paths offset makes the main list easier to read. My personal preference is tabular, but I know people like multiline lists. The offset using the icons is a good compromise, good work 🎯

pokrakam (approved):

sbcgua:
Imho, not that I'm a big of the improvement this time :(

image

- (1) Issue with packages, probably needs some more spacing 
- (2) changed by looks very tight, is the border around user needed? Imho quite cluttered. Maybe leave just the icon ?
- In my opinion line are too tight and the visual separation became worse
- diff on status - sorry, but this is really unobvious. I can guarantee that it would be a confusion for 90% of people, both current users and new ones
- Minor: probably I'd experiment with file count (by the way is it file count or diff count ?? also unobvious - I didn't look through the code, it is my "user experience"), probably make them grey (as they distract from the main content) and smaller to match status font

FreHu:
> * (1) Issue with packages, probably needs some more spacing

I will fix this

> * (2) changed by looks very tight, is the border around user needed? Imho quite cluttered. Maybe leave just the icon ?

> * In my opinion line are too tight and the visual separation became worse

Border could be removed as well as some slight padding, or maybe up the line height a little bit more?

> * diff on status - sorry, but this is really unobvious. I can guarantee that it would be a confusion for 90% of people, both current users and new ones

Maybe rename "Status" to "Diff"?

> * Minor: probably I'd experiment with file count (by the way is it file count or diff count ?? also unobvious - I didn't look through the code, it is my "user experience"), probably make them grey (as they distract from the main content) and smaller to match status font

It's the file count. This was already there if you turned off file names or enabled folders (then it shows file count in that folder). I think the diff count is not available to show in this page, only in diff/patch.

larshp:
@FreHu check conflicts

pokrakam:
> Maybe rename "Status" to "Diff"?

That makes more sense to me 👍

sbcgua:
image

what about:
- status not clickable
- instead, click on object name would show diffs (which is also hinted in column name)
- the "open tool" command is displayed on hover separately
- file count is lighter to reduce distraction from the main content
- no borders for user and transport (they really just add too many lines)
- also icons moved to the top (it is quite frequently that icon + text does not fit into column somehow and split into 2 lines which is annoying)
- tooltip of the file count may show file list (though I'm not a big fan of it, maybe it is better to rerender into file view or file view for this item only ... or ... I don't know, it is kind of useful info ... but really not that critically useful)
- finally I still think that object item in the line adds mess and is not that necessary. Though highlighting packages is good ... needs some more experiments

image

sbcgua:
... on the other hand, if the repo view and stage is the same UI, then icon may be dynamically replaced by checkbox which will visually highlight the change of mode ... which could be not bad ...

FreHu:
> status not clickable
> instead, click on object name would show diffs (which is also hinted in column name)
> the "open tool" command is displayed on hover separately

I haven't been active for a while and looking at it now, I couldn't find the diff either. I like this the most so far, I'll try it.

FreHu:
Instead of the hover link,
image
what about using these navigation arrows?
image

FreHu:
Or having the icon on the right so you can better see what you're clicking, because clicking a folder is a different action than an ADT link.
image

This leaves the folder name unused, it could open the diff for a whole folder same as clicking the object opens a diff for all the parts.

FreHu:
The icons on the right turned out quite bad, i put them back. Current state:
image

sbcgua:
arrows: ... interesting, but on the other hand they are small (to click) and textless (thus unclear) **imho** ... and will be far from object name. The links near the object name would ensure less mouse movements, less potential mistakes in clicking the wrong row and potentially extendable with other commands if needed.

Looking at my sketch after 2 days, I'd say the button is too bulky ...

image

`vertical-align: text-top`

Or maybe:

image


P.S. "click to see diff" - i'd suggest "click object name to see diff", otherwise it supposes clicking the header. And maybe getting this hint out of `<a>` because it is a hint and not a column name.

P.P.S what do you think about removing hover underline for links ? It kind of looks superfluous having that there is a hand cursor and a lot of horizontal lines and that the menu links are not underlined.

P.P.P.S and also too much blue color. it because everything is clickable which is OK. But so we loose the power of color differentiation. Hmmm

sbcgua:
P.P.P.P.S And maybe mark modified line with the border instead of highlight to avoid double highlight (on hover)

image

mbtools:
I'm not a fan of "features on hover" (showing some info on hover is ok). Changing the "click on object" behavior will cause confusion and complaints. How about changing the "adt" user setting to three options?
- open in ADT
- open in classic editor
- show diff

To be consistent it should work the same on repo view as well as staging.

sbcgua:
> I'm not a fan of "features on hover"

Why ? :)
My arguments are:
- there are secondary commands
- if they are always displayed, they clutter the view (maybe it is OK or maybe not - the threshold is quite personal of course)
- user always moves mouse towards the object of interest
- so the new options appear in this context

> Changing the "click on object" behavior will cause confusion and complaints

Yes, but diff on object is better. Because:
- the more frequent commands should be more accessible
- diff is more frequent than "open editor" (... we didn't check ... but it's highly probable)
- larger click areas are more accessible
- object name is most visible and large click area

We want to improve the tool, right ?
Control via config is also OK though. I just believe that there should be most optimal default and also that too many configs are difficult to navigate and study. It's just natively so. So adding configs is not always to the better. Look at VSCode for example, how many params are there and how many does the average user know? So I'd vote for "optimal zero config" when it is possible. But, yeah, config is OK, imho AG config is not too overcrouded yet :)

mbtools:
Why? Accessibility 💯  (for example https://accessuse.eu/en/Content-hover-focus.html)  

We don't have a framework that enforces accessibility standards. So it's up to us to avoid pitfalls. There's much more in this regard. Let's not make it worse 😃 

Of course we want to make it better. Always. But change based on what "we" think doesn't necessarily match what the majority of users would say. My mother and wife both complained this week how they hate updating anything since "it always changes everything". 🤣 I use the "open editor on click" since I edit a lot and don't like to type so many tcodes 😉. So in this case it's not just "making it better" but also removing a very familiar and potentially widely used feature. That's why a soft change with an option to switch to the old behavior makes sense to me.

I certainly agree that we don't want many settings and changing things without a new setting is always preferred. 

The current UI supports both "edit" and "diff". I don't see the "better" with the above ideas. 

Let's do this in smaller steps: 
- moving "state" column
- format of "user" and "transport" columns
- click behavior

sbcgua:
https://accessuse.eu/en/Content-hover-focus.html: well this particular one is a bit arguable.
- the commands are not "popup content", they are compact and shown very very close to the pointer area
- they don't disappear (the hover is for the whole row) and in general match quite well to the "Accessibility tips" section of the link

Anyway, this is why they invented anonymous statistics (which I suggested somewhere already) and A/B test :) E.g. you wrote `I use the "open editor on click" since I edit a lot` - that is an important point for the statistics, +1 for the editing. E.g I wouldn't much care if the feature was removed at all, I use it so rarely compared to diff (which is +1 for diff). Would be great to have some realistic distribution of the cases. But probably, even if there is a user consent to send anon stats from AG, it would be very difficult technically :( ... though maybe slack community could help with this, it should be possible to create votings there, hmmm ...

Small steps, sure, totally ok with it, it is just a good place to discuss it because there are real screenshots and code. I hope Frederik doesn't hate me for my comments :)

mbtools:
I'm with you when it comes to usage stats. A cloud version of AG could collect all the stats we want. Lars just has to make AG run in the transpiler. haha.

It's a good discussion. I'm not completely opposed to the hover. We can refactor when I'm old and need keyboard nav and a screen reader 👴 . I'm more concerned about removing the edit (or anything else useful). Does it have to be either or? Maybe show several icons on hover? edit, diff, check, activate?

mbtools:
How about enhancing the state icons (I had coded something like it in https://github.com/abapGit/abapGit/issues/3382)?

Click on local icon -> show object (local files) in AG 
Click on remote icon -> show object (remote files) in AG
Click on delta icon -> show diff

Click on object or file name -> jump to (edit) object 

Something like this:

image

pokrakam:
+1 for Edit. Most of the time I click to edit, and the smaller link on the right to show the occasional Diff is perfect for the way I work with it.
As an ADT user most of my Diff-ing is done on GitHub or ADT. Local history is one of ADT's most underrated features IMHO.

sbcgua:
- added a vote to slack channel to collect some stats, let's try to make a more educated decision :)
- icons: it's OK to assign some functions to icon but: A) icon are small to click B) icons are unobvious to click at the first place (when without text). So I'd say it is OK to leave star icon without text tag, or refresh (although the refresh will be inconvenient to click), because these symbols are so typical. But for diff it will be unobvious. Show local/show remote are OK though. It is just that Diff deserves a larger click spot.

FreHu:
>  I hope Frederik doesn't hate me for my comments :)

Certainly not, I want to get to something that works well. I just have little time and trouble replying to everything at once :).

What I wanted originally was to unite repo and stage. On the staging page, clicking the filename opens the diff. There is no ADT link on the staging page.

I am also taking inspiration from the fiori guidelines, even though abapgit doesn't have to follow them completely. The inspiration for arrows was taken from our apps, although ours are bigger.

image

Clickable rows are something where abapgit deviates, for a single action like navigate or diff, the recommendation is to place it on the right.
https://experience.sap.com/fiori-design-web/action-placement/#actions-in-forms-and-table-rows

Or something like this if there were many more actions (open on remote, undo changes?)
https://experience.sap.com/fiori-design-web/smart-link/ 

Checkboxes for selection should be on the left, so I like the idea of changing the icon for a checkbox eventually.

I like this left border as well as getting rid of the underlining on links.
image

larshp:
@FreHu close?

mbtools:
Sorry for not giving feedback earlier. There are a lot of good things in this PR. Let's try to get this done. 

Things I don't like:

1. Click on object opens diff 

I'm ok with this if we get also a button/link next to it for opening the editor (could be on hover).

2. Status column on left 

Two issues with the status column:
a) it's only filled in some cases giving the table an unpleasant look  
b) it's not the key column (which is type/name and always filled)

We should keep the status column on the right side.

FreHu:
Sorry, I left SAP and left this one unfinished. I don't have the hardware or the motivation to continue, and I left this open for so long I don't remember what's in here anymore.

larshp:
@FreHu no worries, thanks for all

mbtools:
Please keep the branch until I downloaded it. TY 

All the best for your next adventures!

mbtools:
Branch code for reference:
[abapGit-repo-view-followup.zip](https://github.com/abapGit/abapGit/files/9655821/abapGit-repo-view-followup.zip)

PR #5362, opened 2022-02-16
Don't fail if a dictionary object is activated with warnings

raaleksandr:
DD_MASS_ACT_C3 fm returns 4 in case of warnings during data dictionary objects activation.
Warning should not interrupt the process.
So I offer to consider rc > 4 as error but not rc > 0.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,13 +2,12 @@ name: build
 
 on:
   push:
-  pull_request:
+    branches:
+      - main
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     steps:
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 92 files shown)


mbtools:
Good point. However, we should not completely ignore the warnings but bring them to the abapGit log.

So keep `lv_rc > 0` for the `add_errors_to_log` and use `lv_rc > 4` for raising the exception. Then in `add_errors_to_log` also keep the warnings. Probably good to rename it to `add_messages_to_log`.

raaleksandr:
See how warnings during pull will look like
warnings_will_look_like

mbtools:
Nice. We can finally see the little issues (warnings) that exist among the DDIC objects!

mbtools (approved):
👍

larshp:
yea, for enhancement category, enable https://rules.abaplint.org/tabl_enhancement_category/ to keep it all fixed 👍

PR #5369, opened 2022-02-19
Fix DDIC errors related to ref to class/interface

### LINKED ISSUE #5366: Issues concerning DDIC objects with reference types
Activation issues for these DDIC objects:
- Data Element of type reference
- Structure with component being a data element of type reference
- Table Type with line type being a data element of type reference

# To reproduce
- New Online this repository: [branch "issue-5366"](https://github.com/sandraros/test-abapgit/tree/issue-5366)
- Pull → "No active version found for TTYP ZTESTTABLETYPE", "Row type OBJECT is not active or does not exist", "Table Type ZTESTTABLETYPE could not be activated":
  image

sandraros:
Fix #5366

### DIFF
--- a/src/objects/zcl_abapgit_object_dtel.clas.abap
+++ b/src/objects/zcl_abapgit_object_dtel.clas.abap
@@ -212,7 +212,7 @@ CLASS zcl_abapgit_object_dtel IMPLEMENTATION.
     IF iv_step = zif_abapgit_object=>gc_step_id-ddic AND ls_dd04v-datatype = 'REF'
       AND is_abapclass_or_abapinterface( ls_dd04v-domname ) = abap_true.
 
-      ls_dd04v-rollname = 'OBJECT'.
+      ls_dd04v-domname = 'OBJECT'.
     ELSEIF iv_step = zif_abapgit_object=>gc_step_id-late AND ls_dd04v-datatype <> 'REF'.
       RETURN. " already active
     ENDIF.
--- a/src/objects/zcl_abapgit_object_dtel.clas.abap
+++ b/src/objects/zcl_abapgit_object_dtel.clas.abap
@@ -209,11 +209,11 @@ CLASS zcl_abapgit_object_dtel IMPLEMENTATION.
                   CHANGING cg_data = ls_dd04v ).
 
     " DDIC Step: Replace REF TO class/interface with generic reference to avoid cyclic dependency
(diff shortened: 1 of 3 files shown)


sandraros:
**Please put it on hold**, as this PR will conflict with #5370. I think the latter is more correct on data element processing as it's also testing whether the referenced type is a class or an interface, so I'd better adapt this current PR.

Below is part of what I will change in the PR for table types: I would do the same as for data elements in #5370 ; by the way, here is a smaller [repository](https://github.com/sandraros/test-abapgit/tree/issue-5365-ttyp-dtel-ref-to-data/src) to test only table types (branch `issue-5365-ttyp-dtel-ref-to-data`):
- New method declaration:
  ```abap
    METHODS is_ref_to_class_or_interface
      IMPORTING
        !is_dd40v        TYPE dd40v
      RETURNING
        VALUE(rv_result) TYPE abap_bool .
  ```
- New method implementation:
  ```abap
  METHOD is_ref_to_class_or_interface.

    IF is_dd40v-rowkind = 'R'
        AND ( is_dd40v-reftype = 'C'
           OR is_dd40v-reftype = 'I' ).
      rv_result = abap_true.
    ENDIF.

  ENDMETHOD.
  ```
- Method `zif_abapgit_object~deserialize` 
  ```abap
    IF iv_step = zif_abapgit_object=>gc_step_id-ddic AND is_ref_to_class_or_interface( ls_dd40v ) = abap_true.
      ls_dd40v-rowtype = 'OBJECT'.
    ELSEIF iv_step = zif_abapgit_object=>gc_step_id-late AND is_ref_to_class_or_interface( ls_dd40v ) = abap_false.
      RETURN. " already active
    ENDIF.
  ```

mbtools:
please add the 5365 test case to https://github.com/abapGit-tests/DTEL_with_REF

mbtools:
yes, let's check in all cases for `ref to clas/intf` properly. thx

mbtools:
i took care of the conflict in `dtel` and minor addition for `tabl`. you just have to adjust `ttyp`.

sandraros:
In fact, the current fix solves the issue, so I propose to merge it. I'll take care of `ttyp` in another issue with brand new test case to reproduce before fixing.

sandraros:
Test case added to https://github.com/abapGit-tests/DTEL_with_REF
Checked again that it solves (test before/after)

mbtools (approved):
👍

PR #5350, opened 2022-02-14
DTEL: REF TO DATA for DDIC element support

raaleksandr:
When trying to import a project with data element of type 'REF TO DATA' contained in a DDIC structure we got error.
This is fix for the issue.
See issue #5347 for details.

PR #5370, opened 2022-02-20
Fix data elem type ref to class in same repository

sandraros:
To complete #5350. Solution is not very clean, suggestions welcome, thanks!


mbtools (approved):
this is better. thx

PR #5371, opened 2022-02-20
Activate transformations before programs

### LINKED ISSUE #5365: pull fails on XSLT with tt:call-method, if classic program refers to this class
This is another [minimal test case](https://github.com/sandraros/test-abapgit/tree/issue-5365) (branch `issue-5365`) which concerns issues related to XSLT activation. It's almost the same as #5262, but I think it's worth documenting them in separate issues so that we can have all test cases if we agree on a fix (I am working on it). For information, we also have a third test case, which is https://github.com/abap2xlsx/abap2xlsx/issues/993 (with different abapGit behavior, I am also trying to build a minimal test case for it).

My system is ABAP 7.52 kernel 753 SP 500 (before the kernel indicated in the note 2859159 - link below).
I have abapGit 1.114.0, latest pull.

Marc has nicely explained the general problem [here](https://github.com/abap2xlsx/abap2xlsx/issues/993#issuecomment-1040918350), so I just copy/paste it:
> This project contains XSLT and if those objects don't activate properly, then nothing else will work (due to Ddic and code dependencies).
> Therefore, please check SAP note https://launchpad.support.sap.com/#/notes/2859159 first and make sure you have the right kernel version running.
> Secondly, run the very latest version since we continuously improve things like #5303 which shows inactive TABL objects correctly (note: similar handling of other DDIC types is still missing).
> To have a reproducible case, uninstall the complete repo and install it again. Attach output of Utilities > Debug info and screenshot after initial pull.

# To reproduce
- Install the fix #5369 (otherwise there will be some errors not related to transformation)
- New Online https://github.com/sandraros/test-abapgit → Switch to branch `issue-5365`:
  image
- Pull → "Activation cancelled. Check the inactive objects. ":
  image
- Pull again does the same as above

sandraros:
Fix #5365 

To reproduce the error, and explanations:
- A program refers to a class, which refers to a transformation, which refers to a class.
- During the activation step "ABAP":
  - the classes are written directly with active status although they have syntax errors (like using an attribute of a class referred by a Data Element or a DDIC Structure component),
  - then the inactive objects (programs and transformations) are activated. The transformation which refers to a class may be activated even if the class has a syntax error. The program which uses a class with syntax error cannot be activated.

Fix:
- activate the transformations before the programs

### DIFF
--- a/src/objects/core/zcl_abapgit_objects_activation.clas.abap
+++ b/src/objects/core/zcl_abapgit_objects_activation.clas.abap
@@ -65,7 +65,7 @@ CLASS zcl_abapgit_objects_activation DEFINITION
         !ii_log TYPE REF TO zif_abapgit_log
       RAISING
         zcx_abapgit_exception .
-    CLASS-METHODS add_errors_to_log
+    CLASS-METHODS add_errors_and_warnings_to_log
       IMPORTING
         !iv_logname TYPE ddmass-logname
         !ii_log     TYPE REF TO zif_abapgit_log
@@ -164,9 +164,12 @@ CLASS zcl_abapgit_objects_activation IMPLEMENTATION.
       ENDIF.
 
       IF lv_rc > 0.
-        add_errors_to_log(
+        add_errors_and_warnings_to_log(
           iv_logname = lv_logname
           ii_log     = ii_log ).
+      ENDIF.
+
+      IF lv_rc > 4.
         zcx_abapgit_exception=>raise( 'Activation cancelled. Check the inactive objects.' ).
       ENDIF.
 
@@ -183,6 +186,8 @@ CLASS zcl_abapgit_objects_activation IMPLEMENTATION.
 
   METHOD activate_new.
 
+    DATA: lt_objects TYPE TABLE OF dwinactiv.
+
     IF gt_objects IS INITIAL.
       RETURN.
     ENDIF.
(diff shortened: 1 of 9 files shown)


src/objects/core/zcl_abapgit_objects_activation.clas.abap
```diff
@@ -196,7 +198,21 @@ CLASS zcl_abapgit_objects_activation IMPLEMENTATION.
 
     ELSE.
 
-      activate_old( ii_log ).
+      " Activate transformations first
```
mbtools, line 201: this object-type logic is what we want to avoid.

mbtools (changes requested):
The goal is to make it work without this (for systems without the note or kernel patch).

src/objects/core/zcl_abapgit_objects_activation.clas.abap
sandraros: Is, adding a new activation step 1.5-XSLT between 1-DDIC and 2-ABAP, fine?

mbtools:
Let's finish #5369 first. Then I want to test with your issue-5365 repo and come back to this.

I renamed the test repo to https://github.com/abapGit-tests/DDIC_with_REF. Just push the complete issue-5365 to it so we have the full test case for our CI runs.

src/objects/core/zcl_abapgit_objects_activation.clas.abap
sandraros: Should I propose a commit to have a new activation step 1.5-XSLT between 1-DDIC and 2-ABAP, or any other solution you'd prefer?

sandraros:
I prefer to close this PR, first do a fix on #5392, and check the situation after this fix.

PR #5390, opened 2022-03-02
[WEBI] - Sort internal tables

### LINKED ISSUE #5389: [WEBI] - Service definition diffs due to missing sorting order
Sorting order seems to be a common issue for diffs even though there are no technical differences. We have come across WEBI( Service Definition ) having different sorting order returned by function module 'WEBI_GET_OBJECT' for TABLES parameter **pwssoapprop**.

There are several internal tables filled by the above function modules.
image

image

Diffs due to different sorting order between S/4HANA and non S/4HANA
image

Currently only one internal table is sorted after the function module call.
image

**Recommended Solution**:
Sort all the internal tables which are filled by the function module immediately after the function module call.

yellappam:
Function module WEBI_GET_OBJECT returns several internal tables. Sometimes, the sorting order of these internal tables is different in each system ex: S/4HANA and non-S/4HANA. This fixes #5389.

### DIFF
--- a/src/objects/zcl_abapgit_object_webi.clas.abap
+++ b/src/objects/zcl_abapgit_object_webi.clas.abap
@@ -56,6 +56,9 @@ CLASS zcl_abapgit_object_webi DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
       RAISING
         zcx_abapgit_exception
         cx_ws_md_exception.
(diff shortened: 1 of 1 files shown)


mbtools:
Good find. Thx

mbtools (approved):

PR #5393, opened 2022-03-06
Create empty classes/interfaces before DDIC

### LINKED ISSUE #5392: pull fails on program using ddic referencing class or interface
# To reproduce
- new online
  - repository: https://github.com/sandraros/test-abapgit
  - branch: `ddic-ref-to-objecttype-prog-using-ddic-object`
- pull
  image

# Reason
- abapGit first pulls DDIC then programs
- While pulling DDIC objects with reference to class or interface, abapGit replaces the class or interface with generic `OBJECT`
- When activating programs, use of class or interface members from DDIC objects does a syntax error because DDIC objects refer to generic OBJECT which doesn't have any member.
- Workaround: rewrite ABAP programs so that to upcast references of DDIC objects (?=) then abapGit can pull

sandraros:
This PR changes the order of deserializing objects for classes and interfaces. It greatly simplifies the processing of DDIC objects referencing classes or interfaces.

- An `early` step is introduced that runs before the `ddic` step. 
- During the `early` step, classes and interfaces that do not exist yet are created (without any code). 
- Therefore, during the `ddic` step, objects can reference these objects without problems.
- This makes the special logic to deserialize such `ddic` objects twice (during `ddic` and `late` steps) obsolete (1)
- Since `ddic` objects are processed only once, related warning messages will appear only once as well

Closes #5392 

(1) `late` is still required for `tabl` which reference `shlp` or have foreign keys


mbtools:
I clean-up and made some adjustments:
- renamed `pre_ddic` to `early - Pre-process Objects` (to counter `late` step)
- added `late` back for `tabl` since it's used for search helps etc
- removed obsolete methods

mbtools:
Tested 6 repos successfully (so far): 
https://github.com/abapGit-tests?q=ref

mbtools:
No issues found related to this PR when running full CI tests. Approved.

mbtools (approved):
👍

PR #5403, opened 2022-03-23
Remove popup for XML version mismatch

mbtools:
It's unlikely that a XML version mismatch error will happen. Therefore, the popup was replaced with an exception which makes the case testable.


larshp (approved):
yep

PR #5413, opened 2022-03-28
Folder logic "mixed"

### LINKED ISSUE #5410: Long SAP package names
If the package hierarchy contains longer name that do not fit to the custom namespace (does not begin with Y/Z/NAMESPACE), then neither folder logic `prefix` nor `full` will work.

This can happen if the folder structure is predefined by 3rd party (as for example exercism.org).

Example:

```
/concept
/concept/largest-series-product
/practice
/practice/kindergarten-garden
```

With very short package name `ZE` and `prefix` logic:

```
ZE
  ZE_CONCEPT
    ZE_CONCEPT_LARGEST-SERIES-PRODUCT " too long
  ZE_PRACTICE
    ZE_PRACTICE_KINDERGARTEN-GARDEN " too long
```

With `full` logic:

```
ZE
  CONCEPT " not valid
    LARGEST-SERIES-PRODUCT " not valid
  PRACTICE " not valid
    KINDERGARTEN-GARDEN " not valid
```

Proposal:

Introduce a new folder logic named `mixed` which will use the root package name as prefix but *not* add the folders recursively.

```
ZE
  ZE_CONCEPT 
    ZE_LARGEST-SERIES-PRODUCT 
  ZE_PRACTICE 
    ZE_KINDERGARTEN-GARDEN 
```

mbtools:
Introduces a new folder logic named `mixed` which combines `prefix` and `full`. The root package name will be used as prefix for all sub-packages but the package names are *not* concatenated recursively. This will allow for using significantly logic package names. See #5410 for an example.

Also includes several new unit tests especially for nested packages which had not been tested before.

Closes #5410

image

image

image

Todo: Update docs


flaiker:
Won't this common structure cause problems?

- zapp
  - mod1
    - core
    - test
  - mod2
    - core
    - test

zapp_mod1_core and zapp_mod2_core would both become zapp_core ?

mbtools:
`core` and `test` can't exists twice in ABAP, so you can't construct this in a system. 

If you tried to do it in git, then a import will raise an exception (like it does in similar case for `prefix`).

flaiker:
Ah I was thinking of the other direction. You maintain this package structure in the system.

- ZAPP
  - ZAPP_MOD1
    - ZAPP_MOD1_CORE
  - ZAPP_MOD2
    - ZAPP_MOD2_CORE

Then create a repo in abapGit using `mixed`. As I understand it ZAPP would be stripped as the prefix of the top level package. The prefixes of the leafs would be as well? Resulting in this folder structure?

- /src
  - /mod1
    - /core
  - /mod2
    - /core

But if you pull / clone the repo now into ZAPP this would result in this package structure?

- ZAPP
  - ZAPP_MOD1
    - ZAPP_CORE
  - ZAPP_MOD2
    - ZAPP_CORE

Where ZAPP_CORE is not unique. I probably misunderstood something. (Maybe the prefixes of the parent packages of the leafs are not removed, then it might work)

mbtools:
Your `/src` example is the existing `prefix`. 

With `mixed`, it does strip the prefix (root) as you wrote, but keeps the rest as is:

- /src
  - /mod1
    - /mod1_core
  - /mod2
    - /mod2_core

mbtools:
Any objections?

I would like to merge so this can be used for [Exercism repo](https://github.com/exercism/abap)

larshp (approved):
lgtm

PR #5417, opened 2022-03-30
Deserialize textpool for some objects

mbtools:
CI tests have shown that some text pool entries are missing i.e. there are diffs after pulling.

Fixes inserting the textpool for objects that are not handled by `RPY_PROGRAM_INSERT`. This applies for example to function groups and exit includes (`zx...`).


larshp (changes requested):
I'd guess `METHOD deserialize_textpool` is called from somewhere to handle the textpools?

mbtools:
You're correct. But the method wasn't working if there's only the report title maintained.

PR #5445, opened 2022-04-04
DOMA: Deletion of domains with append

mbtools:
On newer releases, suppressing the confirmation popup does not work since the parameter has changed.

Added a call using the new parameter.


larshp:
these two parts look the same?

image

mbtools:
at some point SAP changed the parameter name:
`n/a` 7.02
`no_ask_delete_append` 7.50
`aie_force_deletion` 7.54

larshp:
ahh 👍

PR #5488, opened 2022-04-26
Allow different user/password per repository in background mode

gepparta:
If we have different git projects on same server
and created BOT users + access tokens per project (for example in gitlab)
to be able to auto commit in abapGit background mode,
current implementation is using only first found user+token combo.

Further logins to repos running into errors: 
"Unauthorized access. Check your credentials"
image

one possible solution is mentioned here #5482

Clear auth buffer to allow different user/password per repository in background mode


mbtools (approved):
👍 thank you!

PR #5506, opened 2022-04-30
Improve error handling during activation phase

mbtools:
When the "Activation Popup" is turned off in "Personal Settings", any errors during activation will lead to "Activation cancelled. Check the inactive objects" which is not very helpful.

The change will add the errors to the abapGit log. 

It will also handle the special case when activation of includes leads to error "Include is used in multiple main programs" (EU 404). In such case, you can turn on the activation popup and repeat the abapGit process successfully:

image


larshp (approved):
lgtm 👍

PR #5514, opened 2022-05-01
Repo view: Show icon for different source system

mbtools:
In case an objects originates from a different system (`<> sy-sysid`), an icon will be shown after the object. On hover, you can see the id of the original system. 

Closes https://github.com/abapGit/abapGit/issues/3865

image


larshp:
getting few dumps when pulling

resulting in TOP of function group as inactive, activate manually

image

image

larshp (approved):

PR #5518, opened 2022-05-02
OTGR: Fix errors with parent/child

mbtools:
- Fix "error while instantiating CL_CLS_OBJECT_TYPE_GROUP"
- Remove children since they are created automatically


mbtools:
Successful tests:
- https://github.com/abapGit-tests/OTGR
- https://github.com/abapGit-tests/OTGR_with_parents

PR #5544, opened 2022-05-09
Improve error message for language mismatch

### LINKED ISSUE #5543: System Language Mismatch When Pulling Repo
H!

Just tried to download this repo https://github.com/SAP/code-pal-for-abap.git and when pulling the repo I get this error:
image

"Current login language 'DE' does not match main language 'EN'. Select 'Advanced' > 'Open in Main Language' "

When opening the menu, I do not see this mentioned entry:
image

So, basically I assume I just can switch the login language to proceed, nevertheless either the menu Advanced is missing an entry or the error message is wrong.

christianguenter2:
closes #5543
If no abapGit transaction code is available the error message now suggest to logon in main language instead of pointing to missing feature 'Advanced > Open in main language'.

### DIFF
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -196,11 +196,25 @@ CLASS ZCL_ABAPGIT_REPO IMPLEMENTATION.
     lv_main_language = get_dot_abapgit( )->get_main_language( ).
 
     IF lv_main_language <> sy-langu.
-      zcx_abapgit_exception=>raise( |Current login language |
-                                 && |'{ zcl_abapgit_convert=>conversion_exit_isola_output( sy-langu ) }'|
-                                 && | does not match main language |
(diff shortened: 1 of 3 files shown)


g-back:
Maybe add a text somewhere in the error message that a TCode is required for the functionality? Otherwise it may be a bit obscure for the user why the functionality is missing

mbtools:
an ideal case for a msg longtext... but that doesn't work in AG :-( [it does if you raise an SAP T100 msg but we can't ship our own]

flaiker:
> an ideal case for a msg longtext... but that doesn't work in AG :-( [it does if you raise an SAP T100 msg but we can't ship our own]

Could add a string / string_table attribute for the non-t100-longtext to `zcx_abapgit_exception`, add it as a parameter to constructor / raise-methods and return it in `if_message~get_longtext` ? Seems to work:

image

image

You only see longtexts on hover though with no indication of a longtext being available on hover, unless I missed it. Also the X moves when you try to click it. But those are different issues.

(If I recall correctly this would not work with the message statement but that's irrelevant here)

PR #5569, opened 2022-05-18
Refactor: Commit list popup

mbtools:
- Move popup for selecting a commit to `zcl_abapgit_popup`
- Allow to skip sorting of commits in `zcl_abapgit_git_commit` which often led to "too many commits" error (on GitHub you get the 300 most recent commits)

Popup will be used in `zcl_abapgit_gui_page_sett_remo` and `zcl_abapgit_gui_page_tags` (separate PR)

### DIFF
--- a/src/git/zcl_abapgit_git_commit.clas.abap
+++ b/src/git/zcl_abapgit_git_commit.clas.abap
@@ -14,6 +14,7 @@ CLASS zcl_abapgit_git_commit DEFINITION
         !iv_branch_name       TYPE string
         !iv_repo_url          TYPE zif_abapgit_persistence=>ty_repo-url
         !iv_deepen_level      TYPE i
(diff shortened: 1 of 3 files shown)


g-back:
In which method was the logic changed? It's a bit hard to review when the move to IF and logical changes are in the same commit 😅

mbtools:
I included the change to Repo > Settings > Remote. You can test the commit popup there, for example with the abapGit repo.

g-back (approved):
Thanks, I just couldn't see what logic you changed on first glance - but it seems you've only added the 'iv_sorted' flag.

Looks good to me, works as expected, commits are sorted descending by timestamp and you can actually check out commits from abapgit. No syntax errors either. Should improve performance as well as the unnecessary sorting is skipped.

PR #5591, opened 2022-05-25
Refactor: Create Tag

mbtools:
Replaces the old style page with a nice HTML form for creating a new tag.

Before:

image

image

After:

image

image


flaiker:
Looks good!

Minor issue: For annotated tags the field Comment appears to be mandatory in the UI but this isn't enforced when clicking create. You then get a tag without comment: https://github.com/flaiker/test-interface-method-deletion/releases/tag/tag3

Actually you can create "anonymous" tags by also clearing the mandatory fields tagger name and tagger email:
https://github.com/flaiker/test-interface-method-deletion/releases/tag/tag4

mbtools:
The update of `mo_form_util` was missing. It's not checking for required fields.

Should we allow annonymous tags via AG?

mbtools:
comment, tagger, tagger email can be empty now since git has no problems with it

PR #5592, opened 2022-05-25
Support direct longtexts in exceptions

### LINKED ISSUE #5548: Support longer texts for exceptions
Currently `zcx_abapgit_exception=>raise( )` does not allow to define a longer explanation of the error and possible solutions (long text). 

Suggestion: 
- Add optional longtext parameter (type string)
- Enhance `zcx_abapgit_exception` to return longtext
- Add indication that a longtext exists
- Nice-to-have: Support some HTML like ahrefs

See https://github.com/abapGit/abapGit/pull/5544#issuecomment-1122185236

image

flaiker:
Fixes #5548

This adds a longtext parameter to the exception class which can be used to provide an additional error message longtext without the need for a T100 message. The longtext can be supplied for in any of the 3 scenarios (error message, T100 message, message from exception). If a T100 longtext is also available the longtext parameter is preferred.

The feature is now used to explain the missing "Open in main language" button in the language mismatch error mismatch, as proposed in https://github.com/abapGit/abapGit/pull/5544#issuecomment-1122023003 .

The longtext may contain linebreaks which will be broken up into paragraphs when rendered (also for T100 longtexts, I think that must not have been supported before).

If there is a longtext available a comment in brackets after the (short) error message text is displayed indicating you can hover over the message to see it.

The "action buttons" in the message panel are now aligned to the bottom right. They do not move anymore when you hover over the X to close the panel if a longtext is available. (CSS implementation isn't that nice but works...)


larshp:
feel free to disable the failing unit tests by adding corresponding lines in https://github.com/abapGit/abapGit/blob/main/test/abap_transpile.json, assuming they work in a real system 😄 

I'll go through the disabled test sometime, and do the corresponding fixes in the transpiler

mbtools:
looks good. pulling with the dev version causes a dump because the public interface of `zcx_abapgit_exception` is changing. I'm not sure if this is caused by changing `ii_log` to `log` or if adding the optional `longtext` parameter is already triggering it. 

anyway, one just has to use the standalone version.

mbtools:
Tested and works.

image

image

`(longtext available on hover)` could be nicer esp. since it doesn't go away when hovering. Maybe just `More...` in bold. 

image

image

flaiker:
Yes that makes sense

mbtools (approved):
👍

Issue #5595, opened 2022-05-27
abapGit picks up "translations" to 늑 / 1Q language code for Translation Object Detective


gregorwolf:
Hello abapGit Team,

today I came around a quite interresting issue. @pksingh78 currently tries to fix the SCI issues in the abap2xlsx project (https://github.com/abap2xlsx/abap2xlsx/issues/1012 and https://github.com/abap2xlsx/abap2xlsx/issues/1013). As the ABAP Developer edition isn't available he does that in the customers sandbox system. Each time we continue working we see differences like this in abapGit:

![Screenshot 2022-05-27 at 08 27 02](https://user-images.githubusercontent.com/2639535/170643293-9bf2e245-6a0a-4b6f-9e7b-31dec26c006e.png)

To track down the issue I've looked at the issue with the translations using the language key 늑 I looked at the issues with object type MSAG:

![Screenshot 2022-05-27 at 08 28 16](https://user-images.githubusercontent.com/2639535/170643453-9e02c06b-fcff-417f-ba30-5aa2d27b0bfc.png)

The translations for this object are stored in table T100:

![Screenshot 2022-05-27 at 08 35 39](https://user-images.githubusercontent.com/2639535/170644479-5c0d07fb-99b2-4ff0-bdfd-e69261294c66.png)

So I found that 늑 has the two letter code 1Q and the description "Technical Code 1". A Google search revealed that in this system the:

[Translation Object Detective](https://help.sap.com/docs/ABAP_PLATFORM_NEW/ceb25152cb0d4adba664cebea2bf4670/88a3d3cbccf64601975acabaccdfde45.html?locale=en-US)

is active. By running the report RS_LXE_FILL_TECH_LANG the language 1Q is filled with a "translation" that contains the technical object name. So you can easily identify which SE63 object you need to translate when logging on with language 1Q.

But I think this translation should not be included in the abapGit language export. What do you think?

Best Regards
Gregor

gregorwolf:
In our concrete case for abap2xlsx the language serializers of the following object types would need to be adjusted:

- [ ] MSAG
- [ ] CLAS
- [ ] PROG
- [ ] DTEL
- [ ] DOMA
- [ ] TABL

gregorwolf:
I would suggest to define a constant c_translation_detective in class ZCL_ABAPGIT_OBJECTS_SUPER that is then added to the selection e.g. for the MSAG object type:

[...]

PR #5596, opened 2022-05-27
fix: exclude Translation Object Detective language for MSAG objects

gregorwolf:
Excludes Translation Object Detective language from serialization
Provides a solution for object type MSAG in issue #5595

### DIFF
--- a/src/objects/zcl_abapgit_object_msag.clas.abap
+++ b/src/objects/zcl_abapgit_object_msag.clas.abap
@@ -165,7 +165,7 @@ CLASS zcl_abapgit_object_msag IMPLEMENTATION.
                           WITH NON-UNIQUE DEFAULT KEY,
           lt_dokil             TYPE zif_abapgit_definitions=>ty_dokil_tt,
           ls_dokil             LIKE LINE OF lt_dokil,
-          language_filter      TYPE zif_abapgit_environment=>ty_system_language_filter.
+          lt_language_filter   TYPE zif_abapgit_environment=>ty_system_language_filter.
 
     FIELD-SYMBOLS: <ls_t100>  TYPE t100.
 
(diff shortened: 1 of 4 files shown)


flaiker:
This is quite an interesting feature. I scanned to documentation you linked and noticed in note 2087655 another technical language `2Q` is also mentioned and also excluded from logging. Should abapGit also exclude that language? Note 1943470 explains `2Q` as a technical language for "pseudo translations" to test internationalization features before translation occurs.

gregorwolf:
Hi @flaiker,

thank you for your feedback. I think this language should also be excluded. I've also made setting the language more robust.

flaiker:
Looking ahead a bit these languages would have to be excluded in other places as well, like mentioned here https://github.com/abapGit/abapGit/issues/5595#issuecomment-1139339620 . Some of these are not directly inside of the object handler but for example in `zcl_abapgit_longtexts` and `zcl_abapgit_lxe_texts` will likely want to exclude them as well.

Therefore I suggest moving the logic from the constructor of the object base class to `zcl_abapgit_environment`, maybe method `get_system_language_filter` which returns a range table of spras with `sign = 'E' option = 'EQ' low = '1Q' / '2Q'` depending on if they are installed using the conversion exit, possibly cached after first call. Then all the implementations could easily filter by

```abap
DATA(language_filter) = zcl_abapgit_factory=>get_environment( )->get_system_language_filter( ).
SELECT ...
  WHERE spras IN @language_filter
```

without requiring access to the object_super member variables or additional parameters.

gregorwolf:
I've followed your suggestions. Looking forward for your feedback.

flaiker:
Looks good to me. I checked if the range table should be cached because it is build on every call but since the underlying tables are buffered there's no noteworthy performance impact.

I don't have 1Q or 2Q installed either. So I cloned abapGit-tests/MSAG, translated it to German and excluded that for the test in get_system_language_filter. The diff then disappeared.

:+1:

flaiker (approved):

PR #5597, opened 2022-05-27
LXE: Downport

christianguenter2:
Prevents this dump in systems <= 750  when experimental LXE is active
grafik

grafik
grafik

ERR_MSG doesn't exist for fm LXE_OBJ_TEXT_PAIR_READ and LXE_OBJ_TEXT_PAIR_WRITE.

### DIFF
--- a/src/objects/texts/zcl_abapgit_lxe_texts.clas.abap
+++ b/src/objects/texts/zcl_abapgit_lxe_texts.clas.abap
@@ -63,6 +63,31 @@ CLASS zcl_abapgit_lxe_texts DEFINITION
           iv_object_name     TYPE sobj_name
         RETURNING
           VALUE(rt_obj_list) TYPE lxe_tt_colob .
+    METHODS
+      read_lxe_object_text_pair
+        IMPORTING
+          iv_s_lang                TYPE lxeisolang
+          iv_t_lang                TYPE lxeisolang
+          iv_custmnr               TYPE lxecustmnr
+          iv_objtype               TYPE trobjtype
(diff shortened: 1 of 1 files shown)


flaiker:
Maybe alternatively to exclusively raising the exception based on the release dependent err_msg raising might also be warranted if pstatus = cl_lxe_constants=>c_process_status_failure ('F') in _read and _write, assuming that exists on < 750?

christianguenter2:
Good idea. Tried with test repo https://github.com/abapGit-tests/FUGR_lxe.git but import fails with `lv_pstatus = 'F'` while `err_msg` being empty. Hm...

grafik

grafik

Status 'F' is set here:
grafik
grafik

Same for read:
grafik

So while not knowing much about LXE it seems better to rely on `err_msg` and fail silently for other cases.

flaiker:
> So while not knowing much about LXE it seems better to rely on err_msg and fail silently for other cases.

I agree for this PR. It should be analyzed separately though if it's even working correctly at all when the function modules are returning a failure status.

flaiker:
I also added the change to raise an exception on pstatus = F. https://github.com/abapGit-tests/PROG_lxe worked just fine. For https://github.com/abapGit-tests/FUGR_lxe I get the same error.

mbtools:
Our LXE approach is quite incomplete. Thanks for fixing the dump. 

For any further work on LXE, see also my comment in https://github.com/abapGit/abapGit/issues/4470#issuecomment-1027249544 (#2539 for the overall topic)

PR #5604, opened 2022-05-31
Show installation flavour in footer

flaiker:
This idea suddenly came to mind. You can also find this out less prominently under utilities -> debug info.

image

Might be nice to also display the merge timestamp (currently only a comment in lif_abapmerge_marker so not accessible) or the commit hash for the developer version.


mbtools (approved):
👍

PR #5627, opened 2022-06-09
Fix variant maintainance for abapGit report

### LINKED ISSUE #5626: Can't create a variant for background task
I'm encountering the same issue as [#1843](https://github.com/abapGit/abapGit/pull/1845) using the latest version(s) of abapGit. 
A blank screen is shown instead of the attributes tab. I've tried multiple workarounds but nothing seems to be doing the trick.

christianguenter2:
fixes #5626

### DIFF
--- a/src/zabapgit_forms.prog.abap
+++ b/src/zabapgit_forms.prog.abap
@@ -74,11 +74,16 @@ FORM output.
     TABLES
       p_exclude = lt_ucomm.
 
-  TRY.
-      zcl_abapgit_ui_factory=>get_gui( )->set_focus( ).
-    CATCH zcx_abapgit_exception INTO lx_error.
-      MESSAGE lx_error TYPE 'S' DISPLAY LIKE 'E'.
-  ENDTRY.
+  " For variant maintainance we have to omit this because
+  " it instantiates controls and hides mainainance screens.
+  " Variant mainainance uses SUBMIT which sets CALLD = 'X'.
+  IF sy-calld = abap_false.
+    TRY.
+        zcl_abapgit_ui_factory=>get_gui( )->set_focus( ).
+      CATCH zcx_abapgit_exception INTO lx_error.
+        MESSAGE lx_error TYPE 'S' DISPLAY LIKE 'E'.
+    ENDTRY.
+  ENDIF.
 
 ENDFORM.
 
--- a/src/zabapgit_forms.prog.abap
+++ b/src/zabapgit_forms.prog.abap
@@ -75,8 +75,8 @@ FORM output.
       p_exclude = lt_ucomm.
 
   " For variant maintainance we have to omit this because
-  " it instantiates controls and hides mainainance screens.
-  " Variant mainainance uses SUBMIT which sets CALLD = 'X'.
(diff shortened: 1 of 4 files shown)


src/zabapgit_forms.prog.abap
```diff
@@ -74,11 +74,16 @@ FORM output.
     TABLES
       p_exclude = lt_ucomm.
 
-  TRY.
-      zcl_abapgit_ui_factory=>get_gui( )->set_focus( ).
-    CATCH zcx_abapgit_exception INTO lx_error.
-      MESSAGE lx_error TYPE 'S' DISPLAY LIKE 'E'.
-  ENDTRY.
+  " For variant maintainance we have to omit this because
+  " it instantiates controls and hides mainainance screens.
```
larshp, line 78:
```suggestion
  " it instantiates controls and hides maintenance screens.
```

src/zabapgit_forms.prog.abap
```diff
@@ -74,11 +74,16 @@ FORM output.
...
+  " Variant mainainance uses SUBMIT which sets CALLD = 'X'.
```
larshp, line 79:
```suggestion
  " it instantiates controls and hides maintenance screens.
  " Variant maintenance uses SUBMIT which sets CALLD = 'X'.
```

flaiker:
Maybe check if memory id `%_SCRNR_%` is filled instead? (Include `LSVARF08` form `EXPORT_SCREEN_TABLES` sets it right before the submit)

christianguenter2:
@flaiker Thx. Switched logic to `%_SCRNR_%`. `VM$` is not reliable as it isn't cleared.

mbtools:
This works. However, it's quite confusing since there are not selection fields to fill. You have to press F3 to get the popup to save the variant. Definitely needs an update of docs with screen shots.

christianguenter2:
One of the reasons why it's confusing is that the toolbar is missing. I tried to reinsert toolbar for variant maintenance so that it's a bit more obvious what's going on. Still a bit ugly as the toolbar doesn't appear immediately when entering variant maintenance for the first time. Not sure how to fix that. 

After pressing `ENTER` once it looks like this:
grafik
And after switching screens:
grafik

At least at bit better than before. What do you think?

mbtools:
the toolbar magic, haha. yes, better. i don't think you can update it immediately (i think it's the same when you run zabapgit for the first time). 

A different solution could be to generate one such variant when you save the BG settings  in AG:
https://docs.abapgit.org/settings-background-mode.html

larshp:
but it does solve the problem at hand in https://github.com/abapGit/abapGit/issues/5626 ?

christianguenter2:
> but it does solve the problem at hand in #5626 ?
Yes it does.

larshp:
@mbtools can this PR be merged?

mbtools (approved):
Not great but works. Let's merge.

(If you stay in SE38 after maintaining a variant for program `ZABAPGIT`, and then execute the program, the toolbar remains on since memory id `%_SCRNR_%` is still filled. Once you use `ZABAPGIT` tcode, the toolbar is removed properly.)

PR #5635, opened 2022-06-21
Clear IWOM/IWMO object metadata

yellappam:
It clears metadata related to object types IWOM and IWMO. It uses the #5391 solution.


mbtools (approved):
Thank you 👍

PR #5637, opened 2022-06-21
Clear IWPR and IWSV metadata fields

yellappam:
It clears metadata related to object types IWPR and IWSV. It uses the #5391 solution.


mbtools (approved):
👍

PR #5643, opened 2022-06-22
Diff: Properly show diffs for whitespace

mbtools:
The SAP diff function ignores lines that contain only whitespace. Therefore such differences were not shown in abapGit. Some post-processing now identifies these diffs. It also shows a short message instead of an empty table in case there are no diffs.

Before:

saplogon_883

After:

Showing that on side actually contains spaces

image

If there's really no diff:

saplogon_884


larshp (changes requested):
add unit test?

larshp (approved):

PR #5656, opened 2022-06-27
LinkHints: enable radio button

christianguenter2:
This changes fixes the link hint feature for radios buttons like this.
grafik

Before that change submit wasn't triggered after selecting a radio button with a link hint.


mbtools (approved):
👍

PR #5662, opened 2022-06-27
Improve folder logic popup

christianguenter2:
Replaces this popup
grafik

with this
grafik

Makes it a bit less error prone.


mbtools:
there's also `MIXED` ;-)

larshp (dismissed):
see comment by @mbtools

christianguenter2:
Good catch. Never used that one. Fixed.
grafik

mbtools (approved):
👍

PR #5667, opened 2022-06-27
Rename repo interface method implementations

mbtools:
Renames methods after they were changed to an interface (no code changes). 

Interface methods are now properly named `<interface>~<method>` (instead of `<alias>`). 

This also fixes warnings and navigation issues in `SE24/SE80`.


christianguenter2 (approved):
👍

Issue #5681, opened 2022-07-11
WAPA: Serialization with a binary file


jrodriguez-rc:
There is a problem in the serialization of BSP application that has a binary file, in my case a FontAwesome otf file:
<img width="326" alt="image" src="https://user-images.githubusercontent.com/56556686/178230819-86a2ff4f-df80-4143-a3f4-8d05a9a5f526.png">

<img width="634" alt="image" src="https://user-images.githubusercontent.com/56556686/178231052-3635d24b-4039-43d3-907d-21b6380993b0.png">

The problem seems to be that the "page content" (otf file) is converted from string into xstring, but it is already a binary content
<img width="929" alt="image" src="https://user-images.githubusercontent.com/56556686/178231486-56fe4f3c-47fd-41dc-8305-52c7a032ed18.png">

If we could determine that the file is binary content the conversion to xstring will be skipped, just moving the content to the xstring without conversion

jrodriguez-rc:
The error was from the beginning, but it brings up from #5638

mbtools:
Use `zcl_abapgit_utils=>is_binary` to check

https://github.com/abapGit/abapGit/blob/main/src/utils/zcl_abapgit_utils.clas.abap#L52

mbtools:
A .js certainly should not be considered binary. Maybe check is xml_source is_binary before all these itab/string conversions.

jrodriguez-rc:
Hmm, it seems that P_XML_SOURCE is only returned when content is plain and when it is binary P_XML_SOURCE  is empty, returning only P_CONTENT

https://github.com/jrodriguez-rc/abapGit/blob/978c81f90af482c90f86d006ffc0df4264f6126b/src/objects/zcl_abapgit_object_wapa.clas.abap#L177-L190

mbtools:
Tried with https://github.com/abapGit-tests/WAPA_Fiori but that doesn't have any binary page content either. 🤷

jrodriguez-rc:
It was created with WebIDE some time ago.

I tried to create a new application with the same in WebIDE and BAS, but at the deployment, it was removed:
<img width="171" alt="image" src="https://user-images.githubusercontent.com/56556686/195582977-7bc57b08-3d7d-430a-a1ce-6be3f59eceaf.png">
<img width="291" alt="image" src="https://user-images.githubusercontent.com/56556686/195583013-baced322-d2ca-4389-b2d1-cf0be351269f.png">

mbtools:
[...]

PR #5687, opened 2022-07-15
Serialization binary files in WAPA object

jrodriguez-rc:
Fix for issue #5681


mbtools:
We replaced all of `cl_bcs_convert` before (https://github.com/abapGit/abapGit/pull/4605, https://github.com/abapGit/abapGit/pull/4693, This would bring it back...

Maybe add a `bintab_to_xstring` method to `zcl_abapgit_convert` without using bcs.

larshp (changes requested):
try removing use of `cl_bcs_convert`

jrodriguez-rc:
I'm still working on it, the bcs classes are replaced but the serialized file generated is not valid.
I can't open it downloading from GitHub and the size is 10x

jrodriguez-rc:
I avoid using BCS classes and SOLI[X] types.

The serialization seems to work, but I need help to test the deserialization, I cannot uninstall the application which I tested the serialization because it is under development by a colleage

mbtools:
This needs BAS to reproduce and test but my BAS isn't working 😦

PR #5727, opened 2022-08-22
Fix language conversion, now 7.02 compatible

albertmink:
solves #5726


AndreaBorgia-Abo (approved):
Patched source according to PR, program can be activated and successfully used to install abap2xlsx demos from ZIP as local package. Thanks for the speedy fix!

PR #5739, opened 2022-09-05
Allow use of changelog feature without APACK

fabianlupa:
... by introducing a version constant setting in .abapgit.xml.

Draft. Is this something we want? I think the changelog feature might be something we want to offer without use of APACK and without being restricted to abapGit itself.

This does not solve the issue of the old version value being present after refresh as the load will not be invalidated / refreshed without a restart of abapGit (would need something like #5285 / issue #5284). One could of course directly specify the version in XML but then it's not accessible in ABAP for other uses the repo might itself have.

image

CC @sandraros @AndreaBorgia-Abo @bigld


src/zcl_abapgit_version.clas.abap
```diff
@@ -242,4 +249,27 @@ CLASS zcl_abapgit_version IMPLEMENTATION.
...
 
   ENDMETHOD.
+
+  METHOD get_version_constant_value.
+    DATA: lv_version_class     TYPE string,
+          lv_version_component TYPE string.
+    FIELD-SYMBOLS: <lv_version> TYPE string.
+
+    IF iv_version_constant NP '*=>*'.
+      zcx_abapgit_exception=>raise( 'Version constant needs to use the format CLASS=>CONSTANT' ).
```
larshp, line 259: guess it will also work with interfaces?

larshp:
overall: nice, I like this solution

larshp:
ref https://github.com/abapGit/abapGit/issues/3152

src/zcl_abapgit_version.clas.abap
fabianlupa: Yes it does (see change in abapGit's own .abapGit.xml . I couldn't think of a good way to account for this in the message and also the UI hint. `OBJECT` to me rather refers to instances and `{ CLASS / INTERFACE }=>CONSTANT` might just confuse. Maybe `CLASS=>CONSTANT / INTERFACE=>CONSTANT` would be better.

fabianlupa: Well actually attributes will also work. So `COMPONENT` would be more correct than `CONSTANT`. Attributes will also call `class_constructor` though (do constants do that as well?) which means code execution of the (active) repo code :/.

AndreaBorgia-Abo:
ping me when you need testing / review :)

src/zcl_abapgit_version.clas.abap
mbtools: class constants do not trigger the class constructor. reading class attributes does (on 754)

mbtools:
Maybe add a read-only field below that shows the current value of the selected constant (assuming the code hasn't changed).

mbtools:
- Remove unit test that doesn't work with transpiler
- Added read-only fields for version value:
 
image

Let's get this merged 👍

mbtools (approved):
LGTM 👍

larshp (approved):
lgtm

PR #5741, opened 2022-09-06
DEVC: Fix dump if package has been delete

mbtools:
If a package is deleted in a different session while viewing a repository, refreshing the repo view will dump.

image


larshp (approved):
LGTM

PR #5749, opened 2022-09-07
WDYN: Fix dump if no components exist

mbtools:
If a Web Dynpro contains no components, all (!) longtexts in the system will be read (>650,000). No only is this a performance/memory issue but can also lead to the following dump (due to some inconsistent entries):

image

The solution is to avoid `select for all entries` with an empty table.


larshp (approved):
nice find

need https://rules.abaplint.org/unsecure_fae/ to work better...

PR #5768, opened 2022-09-20
CLAS: Fix pull for deleted interface methods

### LINKED ISSUE #5578: CLAS: Selective Pull for deleted interface methods doesn't work
I added a method to zif_abapgit_gui_event_handler and want to reset to the state of upstream/main. This doesn't work. While the method gets removed from the interface the method implementations are not removed from the implementing classes on selective pull.

I think this was discussed previously but I am not sure we have an open issue for this case (I couldn't find it).

Overview:
image

Diffs:
image

Selective Pull:
image

Note the interface itself already doesn't have the method in question anymore and has no diff.

In SE24 I get this warning:

image

The method in question isn't listed in the form based editor.

In the source code based editor it's there:

image

Regenerate sections didn't help.

Clean up method includes did help and showed this popup:

image

The buttons are in the menu in SE24:

image

mbtools:
In case an interface method was deleted, pulling a class that had implemented such method would not delete the implementation. This resulted in a diff and manual effort to remove the method (or repair the class).

The fix will compare the implemented methods and trigger the repair if necessary.

Closes #5578


src/objects/oo/zcl_abapgit_oo_class.clas.abap
```diff
@@ -195,6 +205,14 @@ CLASS zcl_abapgit_oo_class IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD get_method_includes.
+    " get method includes for implemented interfaces
+    " this will contain also leftover includes for deleted interface methods
+    rt_includes = cl_oo_classname_service=>get_all_method_includes( iv_classname ).
+    DELETE rt_includes WHERE cpdkey-cpdname NA '~'.
```
fabianlupa, line 212: I wonder if that still works if you use an alias in the method implementation (which you should not do but can if I recall correctly).

mbtools: i removed the delete since it should cover the alias case as well. worst case, it calls the repair unnecessarily.

fabianlupa (approved):
Tested with https://github.com/fabianlupa/test-interface-method-deletion, works well 👍

PR #5773, opened 2022-09-23
TABL: Fix deserialize if foreign key exists

### LINKED ISSUE #5694: TABL: Deleted foreign key XYZ is still used in maintenance view
In our simplified repository, we have a table and a maintenance view for it.

The table contains a foreign key relationship and the maintenance view uses that foreign key relationship to display additional fields
I set up a minimum size test repository here: [https://github.com/theconfigteam/tableViewForeignKeyTest.git](https://github.com/theconfigteam/tableViewForeignKeyTest.git)

The table with a foreign key to USR02
image

The maintenance view referencing the USR02 table
image
image

When we do a change to the table in the repository (e.g. checkout branch changedVersion) - the activation fails and leaves the table partially inactive. We have to reset it to the active state, then try changing manually. 

image

In our main repository we have 20 such tables with many foreign keys - pulling anything in that space leads to a very painful side effects because subsequent, other ABAP objects fail to activate.

This could be a side-effect of https://github.com/abapGit/abapGit/pull/3945

mbtools:
In case a foreign key definition already exists for a table and is referenced in another object - like a maintenance view - the activation would fail with the error "deleted foreign key is still used":

image

The reason was that the first deserialize phase activated the table without any foreign keys (to prevent dependency issues, in case the check table doesn't exist). Now only non-existing foreign key definitions will be removed in the first phase avoiding the error.

Closes #5694

### DIFF
--- a/src/objects/zcl_abapgit_object_tabl.clas.abap
+++ b/src/objects/zcl_abapgit_object_tabl.clas.abap
@@ -59,6 +59,8 @@ CLASS zcl_abapgit_object_tabl DEFINITION
     TYPES:
       ty_dd03p_tt TYPE STANDARD TABLE OF dd03p .
     TYPES:
+      ty_dd08v_tt TYPE STANDARD TABLE OF dd08v.
+    TYPES:
       BEGIN OF ty_dd02_text,
         ddlanguage TYPE dd02t-ddlanguage,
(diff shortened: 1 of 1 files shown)


larshp (approved):
LGTM

PR #5778, opened 2022-09-25
Force usage of local object handler classes

fabianlupa:
in standalone version in supported object type determination.

Without this if the installed developer version has an object handler and the standalone version doesn't it gets called as its type is not shadowed. If there's any syntax error in the developer version (like it crashing because the definitions interface is updated at runtime) then the standalone version doesn't work either.

This approach should work better than using RTTS and in my opinion any such dynamic `CREATE OBJECT` should include the absolute type name when running in standalone to prevent accidental cross usage.

CC @mbtools


mbtools:
i learned something. genius solution.

mbtools (approved):
👍

larshp (approved):

PR #5779, opened 2022-09-26
AFF with namespaces

albertmink:
Objects with namespaces were not correctly handled by ABAP file format (AFF) framework.

AFF specifies namespaces according to https://github.com/SAP/abap-file-formats/blob/main/docs/specification.md#file-names

* abapGit now considers files `(abcd).my_object.xxxx.json` as object `my_object` with namespace `abcd` and object type `xxxx`.
* example object in repository https://github.com/abapGit-tests/CHKC_namespace

### DIFF
--- a/src/objects/aff/zcl_abapgit_aff_registry.clas.abap
+++ b/src/objects/aff/zcl_abapgit_aff_registry.clas.abap
@@ -6,8 +6,7 @@ CLASS zcl_abapgit_aff_registry DEFINITION
   PUBLIC SECTION.
     INTERFACES:
       zif_abapgit_aff_registry.
-    CLASS-METHODS:
-      class_constructor.
+
     METHODS:
       constructor
         IMPORTING
@@ -37,16 +36,6 @@ ENDCLASS.
 
 CLASS zcl_abapgit_aff_registry IMPLEMENTATION.
 
-  METHOD class_constructor.
-    register( iv_obj_type = 'CHKC' ).
-    register( iv_obj_type = 'CHKO' ).
-    register( iv_obj_type = 'CHKV' ).
-    register( iv_obj_type = 'EVTB' ).
-    register( iv_obj_type = 'INTF'
-              iv_experimental = abap_true ).
-    register( iv_obj_type = 'SMBC' ).
-  ENDMETHOD.
-
   METHOD constructor.
     IF io_settings IS SUPPLIED.
       mo_settings = io_settings.
@@ -56,7 +45,18 @@ CLASS zcl_abapgit_aff_registry IMPLEMENTATION.
   ENDMETHOD.
 
   METHOD zif_abapgit_aff_registry~is_supported_object_type.
+
     DATA ls_registry_entry TYPE ty_registry_entry.
+
+    IF gt_registry IS INITIAL.
+      register( iv_obj_type = 'CHKC' ).
+      register( iv_obj_type = 'CHKO' ).
+      register( iv_obj_type = 'CHKV' ).
+      register( iv_obj_type = 'EVTB' ).
+      register( iv_obj_type = 'INTF'
+                iv_experimental = abap_true ).
+      register( iv_obj_type = 'SMBC' ).
+    ENDIF.
 
     READ TABLE gt_registry WITH TABLE KEY obj_type = iv_obj_type INTO ls_registry_entry.
     IF sy-subrc = 0 AND ls_registry_entry-experimental = abap_false.
--- a/src/objects/core/zcl_abapgit_filename_logic.clas.abap
+++ b/src/objects/core/zcl_abapgit_filename_logic.clas.abap
@@ -19,8 +19,7 @@ CLASS zcl_abapgit_filename_logic DEFINITION
         extension TYPE c LENGTH 4 VALUE 'json',
       END OF c_json_file.
 
-    CLASS-METHODS:
-      class_constructor.
+
     CLASS-METHODS file_to_object
       IMPORTING
         !iv_filename TYPE string
@@ -67,6 +66,7 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
     REPLACE ALL OCCURRENCES OF '#' IN lv_type WITH '/'.
     REPLACE ALL OCCURRENCES OF '#' IN lv_ext WITH '/'.
     " Assume AFF namespace convention
+    CREATE OBJECT go_aff_registry TYPE zcl_abapgit_aff_registry.
     IF go_aff_registry->is_supported_object_type( lv_type ) = abap_true.
       REPLACE ALL OCCURRENCES OF '(' IN lv_name WITH '/'.
       REPLACE ALL OCCURRENCES OF ')' IN lv_name WITH '/'.
@@ -133,6 +133,7 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
     ENDIF.
 
     " handle namespaces
+    CREATE OBJECT go_aff_registry TYPE zcl_abapgit_aff_registry.
     IF go_aff_registry->is_supported_object_type( is_item-obj_type ) = abap_true.
       FIND ALL OCCURRENCES OF `/` IN rv_filename MATCH COUNT lv_nb_of_slash.
       IF lv_nb_of_slash = 2.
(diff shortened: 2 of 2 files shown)


albertmink:
Manuell test activity, based on the standalone report.
I was able to pull object with namespaces (of type CHKC), make changes and commit/push them back to the git repository.
➡️ Works as expected.

albertmink:
I guess, we need to check first if the namespace approaches of abapGit and AFF algin.

mbtools:
Could you please add a namespaced object to https://github.com/abapGit-tests/CHKC_namespace?

albertmink:
Actually, it seams as if it fails to associate files with objects
image

mbtools (approved):
it would not work if the object name contains any `/` but that's probably not the case for any of the aff objects. 

lgtm

src/objects/core/zcl_abapgit_filename_logic.clas.abap
```diff
@@ -121,8 +127,15 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
     ENDIF.
 
     " handle namespaces
-    REPLACE ALL OCCURRENCES OF '/' IN rv_filename WITH '#'.
-    TRANSLATE rv_filename TO LOWER CASE.
-
+    IF is_item-obj_type = to_upper( c_json_file-extension ).
```
schneidermic0, line 130: For all JSON-based files like CHKO, CHKC, CHKV, EVTB this is fine.

How would we handle this (e.g., for INTF), if the file extension is `.abap`? In this case we would also need a different replacement for namespaces, don't we? Can we distinguish this in this method whether an object supports the ABAP file format?

schneidermic0 (commented):
I am not sure whether the change in `zcl_abap_filename_logic` is sufficient if it only checks for the file extension `.json`.

For the currently supported object types most probably, yes. For future object types with other file extension it seems not to be sufficient.

schneidermic0 (changes requested):
Since @mbtools has already approved. I put this to `request changes` until my question has been clarified.

src/objects/core/zcl_abapgit_filename_logic.clas.abap
mbtools: list of object types + check experimental flag

schneidermic0: Mmmh, I think for CHKV, CHKO, CHKC, EVTB (which only support AFF) we need to do it without experimental flag.
For INTF (and others which still support the old format) we should check also the experimental flag.

Would you suggest to implement it this way?

mbtools:
Needs to be lv_name and first occurrence only for the ( ) and replace

src/objects/aff/zcl_abapgit_aff_registry.clas.abap
```diff
@@ -0,0 +1,78 @@
...
+ENDCLASS.
+
+
+CLASS zcl_abapgit_aff_registry IMPLEMENTATION.
+
+  METHOD class_constructor.
+    register( iv_obj_type = 'CHKC' ).
+    register( iv_obj_type = 'CHKO' ).
+    register( iv_obj_type = 'CHKV' ).
+    register( iv_obj_type = 'EVTB' ).
```
mbtools, line 44: we just added 'SMBC'

mbtools:
there's an `input filter` for the objects included in the transpiler test run:

https://github.com/abapGit/abapGit/blob/main/test/abap_transpile.json

add `zcl_abapgit_aff_registry` to it

schneidermic0 (commented):
Looks good for me.

schneidermic0 (approved):

src/objects/core/zcl_abapgit_filename_logic.clas.abap
```diff
@@ -61,6 +61,9 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
     REPLACE ALL OCCURRENCES OF '#' IN lv_name WITH '/'.
     REPLACE ALL OCCURRENCES OF '#' IN lv_type WITH '/'.
     REPLACE ALL OCCURRENCES OF '#' IN lv_ext WITH '/'.
+    " Assume AFF namespace convention
+    REPLACE ALL OCCURRENCES OF '(' IN lv_name WITH '/'.
+    REPLACE ALL OCCURRENCES OF ')' IN lv_name WITH '/'.
```
mbtools, line 66: This won't work since ( and ) can be part of the object name. For example: `ztest(name).w3mi.data.json` (see https://github.com/abapGit-tests/W3MI_names).

best to also check the registry with `lv_type`.

src/objects/core/zcl_abapgit_filename_logic.clas.testclasses.abap
```diff
@@ -128,7 +128,7 @@ CLASS ltcl_run_checks IMPLEMENTATION.
       exp = 'ZMIME_<>_?'
       act = ls_item-obj_name ).
```
mbtools, line 130: add this test case:

```abap
    zcl_abapgit_filename_logic=>file_to_object(
     EXPORTING
       iv_filename = 'ztest(name).w3mi.data,json'
       iv_path     = '/src/'
       iv_devclass = '$PACK'
       io_dot      = mo_dot
     IMPORTING
       es_item     = ls_item
       ev_is_xml   = lv_is_xml ).

      exp = 'W3MI'
      act = ls_item-obj_type ).
      exp = 'ZTEST(NAME)'
      act = ls_item-obj_name ).
```

src/objects/core/zcl_abapgit_filename_logic.clas.abap
```diff
@@ -88,6 +91,8 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
   METHOD object_to_file.
 
     DATA lv_obj_name TYPE string.
+    DATA lv_nb_of_slash TYPE string.
+    DATA lo_aff_registry TYPE REF TO zif_abapgit_aff_registry.
```
mbtools, line 95: let's make this a class variable `go_aff_registry` and instantiate it in the class constructor.

mbtools (changes requested):
some small changes but works otherwise

src/objects/core/zcl_abapgit_filename_logic.clas.testclasses.abap
```diff
@@ -215,6 +237,18 @@ CLASS ltcl_run_checks IMPLEMENTATION.
       exp = 'zmime_%3c%3e_%3f.w3mi.jpg'
       act = lv_filename ).
```
mbtools, line 239: And this

```abap
    ls_item-obj_type = 'W3MI'.
    ls_item-obj_name = 'ZTEST(NAME)'.

    lv_filename = zcl_abapgit_filename_logic=>object_to_file(
      is_item  = ls_item
      iv_ext   = 'json' ).

      exp = 'ztest(name).w3mi.data.json'
      act = lv_filename ).
```

src/objects/core/zcl_abapgit_filename_logic.clas.abap
albertmink: @mbtools  Would you mind to double-check if UT assert is alright? With the current code I get, however this seam not related to changes in this pr.
image

mbtools: my fault. should be 

```abap
      exp = 'ztest(name).w3mi.json'
      act = lv_filename ).
```

src/objects/core/zcl_abapgit_filename_logic.clas.testclasses.abap
mbtools: see above

src/objects/core/zcl_abapgit_filename_logic.clas.abap
albertmink: thanks

mbtools (approved):
Not sure why the UT is failing in GH but this looks good to me now. 👍

larshp (changes requested):
remove the `class_constructor`s, they are the root of all evil

Looks like AFF registry should be a singleton, see https://docs.abapgit.org/development/tests.html

larshp (approved):

mbtools:
Great. Let's merge

PR #5788, opened 2022-10-02
Improvements in html engine

sbcgua:
to #5781

Just some by-the-way improvement. Refactoring the repo-over (next PR) I stumbled upon the `column` method - a wrapper for `td`. And I thought it was actually a good idea, and universal enough to be moved to the core engine, especially since there are a lot of tables in AG. So:

1. Added `td` and `th` methods as wrapper over text content or zif_html content (!). The latter allows skipping `render` in construction like `ri_html->add( zcl_chunks=>do_smth( )->render( ) )` - which is unreadable, sorry, and also produced unaligned htmls.

2022-10-01_19h42_45

So:

```abap
ri_html->td( '<span>Hello</span>' )
ri_html->td( ii_content = zcl_chunks=>do_smth( ) ) " No render() needed
ri_html->td( 
  iv_content = '<span>Hello</span>'
  iv_id    = 'id'
  iv_class = 'some_style' 
  iv_hint  = 'press this to ...' ) " For extra control

```

2. A universal `wrap` method (as the basis of td/th) which can do the same for any tag (just in case ... I'd not overuse it ....... maybe even make it private ........ because it's readability actually suck compared to dedicated `td` and `th`)
3. Chaining for `add`, and some others.
4. `new` - static constructor. I know it was there before and then removed for some reason. Please leave it, it is so convenient ! (having that there is no `new` operator)

Issue #5781, opened 2022-09-27
UI repo over view page: improvements


sbcgua:
Planning to do a couple of improvements during October:
- some refactoring to merge the related code in main and repo_over classes
- some visual finetunings in repo_over
- grouping tags (local ones) and filtering by tag

PR #3956, opened 2020-09-28
WIP theme refactoring

flaiker:
Draft pull request mainly to indicate work being done. Not ready for review. Feel free to change UI related stuff while this is open. Some ideas I am trying out:

- Make the default theme CSS not the base CSS but rather independent like the other themes
- Merge colors etc back into the common css file using the variables from the theme specific css files
- Add lots of more variable usages to get Quartz Dark to work without redefining lots of stuff like the current dark theme does
- Use color values by using a color picker on screenshots from SAP GUI instead of using the values from experience.sap.com as for some reason they are not exactly identical
- Clean up the CSS files in general (remove commented code and unused styles)

I added a theme test page to make things easier to see and debug (not finished).

image

Hopefully fixes #3530

PR #5790, opened 2022-10-02
UI repo over tweaks

sbcgua:
to #5781, after #5788, #5789

- fix bug accessing Remove command on toolbar
- fix bug accessing D (diff) hotkey
- improve table navigation hotkeys (added numpad up and down)
- improve table footer html structure
- proper (minimal) size for the go-to button
- rework go-to button a bit, a more visible icon ... but mainly larger click area around
- dedicated css styles (`repo-overview` class, before it was `db_tab` for some unknown reason)
- some, a bit arguable, cleanups in js/css interaction - js used `repo` class to identify table rows, the class didn't have any styles in fact, and was used for selection only. I changed it to `repo-overview tbody tr`. Which is more direct (though a bit less readable :( )

image

### DIFF
--- a/src/ui/zabapgit_css_common.w3mi.data.css
+++ b/src/ui/zabapgit_css_common.w3mi.data.css
@@ -626,26 +626,36 @@ table.db_tab pre {
   margin: 0px;
   width: 30em;
 }
-
 table.db_tab tr.firstrow td { padding-top: 0.5em; }
 table.db_tab th {
   text-align: left;
   padding: 0.5em;
+}
+table.db_tab thead tr {
   border-bottom: 1px solid;
+}
+table.db_tab tfoot tr {
+  border-top: 1px solid;
 }
 table.db_tab td {
   padding: 4px 8px;
   vertical-align: middle;
   word-break: break-all;
 }
-
 table.db_tab th.package {
     width: 45ch;
 }
-
 table.db_tab td.data {
   font-style: italic;
 }
+table.db_tab tbody tr {
+  /*border is there but transparent, needed for proper alignment*/
+  border-color: transparent !important;
+  border: 1px solid;
+  border-left-width: 3px;
+}
+
+
 
 li.action_link.enabled{
   visibility: visible;
@@ -1022,8 +1032,16 @@ table.commit tr .title {
 /* Repo overview */
 .repo-overview { font-size: smaller; }
 .repo-overview tbody td { height: 2em; }
-.ro-detail { display: none; }
-.ro-go { font-size: 150%; }
+.repo-overview .ro-detail { display: none; }
+.repo-overview .ro-go a {
+  font-size: 150%;
+  padding: 0px 0.15em;
+  border-radius: 0px;
+}
+.repo-overview .ro-go a:hover { 
+  background-color: white;
+  text-decoration: none;
+}
 
 /* Branch Overview Page */
 .gitGraph-scrollWrapper, .gitGraph-Wrapper{
--- a/src/ui/zabapgit_css_common.w3mi.data.css
+++ b/src/ui/zabapgit_css_common.w3mi.data.css
@@ -1032,24 +1032,14 @@ table.commit tr .title {
   text-align: left;
 }
 .repo-overview thead tr {
-  border-bottom: #ddd 1px solid;
+  border-bottom: 1px solid;
 }
 .repo-overview tfoot tr {
-  border-top: #ddd 1px solid;
+  border-top: 1px solid;
 }
 .repo-overview tr.favorite .icon-star {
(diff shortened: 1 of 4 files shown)


sbcgua:
OK, this one is ready for review and comments.

abaplint - seems to be a false positive ? It is kind of one line ... just the string segmentation ... otherwise it will be longer than 120 which will trigger another rule ?
image

larshp:
it should take that into account, https://github.com/abaplint/abaplint/blob/main/packages/core/src/rules/keep_single_parameter_on_one_line.ts#L53

mbtools:
these oddly wrapped string templates should be avoided.

Use: 

'''abap
s = |...| 
  && |...|.
'''

sbcgua:
fixed

mbtools:
Functionality is all good and we could merge it.

My concern is the consistency of the UI. It wasn't great before, but it still could be better. The look and feel of the HTML tables on "Repo Overview" and "Repo View" are quite different:

image

image

Diffs:
- Background behind the nav bar
- Background of the table
- Table border 
- Table header font and colors
- Hover color

I think the "Repo View" design is more aligned with the rest. The blue/light blue hover is better in "Repo Overview" but would not work with a blue table header.

We standardized all other controls, can we work towards some standard for HTML tables?

sbcgua:
Hmmm. These are all good points. Yet I didn't target to align the design in all pages honestly. To the better or worse. In general, fixing CSS for this PR, and also not for the first time, I though that current CSS structure leave a lot of room for unification and simplification. Many sections are just duplicated. 

However, I don't think we should focus unification (too much) in this PR ... it will just take forever ... I'd suggest fixing easy things and go on. And separate the remaining comment into an issue.

Now quickly to what can be fixed here.
1) nav bar bg - I'll check, but I'm almost sure it is a visual illusion
2) table bg - yeah, it was white. I changed it to `--theme-container-background-color` (which is f2f2f2) to align with dark theme at the very end. But. Side by side I like repo view (white) better. So I'll change it back. Somehow ...
3) table border - you mean that overview does not have it ?
4) tab header color is probably the same ... and style - do you prefer bold or normal like in repo view ? (I'm tentative)
5) Hover color vs COLORED tab header. Hmm ... yes, they will conflict. But that may mean that tab header should not be colored. Because we have hover and then selected row and then tab header. So using the same tint will not work anyway. Will do some small experiments. And also open for constructive ideas. But it is not a simple issue I'm afraid.

Anyway. Let's fix some small thing. But I'd not dive into "conceptual redesign" - it may be a separate and big subject.

Yet thank for good points :+1:

mbtools:
css refactor is a separate issue, indeed. we have https://github.com/abapGit/abapGit/issues/3530. Fabian had started on it in https://github.com/abapGit/abapGit/pull/3956. 

doing small steps is fine 👍 

1. +2. background colors +

  - body: #E8E8E8
  - main area: #F2F2F2 (between the double horizontal lines)
  - table: #FFFFFF

    for the overview, the main area behind the nav.bar is #E8E8E8 but should be #F2F2F2 (and table should be white as you said)

3. yes, add a light border (if the contrast between backgrounds is high enough then I'm ok without borders but we have just light shades of gray so a little border is helpful)

4. +5. i prefer regular font, darker font and background colors compared to the table body (i like to use a gray tone since i associate things colored with interactive elements. so actually, the blue header on the repo view should be gray).

sbcgua:
image

- `--theme-table-border-color`
- `--theme-table-cell-border-color`
- generally aligned with repo table style (+-)
- also added header shade. Better than I thought ... yet ... well ... will do. Gray tone - not sure, there are too many light gray shades on the page already. If choose gray, then maybe darker gray with light text color. But then repo styles will need to be changed as well ... I'd suggest doing it separately, maybe. 

Imho not bad. Thanks for the comments. Pretends to be final for my taste. :) 

BTW I removed `font-size smaller` ... although it made sense because of lots of data on the page. So condensed font made sense. On the other hand it was not in-line with all other pages. So I tentatively removed it.

Also I'd reconsider overflows (together with pkg/branch/user icons in cells). But honestly a bit lazy to do it now. 
image

Maybe if no major objections, let's merge soon. I'd like to finish the label feature which was the main goal of this all ;).

mbtools:
this looks awesome. i see you used more CSS variables. nice. 

we'll tackle the view with details separately. maybe we go for a two-line approach (like "object/files" in repo view). 

will give it a try tomorrow

mbtools:
`font-size:12pt` (16px) feels quite big and causes wrapping (depends on content, of course):

image

`font-size:smaller` is 13.33px, i tried `font-size:14px` which is a bit closer to the rest and works well:

image

mbtools (approved):
LGTM 👍

PR #5793, opened 2022-10-02
Push tags to builds

sbcgua:
to #5784

This should work in theory ... though potentially may require a fix or two ...


sbcgua:
@larshp would you mind if I merge it ? I'll be tracking the job logs for a while (especially after the next version) and will fix if it fails (it is difficult to recreate the test environment for this, in my own repo I just push the fixes until it starts working ...)

larshp (approved):
feel free to merge

sbcgua:
OK, merging then.

sbcgua:
didn't crash so far ... already good :)

PR #5794, opened 2022-10-03
Sort test case exclusions

mbtools:
Same content just sorted by object and test class for easier maintenance


mbtools:
conflict with #5791. will do this later

Issue #5798, opened 2022-10-04
Failing unit tests


christianguenter2:
I get these failing unit tests on 756 SP00 whereas they succeed on NW 752 SP04 dev edition.
![grafik](https://user-images.githubusercontent.com/17437789/193755714-c414ec18-1d9a-4bff-9402-d53b90102bbb.png)

christianguenter2:
Has something todo with [plugins](https://github.com/abapGit/abapGit-Plugins). Without them installed even more tests fail.
![grafik](https://user-images.githubusercontent.com/17437789/193759329-c9827c9a-4796-4c5c-b510-938d7b7937ce.png)

christianguenter2:
These 3 only occur when logged on in D. 
![grafik](https://user-images.githubusercontent.com/17437789/193771303-e9322339-3bf2-4a3d-9443-94ac3174ae7a.png)
It seems there's a new depedency to logon language/object language introduced somewhere NW > 752. Unit tests are exceuted with new report RS_AUNIT_SBOX_CLASS_TEST_INT, which does some language switching causing the issue.
![grafik](https://user-images.githubusercontent.com/17437789/193771751-e59a2362-09de-4df9-83b3-79a54e91c6ce.png)

PR #5801, opened 2022-10-04
Fix failing unit test for SAP_BASIS SP00

christianguenter2:
before this fix ls_requirement-min_patch gets negative for SP00 and therefore failing
grafik
#5798


mbtools (approved):
👍

PR #5803, opened 2022-10-04
AFF: Fix existence check

christianguenter2:
Fixes existence check for AFF objects causing failing unit tests on NW >= 756 SP00
[5798](https://github.com/abapGit/abapGit/issues/5798)
grafik

As abapGit assumes raised exceptions during existence check as existing objects
https://github.com/abapGit/abapGit/blob/1be856d23a571adcf0b6bb4166b998c170fe2fb7/src/objects/zcl_abapgit_objects.clas.abap#L862-L865


mbtools (approved):
👍

PR #5811, opened 2022-10-06
ENHO: Unlock objects in cases of errors

mbtools:
If an error occurs during deserializing enhancements, the enhancement objects remain locked:

image

image

The fix unlocks the objects in such cases.


larshp (approved):
Looks good to me 👍

PR #5814, opened 2022-10-08
Local organization/grouping/labeling of repos

### LINKED ISSUE #5781: UI repo over view page: improvements

sbcgua:
Closes #5781. The final PR for repo overview improvements. (to be merged after #5790)

A quick video overview :)

https://user-images.githubusercontent.com/15635498/198895429-6c6ec741-c7ce-4af1-a412-34fbc0e6ee2e.mp4

2022-10-30_19h17_38
2022-10-30_19h17_51
2022-10-30_18h59_24

Main feature:
- possibility to add labels to repos locally, they will display in repo overview
- filtering by label is possible and easily accessible from the top tool bar. filtering is technically integrated in the common filtering logic
- label list respects favorites show/hide mode; also it respects "no labels" mode and hides the column respectively
- labels can be colored. This is configured in AG settings per user (hmmm ... maybe it should be global?)
  - colors can be chosen from predefined (~20) styles
  - or custom `#fg_rgb/bg_rgb` format
- docs are updated as well - https://github.com/abapGit/docs.abapgit.org/pull/53

Technically:
- `labels` attr in `local_settings` structure
- parsing/validation utils for labels and colors are in `zcl_abapgit_repo_labels` (UT covered)
- labels are rendered by chunk lib `render_label_list` - rather universably and can be re-used. Slight visualization difference if repo list is rendered within a table
- in CSS:
  - color classes start from `rl-`
  - repo label list classes: `repo-labels` and `repo-label-catalog`
  - help hint (below) - `form-field-help-tooltip`
- side features:
  - `html->a/add_a` accepts `iv_query` - to specify `?query` in actions
  - a new feature `render_help_hint` in chunk lib to render hoverable help hint (supposedly to be used in forms for more detailed help - see example in Personal settings for label colors in the screen above)
  - `lcl_table_scheme` helper in repo overview - a candidate for a separate class, potentially a part of table rendering component


src/persist/zcl_abapgit_persist_tag_utils.clas.abap
```diff
@@ -0,0 +1,94 @@
+CLASS zcl_abapgit_persist_tag_utils DEFINITION
+  PUBLIC
+  FINAL
+  CREATE PUBLIC .
+
+  PUBLIC SECTION.
+
+    CONSTANTS c_allowed_chars TYPE string VALUE `-_.a-zA-Z0-9` ##NO_TEXT.
+    " it is easier to whilelist chars, though potentially other chars can be added later if needed
```
fabianlupa, line 9: `whilelist` -> `whitelist` -> `allow list` -> `... allow chars, ...` ?

sbcgua: yeah, i'll fix the typos ... thanks

mbtools:
Avoid "Tag", indeed. How about "Category"? `zcl_abapgit_repo_categories`, package `$abapgit_repo` (or "Group")

fabianlupa:
Or maybe "label".

sbcgua:
Thanks guys ! I was thinking about "group" as well, but label looks like the match ! I'll rename it to label. Maybe move the util class to utils package too.

sbcgua:
No styling yet, but already fun to play with :)

image

sbcgua:
Hmmm :)
image

mbtools:
just thinking about the table layout. maybe "Labels" becomes a checkmark like "Details". Also if selecting "Labels" or "Details" could hide the URL and maybe branch as well to make room. The labels could then be put next to each other - comma separated.

sbcgua:
They will, I didn't do any styling yet as I mentioned. It will be a typical bordered text items. Probably. 

Hiding other fields - probably not. All the current columns are important. But the labels (the column) are hidden if there are no of them.

src/utils/zcl_abapgit_repo_labels.clas.abap
```diff
@@ -0,0 +1,245 @@
...
+  PUBLIC SECTION.
+
+    TYPES:
+      BEGIN OF ty_label_color,
+        label TYPE string,
+        color TYPE string,
+      END OF ty_label_color,
+      ty_label_colors TYPE STANDARD TABLE OF ty_label_color WITH KEY label.
+
+    CONSTANTS c_allowed_chars TYPE string VALUE `-_.a-zA-Z0-9` ##NO_TEXT.
```
mbtools, line 15: This won't work well for many languages. Any reason why there are not more chars allowed? Why not everything except `,` (and no control chars, of course)?

mbtools (commented):
Just thinking about i18n

src/utils/zcl_abapgit_repo_labels.clas.abap
sbcgua: I prefer to start with limited but surely working stuff. Disallowed set can be not that small. controls chars, html escaped chars, comma, semicolon ... there might be some other limitations. Also I haven't seen a lot of localized labels, I mean in other tools. 
Anyway I don't mind, it's just let's do it after the feature is stable and real feedbacks received.

sbcgua:
image

image

image

sbcgua:
OK, I guess it's ready for review. The first post also updated with screenshots, user and tech details and a video :)

larshp (approved):
neat

wish we had some UI testing, but its a much different and large topic

gregorwolf:
@mbtools can you provide details why to avoid "tag"? The SAP Community uses [Tags](https://answers.sap.com/tags.html). In SAP BTP Labels are used as a key / value combination:

Screenshot 2022-10-31 at 08 11 25

mbtools:
@gregorwolf `tag` makes sense in general but in git clients `tags` have a different meaning. So sticking to `labels` for categorizing things like in GitHub or Gitlab looks good to me.

mbtools:
Overall, this is an awesome enhancement. Thank you!

1) Overview

Can we move the label column to the right? Feels much better there and less awkward if they appear after adding them. 

image

The screen is not too crowded but I can image a "Labels on/off" switch like for "Details" could make sense esp. if you use many or longer labels.

2) Colors

Initially, all labels are white. Some default colors would be nice to have.

image

The predefined colors are not that intuitive and don't match the predefined CSS names.

image

Do we really need labels that match the fg or bg colors? Why not use the standard CSS colors https://www.w3schools.com/cssref/css_colors.php? Then one color name would be sufficient instead of fg/bg (based on the colors of that particular page). 

3) Settings

With just a few labels, the input box becomes too small. Let's try textarea and just call it "Label Colors".

image

image


Very cool!

sbcgua:
Thanks for the detailed comments.

> "Can we move the label column to the right" and "Labels on/off"

I think it is a matter of preference. E.g for me the most important columns are: on/offline, repo name, branch and now labels. They should be on the left, others can be hidden.
As for labels - they are a grouping factor, thus imho they should be close to the name.
I don't think moving column to the right and hiding option would be "good for all" solution. "Good for all" solution is a mechanism to completely customize order and visible fields. This would solve all of that with one shot. But not in this PR.

> "Initially, all labels are white."

This is intentional. I wanted to keep them neutral and non-distractive until the user intentionally want to change that.

> Color names.

Not sure I understood the idea 100%. To admit, I also feel that there is some room for improvement in the color naming. But ... how to express it properly ... I spent quite some iterations on coloring to get to the current system so if it has to be changed it should be more systematic and more well-thought than now and has to take all previous "rakes" into account. Namely:

- initially I though to implement just FG rgb color.
- but then I though also to include html colors (the ones you mentioned)
- but then it appeared ugly, because it changes just fg and labels were not that distinguishable
- and I came to fg/bg pairs
- and so I got rid of html color because: 
  - A) they were not designed to match together (fg/bg)
  - B) they don't help unless you're a CSS expert remembering them
  - C) coloring, especially fg-bg relation goes well with HSL color concept, which has nothing to do with predefined html colors
- also I got to the understanding that assigning color pairs, especially in abap, is messy, has poor readability and usability
- in fact, now I'm inclined to remove the fg/bg rgb pairs at all ... they are so messy ... so after a couple of iterations and practical experiment I got to understanding that my initial idea was bad. (I left it to have this possibility but I personally will use styles below)
- so, instead fg/bg pairs, I decided to give user a small, but easy and pretty predefined choice with CSS styles
- ... which also have border style, which looks nicer (so we have already 3 color components)
- ... which is rather easily extendable by later PRs (`rl-` css styles)
- to compare, most of services give a predefined, rather small, yet well balanced choice (github, gmail ...)
- designing the style set (partially inspired by existing color set in gmail theme) I got to the conclusion that white and dark backgrounds naturally goes well with several colors. Thus `dark-w/y/r/b` were added and a couple of other tints.
- so the naming does not represent a single color ...
- finally, and importantly. Does anyone really cares about the color names ? The user just wants to choose a color. The fact is that current approach (with entering label-color pairs in text) is **not** too convenient and dictated more by technical constrains. The best solution would be a palette picker, with **no names**, but just a color. This is possible, but needs a lot of JS or multistep, potentially inconvenient, step-by-step abap procedure. Which is ... not in this PR.

Sorry for long text. Just wanted to explain that current approach is a product of rather long way.

3) Textarea in setting

Yes, good idea. Done :)

2022-11-02_08h05_58
(hmm, but now it has ugly overflow ... maybe I should render comma separator with a space after it ... but next PR)

mbtools:
It's possible that we are overthinking the colors a bit 😉 

GH labels are defined by one, randomized #rrggbb. Simple, works perfectly well.

sbcgua:
Could be, could be ... though I wouldn't harry with "simple". All colors are of limited saturation range, auto bright/dark foreground, text readability seems to be also considered ... I wonder how much engineering and attempts were behind ...
image

Also, there is no limit for perfection and further improvements ... ;P

mbtools (approved):
Let's do it 👍

larshp:
🥳

PR #5818, opened 2022-10-12
DRUL: Use CL_BLUE_SOURCE_OBJECT_DATA instead of CL_DRUL_WB_OBJECT_DATA

uenal-akkaya:
Use in class ZCL_ABAPGIT_OBJECT_DRUL the more generic type CL_BLUE_SOURCE_OBJECT_DATA instead of CL_DRUL_WB_OBJECT_DATA


mbtools:
`CL_BLUE_SOURCE_OBJECT_DATA` is already used in `BDEF` and `DTDC`. So this makes sense

Issue #5836, opened 2022-10-22
CLIF: Crash with CX_OO_INVALID_SUBCOMPO_KIND in UPDATE_META (ZCL_ABAPGIT_OO_INTERFACE)


kjetil-kilhavn:
I have installed the latest stand-alone version in several systems where I work recently. Using the new stand-alone version I pulled the updates of the development version into a system which had a really old version of that (1.89.0).
During deserialization the report crashed due to uncaught exception of type CX_OO_INVALID_SUBCOMPO_KIND. The exception is raised in CL_OO_SECTION_SOURCE method REVERT_PARAMETERS (line 70) - but the last abapGit method in the call stack is UPDATE_META of ZCL_ABAPGIT_OO_INTERFACE and the crash occurs in the last statement; lo_update->revert_scan_result( ).

I've looked at it for some time now without figuring out why it crashes, and when I tried again the deserialization completed. So possibly reproducible, but probably not easy to reproduce.

The object being deserialized when the crash occurs is ZIF_ABAPGIT_COMPARATOR.
When the exception is raised the PARAMETER structure has the following values:

name | value
----|----
CLSNAME | ZIF_ABAPGIT_COMPARATOR
CMPNAME | COMPARE
SCONAME | II_REMOTE
VERSION | 1
LANGU | E
DESCRIPT | 
CMPTYPE | 0
MTDTYPE | 	0
EDITORDER | 1
DISPID | 0
AUTHOR | 
CREATEDON | 00000000
CHANGEDBY | 
CHANGEDON | 00000000
PARDECLTYP | 0
PARPASSTYP | 1
TYPTYPE | 3
TYPE | ZIF_ABAPGIT_XML_INPUT
TABLEOF | 
PARVALUE | 
PAROPTIONL | 
PARPREFERD | 
LOCKED |

mbtools:
What's the SAP Basis release?

Ref:

https://github.com/abapGit/abapGit/blob/e4aeff4f048c4406076a8fb8011c58aec8d7a8d5/src/objects/oo/zcl_abapgit_oo_interface.clas.abap#L134

kjetil-kilhavn:
Ah, sorry about that. This system is on 7.50 (and SP level 24).

kjetil-kilhavn:
The error could possibly occur if ZIF_ABAPGIT_XML_INPUT had not been (fully) activated yet, but unfortunately I didn't dive deep down into the details before trying again.

mbtools:
[...]

PR #5841, opened 2022-10-26
CLAS,INTF: Catch runtime error during update

mbtools:
Prevent dump when deserializing classes and interfaces in case a sub-component was changed in a way that standard SAP code does not handle. The error message is now passed on to the user.

References: 
- https://github.com/abapGit/abapGit/issues/5836
- message `OO_SOURCE_BASED 019`
- SAP Note [2166231](https://launchpad.support.sap.com/#/notes/2166231)


larshp (approved):
LGTM

PR #5855, opened 2022-11-04
DOCV: Fix dump for namespaced documents

### LINKED ISSUE #5854: DOCV: Error "Separator '/' for namespace not paired or in wrong position"
We have just updated from abapgit 1.116 to latest 1.118 and now have an error in ZIP export.

There is an uncatched exception DELIMITER_ERROR of function RS_NAME_SPLIT_NAMESPACE in ZCL_ABAPGIT_OBJECT_DOCV that is being raised:
https://github.com/abapGit/abapGit/blob/93c8fcfb4fc514c80dc8728a5304c64a6adc55cd/src/objects/zcl_abapgit_object_docv.clas.abap#L50-L55

Error Message: e233(EU) "Separator '/' for namespace not paired or in wrong position"
While other calls of that function in  abapGit handle that exception, just doing that will probably not fix the issue for us.

The object are in our namespace and RS_NAME_SPLIT_NAMESPACE is called with name_with_namespace = "UO/NAMESPACE/NAME".  Probably that function needs to be changed as it cannot handle names of DOCV in a namespace with additional characters before /namespace/ ?

mbtools:
Fix handing of documents in classes `UO` and `UP` which use a different namespace logic than `DT`, `IN`, `MO`. 

Regression #5756

Closes #5854

### DIFF
--- a/src/objects/zcl_abapgit_object_docv.clas.abap
+++ b/src/objects/zcl_abapgit_object_docv.clas.abap
@@ -49,10 +49,10 @@ CLASS zcl_abapgit_object_docv IMPLEMENTATION.
     super->constructor( is_item     = is_item
                         iv_language = iv_language ).
 
-    IF ms_item-obj_name(2) = 'UO' OR ms_item-obj_name(2) = 'UP'.
+    IF ms_item-obj_name(2) <> 'DT'. " IN, MO, UO, UP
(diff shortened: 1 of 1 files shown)


abap-weasel:
issue #5854 resolved with that change - thanks! 👍

PR #5857, opened 2022-11-04
Improve repo overview when showing details

mbtools:
The screen got even more crowded with labels. Now showing details will hide the URL and branch columns. 

Before:

image

After:

image


fabianlupa:
The choice of columns for details off will always be very subjective. Personally I never need the repo key unless I am in the db utility or use the api? And all the timestamps and admin data is less important to me than to see if I still have a feature branch / PR checked out. I don't mind, just find the different choices worth pointing out.

fabianlupa:
Or wait: these columns show up in the normal non detailed Layout but are hidden when the detail layout is active? I intuitively thought clicking details shows all columns.

sbcgua:
I'd say it's better to invest in fully customizable table layout. To hide/show fields one needs, maybe change order. Everything else would be a compromise. I'm not against, just that maybe there is a better way

mbtools:
To clarify, the non-detail view remains the same. The detailed view shows the additional fields as usual but hides url and branch. It's a quick fix to reduce wrapped content or horizontal scrolling. A solution for user-defined table views is the ultimate solution (aka ALV). 

To be honest, we are getting closer to coding a full UI framework. I'm not a fan of such activity. There are plenty of those around. For me it's part of a discussion about long term goals and architecture (web version? external api? etc).

mbtools:
Closing in favor of some future redesign

PR #5840, opened 2022-10-25
Reset sort order to ascending in repo overview

kjetil-kilhavn:
Reset mv_order_by_descending to abap_false for action change-order-by in event handler
Solves issue #5839

Issue #5839, opened 2022-10-25
Repository overview: Not possible to sort ascending again after sorting descending


kjetil-kilhavn:
If you sort descending on a column it is not possible to sort descending again on any column.
The reason is that the descending attribute sticks once it has been set by the "direction" action. 

There are several possible solutions, depending on how the behaviour should be. The default is to sort by Name. It is possible to click the column header twice to get the repositories in database order - but is that a feature or a bug? 

If the list should always be sorted, the fix could be as simple as changing one line in zcl_abapgit_gui_chunk_lib->render_order_by_header_cells( ):
iv_act   = |{ zif_abapgit_definitions=>c_action-change_order_by }|
changes to 
iv_act   = |{ zif_abapgit_definitions=>c_action-direction }?direction=ASCENDING|

Personally I think this would be the best solution.
However, that will have a side effect. The repository view page will then also always have to be sorted, and this would perhaps make sense if we could specify sorting by multiple columns - e.g. object type and then object name - but that is a bigger change. 

So, I suppose the best quick fix is to just reset the direction attribute when the sort order is changed also in the repository overview page.

PR #5859, opened 2022-11-06
Repo overview page, Polishing

### LINKED ISSUE #5846: Remove filter autofocus at repo overview
Having that repo overview is quite an interactive page with multiple keyboard shortcuts, table navigation and etc., forced filter field autofocus on first display is very inconvenient. Filtering is by far not the most probable action after AG start.
If no-one objects I'll remove the autofocus and add a key-combination to focus it intentionally.
As a final touch to #5781

### LINKED ISSUE #5848: UI - Repository List - Dark Theme broken in Version 1.118.0
The dark theme seems to be broken in Version 1.118.0.
The background of the table used to be grey but is now white.
This makes it rather hard to read the text.

OLD:
OLD

NEW:
NEW

To reproduce the issue, the dark mode must be activated under "Settings" -> "Personal" -> "User  Interface" -> "Theme".
Settings

sbcgua:
final tuning for #5781

1) Rework reset sorting order (follows #5840, #5839)
2) Better overflow in settings

normalized label separator is `comma+space`. Looks nicer + fixes overflow

before
2022-11-06_16h01_32

after
2022-11-06_17h01_08

3) Remove filter autofocus (fixes #5846)
4) And, just by the way, same for Stage page, couldn't pass by, it was also annoying
5) Fix dark there for overview (fixes #5848, @NeumannJoerg please check)

All fixes commit-by-commit for easier review

### DIFF
--- a/src/ui/zabapgit_css_theme_dark.w3mi.data.css
+++ b/src/ui/zabapgit_css_theme_dark.w3mi.data.css
@@ -20,6 +20,9 @@
   --theme-greyscale-light: #CCCCCC;
   --theme-greyscale-lighter: #E5E5E5;
   --theme-list-hover-background-color: black;
+
+  --theme-table-background-color: #333333;
+  --theme-table-head-background-color: #202020;
 }
 
 /* GLOBALS */
--- a/src/ui/zabapgit_css_theme_default.w3mi.data.css
+++ b/src/ui/zabapgit_css_theme_default.w3mi.data.css
@@ -7,6 +7,7 @@
   --theme-container-background-color: #f2f2f2;
   --theme-container-border-color: lightgrey;
   --theme-table-background-color: white;
+  --theme-table-head-background-color: #edf2f9;
   --theme-table-border-color: #ddd;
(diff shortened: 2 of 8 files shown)


mbtools:
@christianguenter2 you're the master of link hints ;-). `f` does make sense for the filter field. after this, `f` would not anymore as a link hint on the overview. Are you ok changing to another key? `L` ?

https://github.com/abapGit/abapGit/blob/bf8c3590a854911f8a77fbe426b4e1b183eb5990/src/zcl_abapgit_settings.clas.abap#L381-L383

sbcgua:
Yeah, I was going to propose the same. F stands very well for Filter or Find ... Would be great if we could change the defaults for the link hints

christianguenter2:
> @christianguenter2 you're the master of link hints ;-). `f` does make sense for the filter field. after this, `f` would not work anymore as a link hint on the overview. Are you ok changing to another key? `L` ?
> 
Sure go ahead.

mbtools:
I changed the Vimium default to `t` since that must be working well :-)

Issue #5847, opened 2022-11-01
Add method to change the menu for HOC page


mbtools:
For pages created with `zcl_abapgit_gui_page_hoc=>create`, the menu is defined at the time of creation. It does not seem possible to update the menu afterward (or am I missing something @sbcgua?).

Add a method to update the menu - for example, to add items or switch checkmarks - and show the new menu after re-render. This will be essential when migrating more pages to the HOC model.

mbtools:
I have some custom pages where I need this but here's a similar example on our Diff page:

![image](https://user-images.githubusercontent.com/59966492/199562427-39952df2-be15-43dd-a466-9147eee2e259.png)

The menu options are handled by `on_event` and trigger an update of the menu:

https://github.com/abapGit/abapGit/blob/b03a70228f15979e7eac22e06518f9c413cf764d/src/ui/zcl_abapgit_gui_page_diff.clas.abap#L1332-L1337

Callback or an update_menu method on higher level would work.

PR #5860, opened 2022-11-06
Page menu provider for GUI page (and HOC)

sbcgua:
Would this fit for #5847 ? @mbtools 

P.S. I have doubts about naming. Theoretically, there may be more callbacks from top to bottom. And theoretically, it would make sense to unite or unify them somehow. Then it will not be `menu_provider` but some kind of `gui_page_callbacks`. But since there is no such case at the moment, let's not overengineer.


mbtools:
Haven't tested et but looks good. Naming maybe ’page_callback’ and ´zif_abapgit_gui_callbacks’ (with currently one ’get_menu’ method)

sbcgua:
I'd not change it now. Because there are no other callbacks yet. And it is not obvious which will come up. And if they should be merged together. But ... if they do, it might result in renaming and refactoring.

Anyway this is a good prototype. Pls check if it fits you, if yes, let's merge.

mbtools:
Perfect. Thanks!

PR #5861, opened 2022-11-07
Repo overview: Add labels and links to command palette

christianguenter2:
Make labels and 'Only Favorites' + 'Detail' accessible via command palette (F1)
grafik
grafik
grafik

Note, this change introduces CSS class `command` to distinguish anchors which should be accessible via command palette from which which should not. @sbcgua what do you think about that?
`enumerateUiActions` is currently quite messy and contains some redundancies. I plan to do some work on that in the near future. Maybe `command` can help there too to clean things up.

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -2441,7 +2441,7 @@ function enumerateUiActions() {
     });
 
   // command links
-  [].slice.call(document.querySelectorAll("a[class*='command']"))
+  [].slice.call(document.querySelectorAll("a.command"))
     .filter(function(anchor){
       return !!anchor.text;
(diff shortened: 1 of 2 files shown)


sbcgua:
@christianguenter2 nice feature ! certainly better UX integrity.

> CSS class `command`

Imho, CSS is for styling. Finding a class without a CSS content is confusing. I actually stumbled upon this during repo-over refactoring - the table lines were marked with `repo` class, yet had no styles. Appeared the class was used as a selector in JS navigation. I changed the selector to use table class + tr. Which works well and, I hope, makes more sense semantically. Yet, I admit, less readable :((( So there were pros and cons ... maybe there is a better solution.
Anyway I think it is not critical, just a method to achieve a good goal. So no prob, thanks for the feature :+1:

PR #5864, opened 2022-11-08
Repo view: Render labels

christianguenter2:
Also render labels on repo view.
grafik


mbtools (approved):
👍

sbcgua (approved):
:+1: Nice addition. I was hoping someone will take on this :)

Thoughts:
1) I think longterm (separately) `ls_settings = ... lo_label_colors = ...` part should move to the chunk code. This code looks superfluous here, it will repeat each time someone need to render labels, so it's a part of the chunk natively.
2) I think labels do not go well with path ...
image
But ... the reason is that this is a wrong place for the path (imho), not for the labels. And also path is rarely-used (as far as I can judge - no new PRs/changes there ...)

Just thoughts, no changes needed, the PR looks good.

PR #5865, opened 2022-11-08
Command palette refactoring

christianguenter2:
cleanup `enumerateUiActions` a bit


sbcgua (approved):
Looks good. Minor doubt in light of this comment: https://github.com/abapGit/abapGit/pull/5861#issuecomment-1312462322
But I think it is OK.

PR #5880, opened 2022-11-21
Fix type for data config serialize

mbtools:
Avoids implicit string > xstring > string conversion of JSON data


larshp (approved):
good catch

PR #5889, opened 2022-11-25
Repo View: Consolidate "Stage" and "Force Stage"

mbtools:
First part of https://github.com/abapGit/abapGit/issues/5013

"Stage" will be visible as a menu item if there are any diffs. As usual, you select which files to actually push on the stage page.


christianguenter2 (approved):
👍

PR #5892, opened 2022-11-27
Fix initial focus for link hints

christianguenter2:
When link hints are active initial focus is set to first link which is the abapGit logo. Pressing enter then navigates to repo overview. As this often happens accidentally this is a bit annoying. Setting initial focus to a element without an action is much better.
grafik


mbtools (approved):
Indeed

PR #5896, opened 2022-11-29
page repo: add patch function

christianguenter2:
Can save one step
grafik


mbtools:
that was always missing. ty

PR #5905, opened 2022-12-01
Remove references to `cx_oo_clif_scan_error`

### LINKED ISSUE #5903: Syntax Error in 7.02: `cx_oo_clif_scan_error` does not exist in 7.02
I think #5841 introduced a syntax error in 7.02.

Class `cx_oo_clif_scan_error` is only available in releases >= 7.31. Maybe, both usages can be replaced by its super class `cx_oo_source_save_failure` which is also available in 7.02.

schneidermic0:
Fixes #5903.

### DIFF
--- a/src/objects/oo/zcl_abapgit_oo_class.clas.abap
+++ b/src/objects/oo/zcl_abapgit_oo_class.clas.abap
@@ -313,7 +313,7 @@ CLASS zcl_abapgit_oo_class IMPLEMENTATION.
   METHOD update_meta.
 
     DATA: lo_update     TYPE REF TO cl_oo_class_section_source,
-          lx_error      TYPE REF TO cx_oo_clif_scan_error,
+          lx_error      TYPE REF TO cx_oo_save_failure,
           ls_clskey     TYPE seoclskey,
           lv_scan_error TYPE abap_bool.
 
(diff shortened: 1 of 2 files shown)


schneidermic0:
Another alternative might be to access the exception dynamically as it is done here:
https://github.com/abapGit/abapGit/blob/c6e026350b02dc17a5f44e3d022f10b1a3d0c5ff/src/objects/oo/zcl_abapgit_oo_class.clas.abap#L229-L239

However, since it seems we don't want to access any attributes, the other fix seems to be sufficient

schneidermic0:
Is there anything we should do about the abaplint observation? We could also replace `cx_oo_save_failure` by another exception

mbtools:
I checked. `cx_oo_save_failure ` does not exist in 702 either. The exception isn't even class-based. Here's the method sig:

```abap
  methods REVERT_SCAN_RESULT
  abstract
    importing
      value(SUPPRESS_ACCESS_PERMISSION) type SEOX_BOOLEAN default SEOX_FALSE
    exceptions
      REVERT_ERROR .
```

Just replace it with `cx_root` to avoid the syntax error in 702.

schneidermic0:
Sorry, my fault. It should have been `cx_oo_source_save_failure`. It exists is 7.02. 
However, I can replace it by cx_root, too.

Which option do you prefer?

schneidermic0:
Short addition: Method signature in new releases:

```
  methods REVERT_SCAN_RESULT
  abstract
    importing
      value(SUPPRESS_ACCESS_PERMISSION) type SEOX_BOOLEAN default SEOX_FALSE
    raising
      CX_OO_CLIF_SCAN_ERROR
      CX_OO_SOURCE_SAVE_FAILURE.
```


And `cx_oo_clif_scan_error` inherits fro `cx_oo_source_save_failure`

```
class CX_OO_CLIF_SCAN_ERROR definition
  public
  inheriting from CX_OO_SOURCE_SAVE_FAILURE
  create public .
```

mbtools (approved):
Tested in 702

PR #5910, opened 2022-12-02
Local settings: add F4 value help for check varint

christianguenter2:
before
grafik

after
grafik
grafik


sbcgua (approved):
Works nicely :)

PR #5914, opened 2022-12-02
Diff/Patch: Improve beacons

### LINKED ISSUE #3973: Diff/Patch: Description in beacon
Not sure how it works but shouldn't the `@@ CLASS ... IMPLEMENTATION => METHOD method_name` be rendered directly above the `METHOD method_name` line?

image

mbtools:
1. Cases like `report = 'SAPMSVIM'` were detected as beacons:

image

Here's another example. Entries 2 and 6 should not be beacons:

image

Updated regex avoids such cases. 

2. Shorten beacon for methods from `CLASS xyz IMPLEMENTATION => METHOD abc` to `CLASS xyz => METHOD abc`

3. Add beacons for `CLASS xyz PUBLIC|PROTECTED|PRIVATE SECTION`

4. Add unit test.

5. Make beacon reference beginning of diff instead of 8 lines before. 
Closes #3973


christianguenter2 (approved):
👍

PR #5915, opened 2022-12-06
Copy (yank) link texts with y+LinkHints

christianguenter2:
grafik

abapGit_yank

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -1461,7 +1461,7 @@ LinkHints.prototype.handleKey = function(event){
     this.pendingPath += event.key;
     var hint = this.hintsMap[this.pendingPath];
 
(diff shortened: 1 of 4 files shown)


src/ui/zabapgit_js_common.w3mi.data.js
```diff
@@ -1455,10 +1461,15 @@ LinkHints.prototype.handleKey = function(event){
     this.pendingPath += event.key;
     var hint = this.hintsMap[this.pendingPath];
 
-    if (hint) { // we are there, we have a fully specified tooltip. Let's activate it
+    if (hint) { // we are there, we have a fully specified tooltip. Let's activate oder yank it
```
mbtools, line 1464: a bit of German, oder? ;-)

mbtools (approved):
Fancy feature 😄

src/ui/zabapgit_js_common.w3mi.data.js
christianguenter2: heh, good catch

PR #5925, opened 2022-12-08
Repo View: Fallback to main if branch doesn't exist

mbtools:
Instead of getting an error if the selected branch (or tag) does not exist, entering the repository view will automatically switch to the main branch (head) and issue a corresponding message in the status bar.

Old:

image

New:

image

Closes https://github.com/abapGit/abapGit/issues/5331


christianguenter2 (approved):
quite useful addition. Works.
👍

PR #5927, opened 2022-12-09
Improve repository URL validation

### LINKED ISSUE #4319: Issue with error handling in new online dialog
- Use a git host where there is not automatic redirect between user/repo and user/repo.git
- New online
- Enter url without .git
- Press clone
- Get a 404 http error, the clone dialog is still open
- Add .git to url
- Press clone again
- Get error message "Repository user/repo.git is already versioned as ... by ..." and "Package ... is already versioned as ... by ..."

After cancelling the dialog you can see the repo was already created and you need to change the remote there to fix the .git suffix. It is not transparent the repo was created because the 404 error message is issued in the dialog. I also noticed the repo is not added to the favorites.

mbtools:
Certain providers require that repo URLs end with `.git`. The new online repo and remote settings dialogs now validate entered URLs in this regard. Currently, the check is done for GitLab, only. If other providers require the same check, we can add them easily. 

image

Closes #4319

### DIFF
--- a/src/git_platform/zcl_abapgit_git_url.clas.abap
+++ b/src/git_platform/zcl_abapgit_git_url.clas.abap
@@ -13,7 +13,7 @@ CLASS zcl_abapgit_git_url DEFINITION
       RAISING
         zcx_abapgit_exception .
 
-    METHODS validate
+    METHODS validate_url
       IMPORTING
         !iv_url TYPE string
       RAISING
@@ -90,7 +90,7 @@ CLASS zcl_abapgit_git_url IMPLEMENTATION.
   ENDMETHOD.
 
 
-  METHOD validate.
+  METHOD validate_url.
 
     DATA lv_provider TYPE string.
 
--- a/src/git_platform/zcl_abapgit_git_url.clas.abap
+++ b/src/git_platform/zcl_abapgit_git_url.clas.abap
@@ -96,10 +96,16 @@ CLASS zcl_abapgit_git_url IMPLEMENTATION.
 
     lv_provider = zcl_abapgit_url=>host( to_lower( iv_url ) ).
 
+    " Provider-specific check for URLs that don't work
     IF lv_provider CS 'gitlab.com'.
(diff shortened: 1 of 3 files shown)


larshp (approved):
lgtm

src/git_platform/zcl_abapgit_git_url.clas.abap
```diff
@@ -81,4 +88,20 @@ CLASS ZCL_ABAPGIT_GIT_URL IMPLEMENTATION.
...
   ENDMETHOD.
+
+
+  METHOD validate_url.
+
+    DATA lv_provider TYPE string.
+
+    lv_provider = zcl_abapgit_url=>host( to_lower( iv_url ) ).
+
+    IF lv_provider CS 'gitlab.com'.
```
sbcgua, line 99: Are you sure this is only gitlab behavior ? My feeling is that, in opposite, just the gitHUB allows omitting the `.git suffix`. not sure for 100% ... but `.git` suffix was there first ... so I suspect most providers just follow this tradition

larshp: check bitbucket + azure devops, think that covers 90% of the market then

sbcgua: Yeap ... ideally check each vendor if the allow omitting `.git`
And if not, the it's statistically `if NOT provider cs 'github.com'`

christianguenter2: bitbucket works without `.git`

g-back: What about silently just adding a ".git"-suffix if it's not there? Then no provider-specific checks have to be added?

sbcgua: I like the idea with auto suffix.

mbtools: that's no guarantee either... Azure DevOps with `.git`:

image

PR #5951, opened 2022-12-20
Repo View: Consolidate "Pull" and "Selective Pull"

### LINKED ISSUE #5013: Consolidate "Pull" and "Stage" Options

mbtools:
1. "Pull" remains visible on the toolbar
   
   Exception: There are no diffs. 

2. When clicking "Pull", it runs what was previously "Selective Pull" 

   A popup asking to confirm which objects should be pulled is shown.

   Exception: There are only new objects, for example, when doing the initial pull of a repo. In this case, no popup is shown and the objects are pulled automatically.

3. "Advanced > Selective Pull" has been removed from the toolbar

Closes #5013

- [ ] To-do: Update docs

### DIFF
--- a/src/ui/pages/zcl_abapgit_gui_page_repo_over.clas.abap
+++ b/src/ui/pages/zcl_abapgit_gui_page_repo_over.clas.abap
@@ -412,7 +412,7 @@ CLASS zcl_abapgit_gui_page_repo_over IMPLEMENTATION.
 
     lo_toolbar->add(
       iv_txt      = |Pull|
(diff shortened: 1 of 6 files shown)


sbcgua (approved):
seems to work :)

PR #5963, opened 2022-12-29
APACK user exit

simonegaffurini:
If a manifest class implementing either ZIF_APACK_MANIFEST or IF_APACK_MANIFEST, it will fallback to the user exit, that, if implemented, will provide the APACK values.


mbtools:
just `...get_apack_manifest_descriptor( iv_package_name ).`

PR #5972, opened 2023-01-04
Fix menus with many items

mbtools:
Menus with many items are not accessible if the menu becomes longer than the screen height.

Before:

image

After:

image

Note: Setting a relative max-height does not seem to be feasible since it would be relative to the size of the sticky area.


christianguenter2 (approved):
👍

PR #5979, opened 2023-01-09
Allow Object Types with multiple files in AFF

wurzka:
With this change, the abapGit handler class of object types whose AFF files do not consist only of the json file (e.g., [DTEB](https://github.com/SAP/abap-file-formats/tree/main/file-formats/dteb)) can also inherit from super class zcl_abapgit_object_common_aff.
The sub class only has to redefine the method get_additional_extensions and fill it with the extension name and the file name mapper.


mbtools:
looks good. can we do a test repo already in https://github.com/abapGit-tests?

wurzka:
Thanks for your review, @mbtools. Would it be fine to you if I added a test repo for the object type DTEB? So far, there is no abapGit object handler for it but this change was requested by its object type owner.

mbtools:
Sure. Go ahead and add it. I will just have to adjust our CI process to skip the repo until the object handler is implemented.

wurzka:
Example is uploaded in https://github.com/abapGit-tests/DTEB

mbtools:
I'm ok with this but we'll let @larshp take a look too

larshp (approved):

PR #5986, opened 2023-01-12
Update AFF to new AJSON functionality

albertmink:


albertmink:
Manual testing:

- [ ] pull into ABAP system, Object looks good
- [ ] diff page for JSON looks good
- [ ] push from ABAP system

src/objects/aff/zcl_abapgit_json_handler.clas.abap
```diff
@@ -161,15 +159,15 @@ CLASS zcl_abapgit_json_handler IMPLEMENTATION.
       lv_original_language TYPE sy-langu.
 
 
-    lv_iso_language = co_ajson->get_string( '/header/originalLanguage' ).
+    lv_iso_language = co_ajson->get_string( '/header/original_language' ).
```
schneidermic0, line 162: Just for my understanding. We can change it here to the notation with underscore, because we changed the mapping above to `zcl_abapgit_ajson_mapping=>create_to_snake_case( )`, right?

src/objects/aff/zcl_abapgit_json_handler.clas.abap
```diff
@@ -237,40 +235,28 @@ CLASS zcl_abapgit_json_handler IMPLEMENTATION.
...
     map2json_abap_language_version( CHANGING co_ajson = lo_ajson ).
     map2json_custom_enum( EXPORTING it_enum_mappings = iv_enum_mappings
                           CHANGING co_ajson          = lo_ajson ).
 
     CREATE OBJECT lo_filter EXPORTING iv_skip_paths = iv_skip_paths.
-    lo_ajson_filtered = zcl_abapgit_ajson=>create_from(
-                          ii_source_json = lo_ajson
-                          ii_filter      = lo_filter ).
-
-    lo_ajson_filtered->keep_item_order( ).
```
schneidermic0, line 268: Is `keep_item_order` the default in the new API? Or can we skip this?

schneidermic0 (commented):
I added some small questions. And do we need to adapt also following path?
https://github.com/abapGit/abapGit/blob/47dbee1d4884d85b3e66f1a482f3d827349d9a40/src/objects/zcl_abapgit_object_intf.clas.locals_imp.abap#L601

src/objects/aff/zcl_abapgit_json_handler.clas.abap
albertmink: actually, in the old version this was a 'bug' ( `create_from` did not preserved the keep_item_order setting ). So, this is an intentional change.

src/objects/aff/zcl_abapgit_json_handler.clas.abap
albertmink: Yes, the time of `when` the mapping is actually performed changed. So, on a parsed json data with mapping, the field names are mapped already. Not later, e.g. `to_abap` as before.

schneidermic0 (approved):
If we don't need to adapt the implementation for INTF, the changes look good to me.

Thanks @albertmink.

PR #5998, opened 2023-01-16
DOMA: Fix domains with custom conversion exit

mbtools:
Deserializing a domain that is using a custom conversion exit (custom `CONVERSION_EXIT_<name>_IN|OUTPUT` functions) is failing on the initial pull since the functions do not exits.

image

Fix is using two-phase approach: 

1. DDIC phase: 
    If function does not exit, remove the exit function
1. LATE phase:
    If function was removed, change it to the correct exit function

New test repo: https://github.com/abapGit-tests/DOMA_with_exit


christianguenter2 (approved):
👍

PR #5999, opened 2023-01-16
SHLP: Avoid double activation

mbtools:
Refactor handling of dependencies on exit function (using same methods as `DOMA` #5998). This avoids activating the search help twice if the exit function already exists.

Recap of two-phase approach:

1. DDIC phase:
    If function does not exit, replace it with a standard SAP function
2. LATE phase:
    If function was replaced, change it to the correct exit function

Existing test repo: https://github.com/abapGit-tests/SHLP_with_exit


christianguenter2 (approved):
👍

PR #6019, opened 2023-01-31
Update object descriptions - Part 2

mbtools:
Some housekeeping... continues https://github.com/abapGit/abapGit/pull/5314

Standardize descriptions with "abapGit - " prefix


mbtools:
I was hopeful to optimize deserialize so it would work only on changed files... i.e. only xml changes but not abap. Not so easy so unfortunately, pulling this will take a while.

larshp (approved):

PR #6062, opened 2023-02-10
Fix hotkeys repo list

mbtools:
Side effect of #6050


sbcgua:
Hmmm, these hotkeys relate to another component, or do I miss something ?

mbtools:
page_main got merged into page_repo_over page. i had left the ui_component as `main` for the top toolbar but `repo overview` will work, too.

PR #6087, opened 2023-02-24
Fix zip diff command visibility

sbcgua:
When importing from ZIP it happens that only local changes are available. I could not easily recreate it, but it was connected to extra translations in local systems. The below screenshot is simulated. In this case Diff (and also Pull) command is not visible, though it is very important to control what's wrong with the marked objects. Just had an issue with it yesterday when deploying our product.
image

So the fix just shows Diff/Pull for any `unchanged` status
2023-02-24_15h29_33


mbtools (approved):
👍

PR #6089, opened 2023-02-24
Refactor popup_to_select_from_list

sbcgua:
It appeared, that `popup_to_select_from_list` holds 30% of `zcl_popup`, most of private methods and most of attributes ... which are kind of global vars in this way. So it is a class in class. And I separated it into a local class for decoupling and better code structure. Would be nice to merge it soon, cause I have other plans to improve this dialog. It's just this refactoring is already huge and not to mix it with the feature.

It is mostly move and restructure the code. Changes are:
- the dialog does not support different structures on in and out anymore (why should it ?! )
- the object category button is now exclusive, deselects all "other" objects before marking the selected category. I think it was a bug at the first place.

The rest is the same just in better structure

### DIFF
--- a/src/apack/zcl_abapgit_apack_helper.clas.abap
+++ b/src/apack/zcl_abapgit_apack_helper.clas.abap
@@ -80,7 +80,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_apack_helper IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_APACK_HELPER IMPLEMENTATION.
 
 
   METHOD are_dependencies_met.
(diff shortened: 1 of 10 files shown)


mbtools:
ref #6009 

It intentionally didn't deselect when using the category button. The way it was you can easily select "DEVC + DDIC", for example, or "Code + Enhancements". 

Maybe `textline1` could be better: `Which objects should be added to the selection?`

sbcgua:
> It intentionally didn't deselect when using the category button. The way it was you can easily select "DEVC + DDIC", for example, or "Code + Enhancements".

Aha ! So that was intentional. Ok, fixing it back. Maybe let's remove the "All objects" point ? It is equivalent to select all which is a dedicated button

> Otherwise, the change makes sense (unless you want to move the complex popup to its own global class).

I don't see the reason, at least for now. There is a common popup interface and etc. It's more internal implementation details

mbtools:
true. no need for "all objects"

sbcgua:
done

mbtools:
Tested and it works well except the popups are not wide enough since package column was added:

image

In `zcl_abapgit_services_repo->popup_overwrite` and `popup_package_overwrite`, pass `iv_end_column = 110`. Then it should be ok:

image

sbcgua:
Hmm, seems like we have different screen scale settings. But yes, I'll fix it. Wider table is better anyways

sbcgua:
Done (default value 110, it is not explicitly passed from the callers so default change should be ok)

mbtools (approved):
👍

PR #6093, opened 2023-02-25
Better pull dialog, Screen status

### LINKED ISSUE #6090: Improve pull decision dialog, deployment mode
I find that the pull decision dialog is actually a very nice feature, showing deeper details on what will be done and gaining some control over deserialization sequence. It helps deploying code a lot. Yet there is a room for improvement. For me it boils down to a possibility to select a particular sub-package and deserialize only it, and then next and then next. In a wider sense - there should be a possibility to select/filter groups of object and process them only. The new feature that allows selecting categories of objects (DEVC, DDIC ...) is not granular enough.

Fortunately it is very easy to do :) Having that the dialog is an ALV :)

I'm going to rework the toolbar and that's it. But there is one pitfall - screen status. The existing screen status does not have more buttons. The standard SALV status, unexpectedly, does not have Cancel button (X). Any ideas about finding a reliably standard dialog screen status ?

Another option is to create one. And the standalone version should just generate it on the fly. Like the abapgit table. Or fixing the 1001 screen. Seems possible.

sbcgua:
Closes #6090

Don't look at the code yet, it is based on the #6089 branch

- [X] Screen status
- [X] autocretation of CUA
- [X] abapmerge stuff - https://github.com/larshp/abapmerge/pull/366
- [x] wiring, testing and finetuning 
- [x] change detection (how to recreate, if GUI status was updated)

## Implementation details:
- `zcl_abapgit_objects_program=>ty_cua` became public
- `zcl_abapgit_persist_migrate` got 2 new methods `gui_status_create/exists`
- ... and a local class that inherits from `zcl_abapgit_objects_program` to reuse `serialize/deserialize cua` (**pls review the activation logic**)
- ... and a local class `lcl_own_cua_provider` with `@@abapmerge include-cua` pragma
- idea is that migration checks the existance of GUI status at startup and creates it if missing
- the dialog ALV logic is rather simple, reuses the new GUI status and falls back to `SAPMSVIM` in case of something unusual
- GUI status `DECIDE_DIALOG` is available for further improvements if any

## Demo

https://user-images.githubusercontent.com/15635498/221364038-8ef25b5d-a59c-42cb-9ce2-ecb0e57802b7.mp4


sbcgua:
OK, ready for review and test conceptually. I added changes description to the main message above. The testing should be done both over the full AG (should be no problem as the GUI status is a part of XML) and the standalone version (must use new feature of abapmerge - https://github.com/larshp/abapmerge/pull/366).

One unresolved point is **detection of CHANGES**. So that if I install AGS, run it (CUA autocreated), then update the AGS, and the CUA has been changed in the new code, but it is already created - then there is no clear trigger to recreate. Actually no big issue - just have to compare the new and existing structure.

**Question**: is there a standard way to compare 2 deep structures ? Any FM or class ? Currently I'm considering `import/export from data buffer` + `zcl_abapgit_hash=>sha1` ...

sbcgua:
2023-02-26_20h36_30

Also added state column

christianguenter2:
Debugger does it with `CL_TPDA_DIFF*` classes maybe that's helpful.

sbcgua:
> Debugger does it with `CL_TPDA_DIFF*` classes maybe that's helpful.

Mmmm, seems like it goes via stack in some indirect way ... I ended up with `sha1` way. Actually works well. But thanks anyway.

So. I guess now it is really ready for test and review

sbcgua:
lint-and-merge error is due to old abapmerge
unit test error ... not sure, no new UTs were created

larshp:
https://github.com/open-abap/open-abap-core/pull/564

mbtools:
Tested... on 702 😁 It works! 👍 

image

Small issues:

- Not wide enough
- "Mark selected F7" does not work and selects everything

The following work better:

```abap
  METHODS popup_to_select_from_list
    IMPORTING
...
      !iv_start_column       TYPE i DEFAULT 10
      !iv_end_column         TYPE i DEFAULT 125
      !iv_start_line         TYPE i DEFAULT 10
      !iv_end_line           TYPE i DEFAULT 30
```

image

sbcgua:
Or maybe just rename to "Mark visible"

mbtools:
oh, it's almost twice as wide for your screen (mine is 1920x1080). 

Maybe call it "Select filtered objects (F7)"

"Select highlighted objects" would be a good addition since clicking a checkbox is a round-trip:

image

sbcgua:
OK, both work. If anything is selected then it is marked, if nothing - then all visible (after filter) are selected. Text is "Mark selected or all visible" - should be a bit clearer

sbcgua:
I also have 1920x1080, scaling 125% ... anyway, I think this dialog wins from size ...

sbcgua:
Actually one usability concern ... intuitively, I'd expect that "Select All" (and also Deselect All) is applied to **visible** records ... Although it is not a hard rule and also a matter of habit, but may be a source of confusion ...
Just thoughts ...

sbcgua:
Just in case, no further plans from my side to add anything here. The #6107 will certainly be a separate PR (PRs). The "usability concern" also (if necessary at all).

mbtools:
lgtm 😃

christianguenter2:
I work with this change since a few days in dev and standalone edition, no problems so far. Good Job. 👍

mbtools (approved):
Let's go for it!

sbcgua:
:tada:

sbcgua:
Deployments just became easier :)

mbtools:
just realized what was odd... "OK" (the default) should be on the left:

image

sbcgua:
I can change, but this is arguable. In fact I move them to the right intentionally. Because if it's pressed by mouse it is always on the same place, in particular:
- in case of dynamic tool bar change (filter reset button)
- in case of new button addition. Both standard alv or new functions - e.g. I want to add a possibility to change package assignments only and this will probably be a separate button, not a filter, because it will need to pass specific params to the deserializer

But if there is a grounded or massive opinion that it's better be the left-most, I can change

mbtools:
default (main) button on left matches our HTML form design.

sbcgua:
Hmm maybe a matter of a better hotkey ? I mean key navigation is a very good reason. But TAB-TAB-Enter is then sub-optimal. It's better to have Ctrl+Enter (or similar) at any place of the screen ...

PR #6095, opened 2023-02-26
Fix tag deletion

mbtools:
Deletion of lightweight as well as annotated tags was not working correctly. 

Closes https://github.com/abapGit/abapGit/issues/6072


christianguenter2 (approved):
Works 👍 
Btw. what does `peel` mean?

mbtools:
"Peel" is a term in the git pack protocol https://github.com/git/git/blob/master/Documentation/gitprotocol-pack.txt. I have no idea how someone came up with this. 🤷

PR #6102, opened 2023-02-26
CLAS: Fix error when changing redefinitions

mbtools:
Avoid "Error updating where-used list for CLAS .... Check for syntax errors" and need to manually clean-up method redefinitions. 

Test repo: https://github.com/mbtools/TEST_REDEF, switch between `main` and `redef` branches and pull.

Note: If clean-up is triggered, some class sections might be regenerated. This can lead to diffs. Another pull will then correct the class. 

Closes https://github.com/abapGit/abapGit/issues/6067


larshp (approved):
looks good to me

lets try this

PR #6121, opened 2023-03-02
Fix ITAB_DUPLICATE_KEY dump during deserialize

mbtools:
Dumps when object is moved to different package

Regression #6093

### DIFF
--- a/src/objects/core/zcl_abapgit_file_status.clas.abap
+++ b/src/objects/core/zcl_abapgit_file_status.clas.abap
@@ -151,7 +151,7 @@ ENDCLASS.
 
 
 
-CLASS ZCL_ABAPGIT_FILE_STATUS IMPLEMENTATION.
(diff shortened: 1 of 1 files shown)


sbcgua:
Interestingly I didn't change this code, just moved it, literally from previous line - https://github.com/abapGit/abapGit/pull/6114/files#diff-7a848ddfe87ea2395b7a3bbf2beb14a8e2bb7bd32b1dafa69918d1b3c91f3bbbL596

But yeah, thanks for the fix. More UTs will save us :P

mbtools:
Added screenshots. Looks like data in `lt_item` was not unique before ;-)

sbcgua:
Yes, I noticed that too. Actually thats why that sort is there ... anyway, it ok, just strange

sbcgua:
I guess fix is for good anyway, because we are looking for unique objects. Package was superfluous there.

PR #6124, opened 2023-03-03
File status refactoring, part 2

sbcgua:
to #6107

OK, this one is also mainly refactoring but also already some additions

1) Refactor and simplify `rebuild_checksums`
  - it is called from `new_on/offline`, user `refresh_local_checksum` command
  - `iv_branches_equal` param was not used
  - in fact the rebuild didn't use the remote, so I dropped remote-related logic
  - now it is simpler, it just takes the current state and copies it to the checksums
  - and a comment was left there as well explaining the logic why
2) `zcl_abapgit_filename_logic` - extract obj definition file detection (more a refactoring to reuse in file status)
3) status checks were moved to a local class. status calculation and checks are actually different roles so they deserve separate classes. Otherwise the code becomes a bit messy, too few space for both in one class. But no need for a new global class as well
4) file_status class move to REPO package, item_state helper moved to UTILS
5) finally `zif/cl_abapgit_sap_namespace` is created to abstract the select, and respecting factory stuff

So no significant logic changes except simplification of `rebuild_checksums`

will continue in part 3 ...


src/repo/zcl_abapgit_file_status.clas.locals_imp.abap
```diff
@@ -0,0 +1,301 @@
...
+    FIELD-SYMBOLS <ls_result> LIKE LINE OF it_results.
+
+    " Collect all namespaces based on name of xml-files
+    LOOP AT it_results ASSIGNING <ls_result>.
+      FIND REGEX '^#([a-zA-Z0-9]+)#.*\..*\.xml$' IN <ls_result>-filename SUBMATCHES lv_namespace.
+      IF sy-subrc = 0.
+        lv_namespace = '/' && to_upper( lv_namespace ) && '/'.
+        COLLECT lv_namespace INTO lt_namespace.
+      ENDIF.
+    ENDLOOP.
```
mbtools, line 190: needs to get fixed since it does not work with AFF `json` which also uses `( )` instead of `#` for namespace in filename (was wrong before but not noticed)

src/zcl_abapgit_sap_namespace.clas.abap
```diff
@@ -0,0 +1,28 @@
...
+    INTERFACES zif_abapgit_sap_namespace.
+  PROTECTED SECTION.
+  PRIVATE SECTION.
+ENDCLASS.
+
+
+
+CLASS ZCL_ABAPGIT_SAP_NAMESPACE IMPLEMENTATION.
+
+  METHOD zif_abapgit_sap_namespace~namespace_exists.
```
mbtools, line 16: `zif_abapgit_sap_namespace~exists`

src/zcl_abapgit_sap_namespace.clas.abap
```diff
@@ -0,0 +1,28 @@
...
+    DATA lv_editflag TYPE trnspace-editflag.
+    SELECT SINGLE editflag FROM trnspace INTO lv_editflag WHERE namespace = iv_namespace.
+    rv_yes = boolc( sy-subrc = 0 ).
+  ENDMETHOD.
+
+  METHOD zif_abapgit_sap_namespace~namespace_is_editable.
```
mbtools, line 22: `zif_abapgit_sap_namespace~is_editable`

src/zif_abapgit_sap_namespace.intf.abap
```diff
@@ -0,0 +1,16 @@
+INTERFACE zif_abapgit_sap_namespace
+  PUBLIC .
+
+  METHODS namespace_exists
```
mbtools, line 4: `exists`

src/zif_abapgit_sap_namespace.intf.abap
```diff
@@ -0,0 +1,16 @@
...
+    IMPORTING
+      iv_namespace TYPE trnspace-namespace
+    RETURNING
+      VALUE(rv_yes) TYPE abap_bool.
+
+  METHODS namespace_is_editable
```
mbtools, line 10: `is_editable`

src/repo/zcl_abapgit_file_status.clas.locals_imp.abap
mbtools: see `zcl_abapgit_filename_logic->file_to_object`

sbcgua:
renamings: agreed, done. For the namespece check see the comment above

src/repo/zcl_abapgit_file_status.clas.locals_imp.abap
sbcgua: Probably yes, but I'm far from the topic. I suggest keeping it in this PR as it. In "part3" I'll do some adjustments to checks and I'll create the branch in this repo so you could commit the change as well.

mbtools: keep as is. reminder: https://github.com/abapGit/abapGit/issues/6126

sbcgua:
just in case: no plans to add anything to this PR. ready to merge when approved.

mbtools (approved):
👍

PR #6128, opened 2023-03-03
File status refactoring, sidetrack: show local state for offline repos

sbcgua:
A sidetrack from #6107 - a feature to show a change between actual serialized state and previously saved checksums. Sometimes useful. Although obviously functionality is limited as there are no diffs, just the fact that something was changed.

Not ready for review yet, waits for #6124 + need some adjustment in file_status logic

https://user-images.githubusercontent.com/15635498/222825035-ec62dcd2-27be-413e-a5f7-5eb22892f022.mp4

### DIFF
--- a/src/repo/zcl_abapgit_repo_content_list.clas.abap
+++ b/src/repo/zcl_abapgit_repo_content_list.clas.abap
@@ -5,18 +5,25 @@ CLASS zcl_abapgit_repo_content_list DEFINITION
 
   PUBLIC SECTION.
     METHODS constructor
-      IMPORTING io_repo TYPE REF TO zcl_abapgit_repo.
+      IMPORTING
+        io_repo TYPE REF TO zcl_abapgit_repo.
 
     METHODS list
-      IMPORTING iv_path              TYPE string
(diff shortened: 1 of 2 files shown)


mbtools:
do we need the extra menu option? why not always show that something was changed (compared to the last zip import)?

sbcgua:
Interesting question. Maybe you're right. I'll have a look with fresh head tomorrow. This should have been a quick feature of the evening :) But fighting with `file_state` took major part of concentration.

sbcgua:
Posted #6129 to discuss Marc's question

sbcgua:
after #6124, to avoid code conflicts

sbcgua:
Cancelling the this for now. The code is actually ready, but with the current state logic it's crappy (always `A`) and does not make an improvement. I tried to patch status calculation so that this works for offline repos and does not break for everything else but failed. The state logic has to be heavily refactored. I'll keep the branch and PR it later after the state refactoring is done. 

P.S. the switching was removed - I agree there is no sense in 2 modes here, just serialize always. Simpler is better.

image

image

Issue #6048, opened 2023-02-07
Refactor: Change remaining pages to HOC


mbtools:
Switch the following pages from `ZCL_ABAPGIT_GUI_PAGE` to `ZCL_ABAPGIT_GUI_COMPONENT`:

- [x] `ZCL_ABAPGIT_GUI_PAGE_DATA`
- [x] `ZCL_ABAPGIT_GUI_PAGE_DB`
- [x] `ZCL_ABAPGIT_GUI_PAGE_DB_DIS`
- [x] `ZCL_ABAPGIT_GUI_PAGE_DB_EDIT`
- [x] `ZCL_ABAPGIT_GUI_PAGE_DIFF` (`PATCH`)
- [x] ~`ZCL_ABAPGIT_GUI_PAGE_MAIN`~
- [x] `ZCL_ABAPGIT_GUI_PAGE_MERGE`
- [x] `ZCL_ABAPGIT_GUI_PAGE_MERGE_RES`
- [x] `ZCL_ABAPGIT_GUI_PAGE_RUNIT`
- [x] `ZCL_ABAPGIT_GUI_PAGE_STAGE`
- [x] `ZCL_ABAPGIT_GUI_PAGE_CODI_BASE` (`CODE_INSP`, `SYNTAX`)

General migration steps:

```abap
CLASS zcl_abapgit_gui_page_xyz DEFINITION
  PUBLIC
  INHERITING FROM zcl_abapgit_gui_component
  FINAL
  CREATE PRIVATE.

  PUBLIC SECTION.

    INTERFACES:
      zif_abapgit_gui_event_handler,
      zif_abapgit_gui_hotkeys, " if page has hotkeys
      zif_abapgit_gui_menu_provider, " if page has toolbar
      zif_abapgit_gui_renderable.

    CLASS-METHODS create
      IMPORTING
        ...
      RETURNING
        VALUE(ri_page) TYPE REF TO zif_abapgit_gui_renderable
      RAISING
        zcx_abapgit_exception.

    METHODS constructor
      IMPORTING
        ...
      RAISING
        zcx_abapgit_exception .

    " remove all REDEFINITIONS

```

```abap
  METHOD constructor.

    super->constructor( ).
    ....

    " remove ms_control-... assignments

  ENDMETHOD.

  METHOD create.

    DATA lo_component TYPE REF TO zcl_abapgit_gui_page_xyz.

    CREATE OBJECT lo_component
      EXPORTING
        ...

[...]

PR #6134, opened 2023-03-06
Refactoring of DB pages, and new UI features

sbcgua:
ref: #6048

... to switch my brain from that file_status stuff ...

This PR converts DB pages to reuse the HOC. Also `db_dis` and `db_edit `were merged into one - `db_entry`. Because it was semantically one page. Also allowed to get rid of `get_state_db_edit` in the router. A bit better code structure and html structure in general as well as slight CSS simplification. That all for this part of the subject.

## New UI features and experiments

I intentionally picked less critical pages to introduce the new mechanisms. It looks like a huge PR, yes it is, but it really impacts just the DB pages for the moment.
And also to be able to build complexity in steps (especially with the table component - which can be improved on later stages)

### Page based CSS

So CSS management in one file became difficult already for long. It makes sense to handle CSS per component.
- Both DB overview and DB entry pages have dedicated MIME object (name convention: `abapgit_css_page_...`).
- Editing fits well in W3MI_POLLER flow.
- For the standalone version it merges directly into the component class, thus couples code and style close together
- Note! Theming theoretically becomes more difficult because there is no differentiation anymore. But I think this is for good. On a deeper glance theming - is the set of color variables ! This requires some discipline with colors. Which is also easier to maintain in small files. So this is how it should be properly.

Technically:
- a component has an acces to `gui_services`
- which exposes `asset_manager->register_asset`
- which allows registering direct content (abapmerge) or the mime_object_name (like in UI_FACTORY)
- `page` class has a new attribute - `extra_css` (and also `extra_js` for future use) - which allows referring an additional css sheet (via the `HOC`)
- extra CSS is then taken care of by CSS processor
- all very aligned with the ui_core architecture :)

And I hope it makes sense in the light of future UI plans too.

### Table component

Second major feature is a framework class for tables. It is rather simple at the moment. Intentionally. I'd say don't use it for other refactoring yet. Next step would be to apply it in stage or repo_view or repo_over to implement filtering and other complex stuff.

Brief architecture:
- there is `zcl_abapgit_html_table` class, you initiate it passing the cell renderer interface (`zif_abapgit_html_table`)
- then you define columns (descriptively, like in the html_form)
- then you call render passing some global css styles, ids, but most importantly the data table
- the interface is supposed to be implemented by the component or its local class (!!! e.g. in case of several tables !!!). It is then called for extra info for each row, and to render each cell
- cell renderer receives the row index, row structure and, when applicable, field value
- cell renderer returns the content (string ot html instance) and the needed CSS classes and meta stuff
- this is very flexible for cell formatting. Also in fact repeats the `render_item` pattern, which can be seen in almost all pages, but without all the routine clutter `<table>, <tr>, <td> and stuff`
- cleaner and more standartised, also in the light of potential binding JS to the HTML structure on the client side

### Small stuff

- dedicted `validate_entry_type` in persistence_db to remove it from pages
- possibility to change page title from a component on re-render (page_title_provider)
- small improvements in `td/th` methods of `html`

TODO soon: describe new CSS approach in docs, +- move the above text there


mbtools:
Haven't tested so just general comments
- DB page refactor: 👍 
- Page-based CSS: I'm undecided if this is a good idea. Now one potentially has to hunt to find where the CSS is maintained. Ideally all CSS is component-based and pages are based on (various) components so should not need any special CSS. This is now going into the opposite direction. How many CSS files will we end up with? Same question for JS... I prefer to have just two central files. 
- Table-component: 👍 

Note: several files need a pretty print ;-)

sbcgua:
Hmmm. To rise the contrast: would you prefer developing AG as a set of objects or in a single file like standalone version ? Also look at real web projects as well as to most popular js libs approaches and I think you'll change your opinion.

mbtools:
Let's take the DB css as an example. It's all for styling a table... it should be a table component with default styles that work in the majority of cases (there might be data/column specific styles which are passed with the column spec). Why do we need extra css for that? The current css needs refactoring as stated in https://github.com/abapGit/abapGit/issues/3530 and imho all page-specific styles should be replaced with component-styles. 

I'm not sure what a "real web project" is but the AG UI is quite different. The UX is "simple" and that's what many like about it. Pages, toolbars, tables, forms. All but "tables" are in good shape. Standardizing "tables" via `zcl_abapgit_html_table` will allow to have a very clean implementation. I don't see where we need "extra css" or "extra js". There should be less and less of either (one example is "staging" where the JS is causing multiple issues).

sbcgua:
It's a very good example in fact.

1) Table. Absolutely agree. That's a good next step. But the main reason is that the table becomes a component. As well as the Form before. Now it's unified. And by the way nothing requires for table CSS styles to be in the common file. It may be assigned to the class like with the db page. (I didn't think if it is needed, but it is a possibility)
2) At the same time look at `.db-entry pre` or to some extent `.entry-tag`. They are DB page specific.

Problem is that designing a universal style set is quite a challenge. And from what I can see real project do not end in pure Bootstrap or Material design. There are always some exceptions. Look at Fiory. Look at SAP GUI ! (while being a very good UI library IMHO ! it's still limited and sometimes UI optimizations are quite desirable). So, yes, there should be commonly used components. But they won't cover 100%. Also most of modern JS-IN-CSS libraries assume that styles are close to components which use them - it's because people gravitate to that approach in practice. Common approach is to develop in separate dedicated files, then preprocess, bundle, treeshake, compress, whatever to optimize production objectives.

Developing of separate stylesheet is more convenient. As you're focused. As you should care less about name conflicts (by the way separate CSS mean less CSS styles at the same moment on a page). The testability of components is better. The convenience of access is solved by tooling. W3MIPOLLER is quite good already, but I have a couple of improvement ideas to simplify multifile development. Good tooling rules. This is the direction to improve things. E.g. I'd like to have CSS in SCSS notation somehow. Ideally if AG UI and backend are far enough, and if AG UI is localized it would make it possible to develop it on vscode or similar, using preprocessors and other modern stuff.

Or JS. A page uses maximum 20% of the code there. Yet navigation in that huge file is terrible. It's really like development of ag_standalone.

sbcgua:
Ideally, how I'd see the UI dev for AG if we want to keep both serverside rendering and have some code reuse for external UI. I'd like to have some kind of Vue.js approach. Where there is a component that consists of html template + JS + CSS as a view. And then a model filling that view. CSS and JS is kind of solved with this PR. html template - theoretically we can look into direction mustache templates (and [this](https://github.com/sbcgua/abap_mustache) - though it need a lot of improvement). Then the abap page becomes a model/controller. And the html template can be reused extenally (because mustache syntax is universal). But that's just an idea.

src/ui/lib/zcl_abapgit_html_table.clas.xml
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_CLAS" serializer_version="v1.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <VSEOCLASS>
+    <CLSNAME>ZCL_ABAPGIT_HTML_TABLE</CLSNAME>
+    <LANGU>E</LANGU>
+    <DESCRIPT>abapGit HTML table abstraction</DESCRIPT>
```
mbtools, line 8: abapGit - GUI HTML Table

src/ui/lib/zif_abapgit_gui_page_title.intf.xml
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_INTF" serializer_version="v1.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <VSEOINTERF>
+    <CLSNAME>ZIF_ABAPGIT_GUI_PAGE_TITLE</CLSNAME>
+    <LANGU>E</LANGU>
+    <DESCRIPT>abapGit page title provider</DESCRIPT>
```
mbtools, line 8: abapGit - GUI Page Title Provider

src/ui/lib/zif_abapgit_html_table.intf.xml
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_INTF" serializer_version="v1.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <VSEOINTERF>
+    <CLSNAME>ZIF_ABAPGIT_HTML_TABLE</CLSNAME>
+    <LANGU>E</LANGU>
+    <DESCRIPT>abapGit HTML table abstraction</DESCRIPT>
```
mbtools, line 8: abapGit - GUI HTML Table

src/ui/pages/zabapgit_css_page_db.w3mi.xml
```diff
@@ -0,0 +1,23 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_W3MI" serializer_version="v2.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <NAME>ZABAPGIT_CSS_PAGE_DB</NAME>
+   <TEXT>Page DB Styles (CSS)</TEXT>
```
mbtools, line 6: abapGit - DB Page Styles

src/ui/pages/zabapgit_css_page_db_entry.w3mi.xml
```diff
@@ -0,0 +1,23 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_W3MI" serializer_version="v2.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <NAME>ZABAPGIT_CSS_PAGE_DB_ENTRY</NAME>
+   <TEXT>abapGit - DB Enrty page CSS</TEXT>
```
mbtools, line 6: abapGit - DB Entry Page Styles

src/ui/pages/zcl_abapgit_gui_page_db_entry.clas.xml
```diff
@@ -3,7 +3,7 @@
  <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
   <asx:values>
    <VSEOCLASS>
-    <CLSNAME>ZCL_ABAPGIT_GUI_PAGE_DB_DIS</CLSNAME>
+    <CLSNAME>ZCL_ABAPGIT_GUI_PAGE_DB_ENTRY</CLSNAME>
     <LANGU>E</LANGU>
     <DESCRIPT>abapGit - GUI Database Display</DESCRIPT>
```
mbtools, line 8: abapGit - GUI Database Entry

src/utils/zcl_abapgit_string_buffer.clas.xml
```diff
@@ -3,9 +3,9 @@
  <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
   <asx:values>
    <VSEOCLASS>
-    <CLSNAME>ZCL_ABAPGIT_GUI_PAGE_DB_EDIT</CLSNAME>
+    <CLSNAME>ZCL_ABAPGIT_STRING_BUFFER</CLSNAME>
     <LANGU>E</LANGU>
-    <DESCRIPT>abapGit - GUI Database Edit</DESCRIPT>
+    <DESCRIPT>abapGit string buffer util</DESCRIPT>
```
mbtools, line 8: abapGit - String Buffer

mbtools (changes requested):
Works well. Just align the descriptions to `abapGit - ...` format

sbcgua:
Done and rebased

mbtools (approved):
👍

mbtools:
Pulling can result in an error since a method in `zcl_abapgit_asset_manager` was changed to an interface + alias. Standard SAP does not like such change when it scans the class/interface source. Similarly, you can get errors like this (which is from a different PR  since I didn't screenshot the issue this time):

image

I'm not sure if it's caused by the change to interface or by having an alias. Anyway, workaround is to pull twice (or three times).

mbtools:
needs rebase

sbcgua:
rebased.

Aliases: yes, I noticed this before as well. I think it makes sense to go through where used and adjust to use interface and the remove the alias. I didn't do it because the PR would become even larger.

mbtools:
tested again and works. merging.

sbcgua:
And somewhat improved dev flow for multiple files as promised ... (see video demo) and [w3mimepoller readme](https://github.com/sbcgua/abap_w3mi_poller)

https://user-images.githubusercontent.com/15635498/225389675-01d92ec0-f8bf-4d5d-900d-42e45f941b52.mp4

PR #6137, opened 2023-03-09
Pull dialog: Usability

mbtools:
Ref https://github.com/abapGit/abapGit/issues/6107

- Move "Continue" button to left  
  It was there before and matches layout for other AG forms.
- Add "Ctrl-S" shortcut for "Continue"  
  This allows using "Enter" to select/deselect rows and submitting popup without having to tab anywhere else

After:

image

### DIFF
--- a/src/persist/zcl_abapgit_persist_migrate.clas.abap
+++ b/src/persist/zcl_abapgit_persist_migrate.clas.abap
@@ -30,24 +30,24 @@ ENDCLASS.
 
 
 
-CLASS ZCL_ABAPGIT_PERSIST_MIGRATE IMPLEMENTATION.
+CLASS zcl_abapgit_persist_migrate IMPLEMENTATION.
 
 
   METHOD gui_status_create.
 
     DATA ls_cua TYPE zcl_abapgit_objects_program=>ty_cua.
 
-    IF gui_status_exists( ) = abap_true.
-      RETURN.
-    ENDIF.
-
     IF zcl_abapgit_factory=>get_environment( )->is_merged( ) = abap_false.
       RETURN. " No autocreation for full version
     ENDIF.
 
+    IF gui_status_exists( ) = abap_true.
+      RETURN.
+    ENDIF.
+
     ls_cua = lcl_own_cua_provider=>get( ).
 
-    IF ls_cua IS INITIAL. " Full version or Something wrong with abapmerged version
+    IF ls_cua IS INITIAL. " Full version or something wrong with abapmerged version
       RETURN.
     ENDIF.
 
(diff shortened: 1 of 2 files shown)


sbcgua:
> Move "Continue" button to left, It was there before and matches layout for other AG forms.

I object. This does not matter how it was before or how it's done elsewhere. It matter what's better. As I mentioned in my other messages:
- there is POV of a mouse user: for them it is better that most used buttons have fixed positions, independent of dynamic toolbar change (jumping) or further development improvements. Besides, OK and Cancel just group together naturally.
- then there is a Keyboard user POV (as Christian described, and I confess I missed that perspective initially). what he actually described is that after selecting the lines with keyboard, the next TAB-stop was the Continue button (before) and that was fast and convenient. And I totally agree. But my idea with Ctrl+S or F2 was that it is actually event faster and more universal and accessible from any point at the screen, including, when there is no need to select (and thus tab-through) all lines. Thus a dedicated shortcut is just superior to the button position for a keyboard user.

I'd suggest moving the button back.

Using the opportunity, I'd also suggest adjusting the keys a bit (as a proposal).
Move SEL_CAT to F8, so that all selection related buttons are in line and in the same sequence. Also I had an idea to change the logic of (De)Select All as they are a bit confusing now when there is a filter applied. But I can do it n a separate PR as well
image

mbtools:
I made this change based on my experience with this (changed) dialog. I use mouse and keyboard. 

It's significantly more difficult to find/click the button in the middle of a toolbar esp. since you use "ok" almost every time (in general and here too). It's also a shorter distance from the checkboxes. Consistency matters as well. Why should this dialog be different from all others? It's better with the button on the left. 

Ctrl-S was added for keyboard and it works very well. Where's the problem?

PS: I moved SEL_CAT to F8.

sbcgua:
I guess the problem is in different approach to product design. Whatever. It's a tool, it works

christianguenter2:
As a most of the time keyboard only user I now have several options to confirm the popup.
- CTRL+TAB, Enter/Space
- Enter when cursor is on any non hotspot column
- CTRL-S
As I like to have options I like this change especially as I can still use my old way to confirm this popup (first bullet point)

When I'm forced to use the mouse, e.g. while working on Citrix with different keyboard layout, this point is crucial to me:
> It's significantly more difficult to find/click the button in the middle of a toolbar esp. since you use "ok" almost every time (in general and here too)

So while understanding Alex arguments I vote for merging this PR. Maybe we can get a few more opinions.

sbcgua:
@christianguenter2 thanks for a practical and grounded opinion.
I think this is the reasoning the design should be based on. Of course more opinions would be valuable. Ideally an A/B test, though this is not that critical feature :)

PR #6144, opened 2023-03-12
LXE: more objects

sbcgua:
ref: #4415, #4470, #4494, #2539

Extending the work of #4415 to more objects to increase the testing surface for future improvements in this direction.
PARA, TRAN, MSAG, CLAS, TABL, VIEW, SHI3, DTEL, DOMA

Generally should be safe changes, but CLAS could benefit from some testing as the code change there was a bit less trivial. And also a review may be useful for `ZCL_ABAPGIT_OBJECTS->deserialize` - it was certainly a "make it work" piece of code.

`zif_abapgit_xml_input~i18n_params` has the same logic as his `output` counterpart.

Known issue (it was there before) - LXE is not applied to new objects from the first time. This is in plans. This PR is to add support for the feature to more objects only and to find patterns

### DIFF
--- a/abaplint.json
+++ b/abaplint.json
@@ -313,7 +313,7 @@
     "max_one_statement": true,
     "message_exists": false,
     "method_length": {
-      "statements": 100,
+      "statements": 105,
       "checkForms": true,
       "ignoreTestClasses": false,
       "errorWhenEmpty": false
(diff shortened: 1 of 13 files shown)


sbcgua:
@larshp can we raise number the max statement rule ? I'm not sure it makes sense to do any refactoring of those 2 methods in this PR. I can ... just that it will just add review complexity and potential bug spots.

larshp:
yea, just raise it a bit, eg. 105

mbtools:
I put it through CI (without the LXE setting). Still works. 👍 

image

image

image

We will need some new repos with [LXE](https://github.com/abapGit-tests?q=lxe&type=all&language=&sort=). Currenly only `PROG` and `FUGR`. Maybe a combined one with one object of each supported type with LXE.

mbtools (approved):
👍

PR #6169, opened 2023-03-24
Improve pull dialog button logic

sbcgua:
Back to this comment: https://github.com/abapGit/abapGit/pull/6093#issuecomment-1446787779

> Actually one usability concern ... intuitively, I'd expect that "Select All" (and also Deselect All) is applied to visible records ... Although it is not a hard rule and also a matter of habit, but may be a source of confusion ...

So after some practical hands on, my own, and my colleagues, I think the current logic is indeed imperfect. And the double nature of "halfpainted" button was confusing. Also the fact the "(de)select all" does something that is not visible on the screen. Here are some changes:

image

Button 1 - marks all VISIBLE, which means mark ALL in case of no filter
Button 2 - UNmarks all VISIBLE, which means UNmark ALL in case of no filter
Button 3 - mark SELECTED and resets the selection (the use case from Marc)

Feels better.

@mbtools @christianguenter2 What do you think ?

### DIFF
--- a/src/ui/zcl_abapgit_popups.clas.locals_imp.abap
+++ b/src/ui/zcl_abapgit_popups.clas.locals_imp.abap
@@ -95,6 +95,7 @@ CLASS lcl_object_descision_list DEFINITION FINAL.
     METHODS mark_indexed
       IMPORTING
         iv_selected TYPE abap_bool DEFAULT abap_true
+        iv_invert TYPE abap_bool DEFAULT abap_false
         it_scope TYPE lvc_t_fidx.
 
 ENDCLASS.
@@ -432,7 +433,9 @@ CLASS lcl_object_descision_list IMPLEMENTATION.
     lt_scope = mo_alv->get_selections( )->get_selected_rows( ).
 
     IF lines( lt_scope ) > 0.
-      mark_indexed( lt_scope ).
+      mark_indexed(
+        it_scope  = lt_scope
+        iv_invert = abap_true ).
       mo_alv->get_selections( )->set_selected_rows( lt_clear ).
     ELSE.
       MESSAGE 'Select rows first to mark them' TYPE 'S'.
@@ -490,7 +493,12 @@ CLASS lcl_object_descision_list IMPLEMENTATION.
 
       ASSIGN COMPONENT c_fieldname_selected OF STRUCTURE <ls_line> TO <lv_selected>.
       ASSERT sy-subrc = 0.
-      <lv_selected> = iv_selected.
+
+      IF iv_invert = abap_true.
+        <lv_selected> = boolc( <lv_selected> = abap_false ).
+      ELSE.
+        <lv_selected> = iv_selected.
+      ENDIF.
 
     ENDLOOP.
 
--- a/src/ui/zcl_abapgit_popups.clas.locals_imp.abap
+++ b/src/ui/zcl_abapgit_popups.clas.locals_imp.abap
@@ -1,11 +1,11 @@
 CLASS lcl_object_descision_list DEFINITION FINAL.
   PUBLIC SECTION.
 
-    CONSTANTS c_default_column TYPE abap_componentdescr-name VALUE 'DEFAULT_COLUMN'.
+    CONSTANTS c_default_column     TYPE abap_componentdescr-name VALUE 'DEFAULT_COLUMN'.
     CONSTANTS c_fieldname_selected TYPE abap_componentdescr-name VALUE 'SELECTED'.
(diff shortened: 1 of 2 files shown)


sbcgua:
@larshp Is it missing definition in deps ?
image

larshp:
set_selected_rows, yea, add it to https://github.com/abaplint/deps

mbtools:
definitely an improvement :-)

for button 3, how about changing it to a toggle? reverse the checkbox of selected rows.

sbcgua:
> set_selected_rows, yea, add it to https://github.com/abaplint/deps

But where exactly ? https://github.com/abaplint/deps/tree/main/src/alv ?
Just add a class definition ? Do you create XMLs manually ?

sbcgua:
> for button 3, how about changing it to a toggle? reverse the checkbox of selected rows.

I was a bit tentative about this. How to understand that users wants to DEselect ? check the first selection line ? Or maybe if ALL in the selection are marked then UNmark, otherwise REmark all (including those without mark yet) ?

mbtools:
loop over all selected: if it's unmarked, mark it. if it's marked, unmark it. 

description of button: "Toggle selected rows"

sbcgua:
> loop over all selected: if it's unmarked, mark it. if it's marked, unmark it.
> description of button: "Toggle selected rows"

Hmmm, are you going to use it like that ? (I just don't use this 3rd button ... so I'm trying to understand the use case :)

sbcgua:
Hmmm, for me it is contra intuitive ... especially since AG marks some (new) items automatically ... I'd suggest unmark only if all selected were marked ... but we need some opinions, I'm neutral because I'm not using this button.


https://user-images.githubusercontent.com/15635498/227607027-8e1840cc-6b94-40ab-9e4d-6b0d523e2c6a.mp4

sbcgua:
Try the last version - I don't strongly insist but it really makes sense. Also because it is possible to select rows not unpressing the mouse so just drag over. So it is easier to select range. Inversion ... I just don't see the real use case.

> it will be clearer if you keep the rows selected.

Yes ... to some extent ... but keeping them selected has another issue - how to deselect all ?

And by the way

https://user-images.githubusercontent.com/15635498/227610935-fbc8040e-4334-4e11-8ebb-37a486827e62.mp4

mbtools:
Dinner break 👨‍🍳

src/zabapgit.prog.xml
```diff
@@ -101,15 +101,15 @@
...
+      <FUN_TEXT>Deselect All Visible</FUN_TEXT>
      </RSMPE_FUNT>
      <RSMPE_FUNT>
       <CODE>SEL_KEY</CODE>
       <TEXTNO>001</TEXTNO>
       <TEXT_TYPE>S</TEXT_TYPE>
       <TEXT_NAME>ICON_SELECT_BLOCK</TEXT_NAME>
       <ICON_ID>@4C@</ICON_ID>
-      <FUN_TEXT>Mark selected or all visible</FUN_TEXT>
+      <FUN_TEXT>Mark/Toggle selected</FUN_TEXT>
```
mbtools, line 112: selected > Selected

mbtools (approved):
Tested and LGTM

I'm not sure why workflow tries to transpile in this class

src/zabapgit.prog.xml
sbcgua: Done

src/ui/zcl_abapgit_popups.clas.locals_imp.abap
```diff
@@ -88,7 +88,20 @@ CLASS lcl_object_descision_list DEFINITION FINAL.
...
+    METHODS mark_indexed
+      IMPORTING
+        iv_selected TYPE abap_bool DEFAULT abap_true
+        iv_invert TYPE abap_bool DEFAULT abap_false
+        it_scope TYPE lvc_t_fidx.
+    METHODS are_all_marked
+      IMPORTING
+        it_scope TYPE lvc_t_fidx
+      RETURNING
+        VALUE(rv_yes) TYPE abap_bool.
```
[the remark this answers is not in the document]
sbcgua: Done (if you meant types alignment)

sbcgua:
Not sure what to do with tests, I'd say we need some description of what to do in such case a and where to look, it's very confusing and unclear :( There is no mistake for get_rows but there is for set, yet the class (alv_selection) is not in the deps at all ... and ... very confusing.

larshp:
green, fixed, https://github.com/open-abap/open-abap-core/pull/599

mbtools (approved):
👍

PR #6183, opened 2023-04-01
fix msag i18n serilization

sbcgua:
a bugfix to #6168

MSAG does not use any FM to extract the translation and uses direct select. That select didn't exclude master lang. Just a bug.


sbcgua:
actually merging, cause it really just a bug fix

PR #6186, opened 2023-04-01
CLAS: Serialize long texts for exception classes

mbtools:
Test repo: https://github.com/abapGit-tests/CLAS_with_SOTR

Serializes the exception longtext (`SOTS`) as part of the class `XML`.

Ref #4480


larshp (changes requested):
please wait with merges to main until after todays call

PR #6198, opened 2023-04-05
CLAS: Refactor skip of SADL classes

mbtools:
Move skip logic from `zcl_abapgit_tadir` to `zcl_abapgit_object_clas`

Ref https://github.com/abapGit/abapGit/issues/4187


larshp (approved):
looks like its more deletions than moving, which is good

PR #6229, opened 2023-04-17
Enhance Serializer Interface

mbtools:
As proposed during community call and in https://github.com/abapGit/abapGit/issues/4187, here are the enhancements of the serialized interface.

https://larshp.github.io/abapGitCC20230404/#/10

New Methods

- get_deserialize_order 
- map_object_to_filename
- map_filename_to_object

New Parameters

- changed_by
- jump

Note: The three new methods are not called yet. This will happen in a subsequent PR.

### DIFF
--- a/src/objects/zif_abapgit_object.intf.abap
+++ b/src/objects/zif_abapgit_object.intf.abap
@@ -90,16 +90,20 @@ INTERFACE zif_abapgit_object
 
   CLASS-METHODS map_filename_to_object
     IMPORTING
-      !iv_filename   TYPE string
-      !iv_path       TYPE string OPTIONAL
-      !io_dot        TYPE REF TO zcl_abapgit_dot_abapgit OPTIONAL
-    RETURNING
-      VALUE(rs_item) TYPE zif_abapgit_definitions=>ty_item
+      !iv_filename TYPE string
+      !iv_path     TYPE string OPTIONAL
+      !io_dot      TYPE REF TO zcl_abapgit_dot_abapgit OPTIONAL
+    CHANGING
+      cs_item      TYPE zif_abapgit_definitions=>ty_item
     RAISING
       zcx_abapgit_exception.
 
-  METHODS map_object_to_filename
-    RETURNING
-      VALUE(rv_filename) TYPE string.
+  CLASS-METHODS map_object_to_filename
+    IMPORTING
+      !is_item    TYPE zif_abapgit_definitions=>ty_item
+    CHANGING
+      cv_filename TYPE string
+    RAISING
+      zcx_abapgit_exception.
 
 ENDINTERFACE.


src/objects/zif_abapgit_object.intf.abap
```diff
@@ -25,46 +26,80 @@ INTERFACE zif_abapgit_object
...
     RETURNING
       VALUE(rt_steps) TYPE zif_abapgit_definitions=>ty_deserialization_step_tt .
+
+  METHODS get_deserialize_order
+    IMPORTING
+      !it_all_objects          TYPE zif_abapgit_definitions=>ty_items_tt
+    RETURNING
+      VALUE(rt_objects_before) TYPE zif_abapgit_definitions=>ty_items_tt.
+
+  METHODS map_filename_to_object
```
larshp, line 91: wondering, should this method be static? it takes a filename to get the object, but it requires the object to call the constructor?

mbtools: Technically it should be but then the method can't be part of the interface.

larshp: below works, its strange to have static methods part of an interface, but I think its possible in ABAP

and either way, how to call the method? supply dummy values to the instantiation/constructor?

```abap
INTERFACE lif.
  CLASS-METHODS foo.
ENDINTERFACE.
```

mbtools: Oh yes. Forgot about that. I will change it

larshp (approved):

larshp:
need https://docs.abapgit.org/development/serializers.html updated, but doesnt have to be now

mbtools:
I changed the other mapping method to `static` and parameters to `changing`. Makes it clearer that AG will pass in the default mapping and an object type can change it if necessary. 

```abap
  CLASS-METHODS map_filename_to_object
    IMPORTING
      !iv_filename TYPE string
      !iv_path     TYPE string OPTIONAL
      !io_dot      TYPE REF TO zcl_abapgit_dot_abapgit OPTIONAL
    CHANGING
      cs_item      TYPE zif_abapgit_definitions=>ty_item
    RAISING
      zcx_abapgit_exception.

  CLASS-METHODS map_object_to_filename
    IMPORTING
      !is_item    TYPE zif_abapgit_definitions=>ty_item
    CHANGING
      cv_filename TYPE string
    RAISING
      zcx_abapgit_exception.
```

andrecarrilho:
Hello,

I've been trying to use the method map_filename_to object in some new serializer classes I've implemented for some "custom objects" but noticed that the call for this is not yet implemented.
Is there any timeframe for when this will be done?
Or how can I help implement/test this feature?

Thank you!

mbtools:
@andrecarrilho the new methods are called already in the latest version. You can see an example in `zcl_abapgit_object_devc`. 

PS: I merged the doc changes.

andrecarrilho:
hi @mbtools,

Sorry to bother you with this but I've set a breakpoint in the method of the class that you provided as an example, but it never stops there, and I've tried with a few different repos.
I saw that this method is called dynamically inside of ZCL_ABAPGIT_FILENAME_LOGIC->MAP_FILENAME_TO_OBJECT, which is called inside of method FILE_TO_OBJECT of the same class.
I've set a breakpoint in there as well and again it never stops.
Also tried setting up a breakpoint for method, (both MAP_FILENAME_TO_OBJECT and FILE_TO_OBJECT ) and got the same result.

Am I looking at this the wrong way or am I missing something? 
Can you point me in the correct direction? 

Thank you for the help!

mbtools:
are you using the dev version? if yes, then you might have to disable parallel processing 
https://docs.abapgit.org/guide-settings-personal.html#developer-version

larshp:
called from 
image

andrecarrilho:
Hi @larshp  and @mbtools ,

@larshp , that is precisely the method I'm trying to stop at.
@mbtools , I have disabled the parallel processing, and still nothing.

I set up the breakpoint at the method like this.
image
image

I noticed that MAP_FILENAME_TO_OBJECT is called inside of the FILE_TO_OBJECT method, then I checked where this method is called.
Looking at that, I set up break points in all the places where ZCL_ABAPGIT_FILENAME_LOGIC->FILE_TO_OBJECT would be called, and I saw that the conditions are never met so that the method would be executed.
Looking in particular to ZCL_ABAPGIT_SERIALIZE->ADD_DATA , I see that the internal table LT_FILES is never filled:
image

Just to double-check all of this I asked another colleague to try and stop at the MAP_FILENAME_TO_OBJECT method, just to make sure I'm not forgetting something, and he also could not make it stop there.

Am I still missing something? 
Just trying to help here, but to me, it looks like the conditions to execute the method are never met.

Let me know if I can provide any more information.

Kind regards,
André

mbtools:
The method is called only for files that exist in the remote repo but not locally. The place you mentioned (`add_data`) is only relevant, if the repo contains data selection.

For example, create a repo for https://github.com/abapGit-tests/DOMA.

image

AG will stop at break-point. Here's the callstack:

image

andrecarrilho:
Hi @mbtools ,

Now it makes sense!
Thank you for the explanation.

Kind regards!

PR #6233, opened 2023-04-20
Enhance Serializer Interface - Follow-up

mbtools:
Adjust the order of method implementations to avoid diffs when changing serializer classes later


mbtools:
Made a little tool to automatically adjust these classes:
https://gist.github.com/mbtools/a0a1ca9391ee69d1b0bc333b590ed7ea

Issue #2722, opened 2019-05-31
ECTC/ECATT: abapGit CI failed - diff


christianguenter2:
![image](https://user-images.githubusercontent.com/17437789/58690738-8add5f80-838a-11e9-826c-af1dd881cb83.png)

Demo repo
https://github.com/abapGit-tests/ECATT.git

christianguenter2:
Error occurs only if abapGit CI runs in background. Then local xml diffs.

![image](https://user-images.githubusercontent.com/17437789/59632220-fe80ba00-9149-11e9-983e-fede089de13c.png)

![image](https://user-images.githubusercontent.com/17437789/59632219-fcb6f680-9149-11e9-93e0-a77213e80b95.png)

mbtools:
Works with 754 SP 7 and latest abapGit

PR #6240, opened 2023-04-23
ECATT: Avoid diffs between dialog and background

mbtools:
Ref https://github.com/abapGit/abapGit/issues/2722

image

Fixes diff due to user-specific settings that are loaded only in dialog processes. The corresponding XML field is now cleared in a way that avoids diffs.


larshp (approved):
looks good to me

PR #6242, opened 2023-04-24
Separate translations from XML, PO files [one file approach], WIP

sbcgua:
Following the suggestion from @mbtools here is an experiment to have one PO file per repo

- related to #6189
- concept considerations (pros and cons) described [here](https://github.com/abapGit/abapGit/issues/2539#issuecomment-1495537247) in **Update 2023-04-24** section
- compared to "many files" in #6189 this PR is really a WIP:
  - built on top of the first PR, thus the code was not optimized, needs more work if we choose to go this way
  - deserialization does not work, because of ... suboptimal architecture in decision logic (in `ZCL_ABAPGIT_SERVICES_REPO=>GUI_DESERIALIZE`)
- Technical-wise
  - [diff between approaches](https://github.com/sbcgua/abapGit/compare/po...sbcgua:abapGit:po-onefile?expand=1)
  - objects relevant to translation are collected during serialization loop into `zcl_abapgit_i18n_manager`
  - which then builds single po file per language and places them to `/translations/` dir

FYI @schneidermic0


sbcgua:
closing this one

PR #6260, opened 2023-05-05
Add warning banner if wrong browser control is used

### LINKED ISSUE #6257: Add warning at abapGit startup when edge browser control is used
Would ease root cause analysis for problems like #6253
https://docs.abapgit.org/guide-sapgui.html#sap-gui-for-windows

I already tried `cl_gui_frontend_services=>get_gui_properties` but it doesn't reveal that information. Maybe there's another way.

https://github.com/christianguenter2/abapGit/tree/debug-sapgui-properties

christianguenter2:
fixes #6257 
grafik

### DIFF
--- a/src/ui/pages/zcl_abapgit_gui_page.clas.abap
+++ b/src/ui/pages/zcl_abapgit_gui_page.clas.abap
@@ -296,17 +296,20 @@ CLASS zcl_abapgit_gui_page IMPLEMENTATION.
 
   METHOD render_browser_control_warning.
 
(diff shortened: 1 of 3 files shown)


mbtools (approved):
It's a bit harsh since the warning doesn't go away. But it's necessary. Hopefully, it will be a temporary thing.

I'm enhancing the footer to show GUI/HTML details which will be helpful also long-term.

fabianlupa:
It's also quite harsh since it overlays parts of the repo view :(

image

PR #6268, opened 2023-05-10
Fix dump related to exit implementations

### LINKED ISSUE #6265: Dump running the standalone version in a system with full version and exit
image

Supposedly standalone tries to create global class which is based on global interface and assign it to local interface

mbtools:
Fixes dump in standalone version if exit class for developer version exists (`zcl_abapgit_user_exit`) but not the exit include for standalone version (`zabapgit_user_exit`).

Closes #6265


larshp (approved):

sbcgua:
Looks good, thanks ! :+1:

PR #6273, opened 2023-05-11
wip, Generic modal popups

mbtools:
This took a while to figure out but here are modal popups in abapGit 😀 

Might need a bit of CSS fine-tuning but works like a charm otherwise.

saplogon_1396

How it works:

- Page contains a reference to popup object (`mo_popup`)
- If popup object is defined, then it's rendered as a modal overlay
- Events from the popup form are then processed by the page again (`on_popup_event`)

To implement a new popup type, create a class inheriting from `zcl_abapgit_gui_popup` and define the form (for example, `zcl_abapgit_gui_popup_picklist`). You probably want to set the popup dimensions and maybe add validations. I don't think it could be simpler.

### DIFF
--- a/src/ui/zabapgit_css_common.w3mi.data.css
+++ b/src/ui/zabapgit_css_common.w3mi.data.css
@@ -1502,4 +1502,57 @@ table.unit_tests {
 /* Warning if wrong browser control is used */
 .browser-control-warning {
   width: 100%;
-}
+}
+
+/* MODAL POPUP */
+/* https://css-tricks.com/considerations-styling-modal/ */
+
+.modal {
+    /* center on screen */
+    position: fixed;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    /* size */
+    width: 680px;
+    max-width: 100%;
+    height: 260px;
+    max-height: 100%;
+    /* infront of overlay */
+    z-index: 1010;
+    /* show by default */
+    display: block;
+    /* style */
+    background: lightgray;
+    border-radius: 6px;
+}
+
+.modal-guts {
+    /* cover the modal */
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    /* spacing as needed */
+    padding: 5px 0px;
+    /* let it scroll */
+    overflow: auto;
+}
+
+.modal-guts .dialog {
+    margin-top: inherit;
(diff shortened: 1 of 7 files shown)


sbcgua:
Looks interesting, Will have a look a bit later today.

ref: #6266

src/ui/zabapgit_css_common.w3mi.data.css
```diff
@@ -1502,4 +1502,57 @@ table.unit_tests {
...
+
+/* MODAL POPUP */
+/* https://css-tricks.com/considerations-styling-modal/ */
+
+.modal {
+    /* center on screen */
+    position: fixed;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
```
sbcgua, line 1515: is it supported by IE control?

sbcgua:
Nice approach. :+1:

- What I like: visual aspect, certainly better than a separate page if the dialog is small
- Arguable: too much popup code in the caller class (event handling), but I think it can be moved
- Doubts: requires rerender of the underlying page, some of which are heavy!
- Doubts 2: hotkeys should be somehow disabled upon the popup
- Also: I think we should avoid inheritance, but rather implement it in a composable way (interfaces)

This one vs #6266 - I actually think we can have both. And choose the proper one per application. This one fits well for the config forms and lightweight pages. Maybe the password dialog? The whole-page one can be used for pull dialog - which is supposed to be large and functional and should avoid re-rendering the repo page underneath.

mbtools:
Except for the diff page, I've not see long render times but it's certainly something to keep an eye on. I agree, bigger dialogs like pull should be pages. 

I'm just not good at "composition" so I used inheritance ;-). Actually, the PR popup should be its own class composed of the picklist popup. Then "list_pull_requests" and "get_pull_request" move to that class. The key of the approach here is that there's only one popup object in the page (`mo_popup`).

src/ui/zabapgit_css_common.w3mi.data.css
mbtools: yes, I only tested on IE

sbcgua:
Sure, make it work, then make it proper. Those are solvable.

> Maybe you can convert this approach to "composition" :-)

Do you mind if I commit to this branch?

I also wonder if the final class can support both in-page and separate-page approaches ... could be nice.

mbtools:
closing in favor of #6277

Issue #5064, opened 2021-10-27
SICF: suggest to change serialized xml file name from obj_name to URL of icf node


boy0korea:
The problem is obj_name of SICF is generated when length of node name is longer than 15.

Test Case:
1. Create WebDynproABAP application (length > 15)
2. Commit
3. Delete application and create same name again
4. Commit

Test Result: 
- 2 files created despite of same URL

Sugestion:
Keep longname info in internal table.
See below types.
> 
     TYPES:
      BEGIN OF ty_item_longname,
        obj_name_fake TYPE tadir-obj_name,
        obj_type_orig TYPE tadir-object,
        obj_name_orig TYPE tadir-obj_name,
        longname      TYPE string,
      END OF ty_item_longname .
    TYPES:
      ty_item_longname_th TYPE HASHED TABLE OF ty_item_longname WITH UNIQUE KEY obj_name_fake .

`longname` is for external name (filename, item list display) and `obj_name_fake` is for internal object name (in program).
`obj_name_fake` is hash value of `longname`.
longname of SICF is URL.

for example:
- current version
![image](https://user-images.githubusercontent.com/75079431/138997926-f0a315df-7de4-4bb0-9b63-ece60c414956.png)
- developing for longname
![image](https://user-images.githubusercontent.com/75079431/138990818-13c1d100-f3fb-4782-9496-7458efdfdc5d.png)

mbtools:
Hi @boy0korea 

sometimes things take a while, but I finally refactored the serializer interface and made changes to `zcl_abapgit_object_sicf` to address the issue with services generated for WD applications with names longer than 15 characters. 

The final piece is included in https://github.com/abapGit/abapGit/pull/6275.

Best,
Marc

PR #6275, opened 2023-05-11
SICF: Remove special TADIR logic

### LINKED ISSUE #6172: SICF: Issues due to special TADIR logic
This is a collection of issues arising from the fact that `SICF` has a special logic for mapping TADIR entries to items and therefore filenames:

- #5064 (still relevant)
- #4187 
- Unable to determine transport (see below)

image

image

image

Proposed solution:

1. Remove special sicf logic from `zcl_abapgit_tadir`. 
2. Move the special logic to `zcl_abapgit_filename_logic`. 
3. Implement the special item-filename mapping using new methods in `zif_abapgit_object` (or an additional interface) 

Result:

The AG item will contain the same data as the tadir entry (no more exceptions).  The new logic can then be used by other object types which contain system-specific data in the tadir object name (for example `BMFR` and `SOTR/SOTS` https://github.com/abapGit/abapGit/issues/4480).

mbtools:
Leveraging the new serializer interface. The change removes the special logic translating TADIR into ITEM for `SICF` and replaces it with an equivalent mapping between object and filenames. 

The implementation of the serializer class is now much more straight forward. It's always using `ms_item` to refer to the object which is equivalent to the tadir entry like for other objects. 

Ref #3453, #4187, #5064 
Closes #6172

PS: Breaks `zcl_abapgit_ci_repo` (https://github.com/abapGit/CI/pull/228)

### DIFF
--- a/src/objects/zcl_abapgit_object_sicf.clas.abap
+++ b/src/objects/zcl_abapgit_object_sicf.clas.abap
@@ -8,10 +8,10 @@ CLASS zcl_abapgit_object_sicf DEFINITION
 
     INTERFACES zif_abapgit_object .
 
-    TYPES: ty_hash TYPE c LENGTH 25.
-
   PROTECTED SECTION.
   PRIVATE SECTION.
+
+    TYPES ty_hash TYPE c LENGTH 25.
 
     TYPES:
       ty_icfhandler_tt TYPE STANDARD TABLE OF icfhandler WITH DEFAULT KEY .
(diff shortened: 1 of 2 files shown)


larshp:
tested ok, no diffs

image

larshp (approved):

src/objects/zcl_abapgit_object_sicf.clas.abap
```diff
@@ -10,21 +10,6 @@ CLASS zcl_abapgit_object_sicf DEFINITION
 
     TYPES: ty_hash TYPE c LENGTH 25.
```
larshp, line 11: can this TYPE be moved to private?

mbtools: done

PR #6277, opened 2023-05-12
Modals unification, in-page, own page, better abstraction, on example of settings_remote page

sbcgua:
ref: #6266, #6273

Quickly: 
- branch, tag and pull_requests are now on new popups
- pull_requests intentionally left on a separate page to serve as a POC - will move to in-page popup after comments/discussions
- proper UI design is intentionally out of this PR 

TODO: describe details

Screenshots:

image

image


sbcgua:
ready for testing and review. See the PR description. I will also add more technical details later.
@larshp @mbtools

sbcgua:
1) ... Not sure I understood the comment :)
2) Yes, that's the plan. They are intentionally abstract enough so if the direction looks good I'll move them yet in this PR
3) Will have a look - that was some disappearing bug so thank for the sequence
4) ... mmm, I just wanted to avoid designing in this PR cause I will then end up rewriting the whole component (I'm sure it must not be the `form` in the background) ... and I'm a bit tired of this piece of code already :) I'll see, maybe will make some CSS border over the options ...

mbtools:
Some details

1. The new CSS overwrites max-width:
 
Before:

image

After:

image

4. `zcl_abapgit_gui_picklist->get_form_schema`

```abap
    ro_form->radio(
      iv_name     = c_radio_name
      iv_condense = abap_true     "<<< insert
      iv_label    = mv_title ).
```

After, it's nicer for branches (few entries):

image

but for tags it becomes too long:

image

sbcgua:
- fixed 2 and 4 (though arguable ... maybe will try with border yet ...)
- added a graceful back feature - allows A) exiting popups with Esc (hmm ... maybe also applicable for the hotkeys ?) B) graceful shutdown of a page (if ever needed)
- 3 - didn't check yet
- 1 - I don't understand. But the modal is a completely new CSS! How can it impact the existing layouts?

src/ui/lib/zif_abapgit_html_popup.intf.xml
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_INTF" serializer_version="v1.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <VSEOINTERF>
+    <CLSNAME>ZIF_ABAPGIT_HTML_POPUP</CLSNAME>
+    <LANGU>E</LANGU>
+    <DESCRIPT>abapGit HTML popup interface</DESCRIPT>
```
mbtools, line 8: abapGit - HTML Popup Interface

mbtools:
looks good. i will check the css issue

src/ui/lib/zcl_abapgit_html_popups.clas.xml
```diff
@@ -0,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<abapGit version="v1.0.0" serializer="LCL_OBJECT_CLAS" serializer_version="v1.0.0">
+ <asx:abap xmlns:asx="http://www.sap.com/abapxml" version="1.0">
+  <asx:values>
+   <VSEOCLASS>
+    <CLSNAME>ZCL_ABAPGIT_HTML_POPUPS</CLSNAME>
+    <LANGU>E</LANGU>
+    <DESCRIPT>abapGit HTML popul library</DESCRIPT>
```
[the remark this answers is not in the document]
sbcgua: Done

src/ui/lib/zif_abapgit_html_popup.intf.xml
sbcgua: Done

mbtools:
that works

1. `<body class="centered">` was changed to `<body class="">`

sbcgua:
> 1. `<body class="centered">` was changed to `<body class="">`

In this PR ?

sbcgua:
3 (pages "stuck", disabled events) - this one is a bit tricky ... I see an easy path via the graceful back but it's not as universal as I'd like to ... hmmm

sbcgua:
PRs

image

sbcgua:
OK, now I think it's not bad :)

There is potential room for unifications for popup handling in `render` and `on_event`, but I wouldn't do it now - needs more use-cases to observe.

ready for testing, review and test with the transpiled version ;)

Next steps:
- rewrite picklist internals - it should be based on a `form`. In particular, no clear way to specify a value selected-by-default. And no room for picklist-specific features (maybe extra buttons ? honestly, I didn't think what are those features, but I feel there are some)

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
```diff
@@ -984,19 +1017,35 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_SETT_REMO IMPLEMENTATION.
...
+
     " If staying on form, initialize it with current settings
-    IF rs_handled-state = zcl_abapgit_gui=>c_event_state-re_render.
-      mo_form = get_form_schema( is_settings  = ms_settings_old
-                                 io_form_data = mo_form_data ).
-      CREATE OBJECT mo_form_util
-        EXPORTING
-          io_form = mo_form.
+    IF rs_handled-state = zcl_abapgit_gui=>c_event_state-re_render AND mo_popup_picklist IS NOT BOUND.
+      " TODO refactor: same as in constructor
```
sbcgua, line 1037: I'm not quite sure what this code (in this IF block) is doing ... looks like a part of the construction.

sbcgua:
on "Change ELSE part to ELSEIF" check. I'm not insisting and will fix it if requested, but ...

I really think this rule causes more harm than benefits. This is a good example. What we have:

```
if popup is created
  do this
else (if popup has returned)
  do that
```

what rules suggests:

```
if popup is created
  do this
elseif popup->is_not_cancelled
  do that
```

Although the latter "optimize" the code by 1 line it makes it less readable. In the first case, there is a clear distinction between "what happened" modes. It is also easy to refactor it (if ever needed) and split it into 2 methods. While `elseif` need a bit more effort. Yet my main claim is about readability. In the latter case, I need to additionally think that `elseif something` also means the "return" mode of the popup. Or is it only so for me?

@larshp

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
```diff
@@ -200,27 +211,38 @@ CLASS ZCL_ABAPGIT_GUI_PAGE_SETT_REMO IMPLEMENTATION.
...
+    DATA ls_branch      TYPE zif_abapgit_git_definitions=>ty_git_branch.
 
-    IF mo_form_data->get( c_id-offline ) = abap_true.
-      RETURN.
-    ENDIF.
+    IF iv_is_return = abap_false.
 
-    lv_url = mo_form_data->get( c_id-url ).
-    lv_branch_name = zif_abapgit_definitions=>c_git_branch-heads_prefix && mo_form_data->get( c_id-branch ).
+      IF mo_form_data->get( c_id-offline ) = abap_true.
```
larshp, line 220: this is a 

```abap
IF sdf.
  ...
ELSE.
  IF ffds.
     ...
  ENDIF.
ENDIF.
```

it can be changed to

```abap
IF sdf.
 ...
ELSEIF ffds.
  ...
ENDIF.
```

sbcgua: just posted a separate comment on that above. Could you pls have a look? But will fix if you insist, no prob

larshp: is your agument that
```abap
ELSE.
  IF something.
  ENDIF.
ENDIF.
```
is easier to understand/less effort than following?
```abap
ELSEIF something.
ENDIF.
```

sbcgua: essentially yes. If the upper `if` has a clear binary nature. IMHO :)

mbtools: For me it depends. 

```abap
IF color = red. 
ELSEIF color = blue.
ELSE.
  IF shape = circle.
  ELSEIF shape = square. 
  ELSE.
  ENDIF. 
ENDIF. 
```

This is easier to follow than the alternative (one if) because I can better understand what "else" means (ie other colors) before checking shapes. 

In general the rule makes sense but there are exceptions where separating conditions into statements can make it better. 🤷‍♀️

larshp: I guess that code would be better with a CASE

anyhow, for binary

IF binary.
```abap
IF binary.
ELSE.
  IF something.
  ENDIF.
ENDIF.
```

perhaps do

```abap
CASE binary.
WHEN true.
WHEN false.
  IF something.
  ENDIF.
ENDCASE.
```

?

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
mbtools: That block came from https://github.com/abapGit/abapGit/pull/5483

Usually, the schema is defined only in the constructor and never changed. But here, after form values change - like switching online/offline, or changing the head type -, the form schema needs to be updated.

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
larshp: yea, well, that also fixes it, at least for now(hope to get some more rules done sometime)

larshp (approved):
pulled ok, tested ok

@mbtools any comments from you?

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
sbcgua: I decided not to block the PR so fixed it this way. The question with rules is more a discussion. But i'd keep it somewhere:) maybe an issue?

Maybe my "binary" comment was too abstract. Marc's comment actually shows a similar situation better. **The readability decreases if contexts are mixed**. In Marc's example it is color and shape. 2 independent contexts. In my case it is "initation- return" and "examining the return results". So for discussion-sake I still believe this linter rule makes harm.

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
sbcgua: Aha. Thanks. This explains indeed. I started some refactoring to fix it. But it appears to be larger than I expected. So in a separate pr right after this one. 

Maybe will experiment with layout definition too. It's full of IFs. I think it can be improved.

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
larshp: for Marcs case I think using CASE would fit, triggering https://rules.abaplint.org/change_if_to_case/

just `IF binary ELSE` doesnt trigger that rule

but CASE clearly communicates 2 different scenarios to the reader IMHO

mbtools:
The issue with missing class in `<body>` remains.

The new constructor in `zcl_abapgit_gui_page_hoc` overwrites the default for `ms_control-page_layout = 'centered'` (set in `zcl_abapgit_gui_page`) with `initial`.

mbtools (approved):

sbcgua:
🎉 🎉 🎉 
Nice feature. A piece of good collaboration from my perspective :)
Lars, thanks for cold-starting the hard stuff (as usual), Marc, thanks for the cool idea and POC with in-page popups.

On my side:
- some docs on popups
- some more refactoring of the sett_remo (started but not finished. was very busy with personal matters this weekend)

sbcgua:
Docs draft: https://github.com/abapGit/docs.abapgit.org/pull/137

PR #6283, opened 2023-05-15
Add fallback for ADT transport link

### LINKED ISSUE #6278: Transport jump doesn't fallback to SAP GUI anymore
[3df7b9e58a3d3826eaba21c192129bba7258a235](https://github.com/abapGit/abapGit/commit/3df7b9e58a3d3826eaba21c192129bba7258a235#diff-d92cd94466f347800b0ebdf5aad50483a10512ee5042cc9599c1df150c6602d5L479) seems to have broken my workflow of pressing "Deny" on the popup to open ADT so that the link is opened in SAP GUI.

`cl_gui_frontend_services=>execute` returns `error_execute_failed` in that case and the fallback solution of using `TR_DISPLAY_REQUEST` is no longer part of the exception handling.

mbtools:
Closes #6278


fabianlupa (approved):
It works

PR #6295, opened 2023-05-21
Persist settings of repository list

mbtools:
Saves the state of filter, favorites, details, and order of repository list per user


sbcgua (approved):
Work for me, nice usability feature.

P.S. maybe it makes sense to add a Reset command to instantly get back to defaults. Probably in a separate PR. E.g. collect the Only Favs, Detail and Reset into a drop listnear the filter (as it would become too crowded), also it will allow longer and clearer name for commands (Detail is confusing for example, just "Reset" will also be)

PR #6300, opened 2023-05-22
Refactoring in settings-remote (WIP)

sbcgua:
- originated by: https://github.com/abapGit/abapGit/pull/6277#pullrequestreview-1433304809
- rework form initialization
- ~~maybe experiment with IF-less multi-layout forms (descriptive definition)~~ - **NO**
- also don't like the constellation of form/form_data/form_utils/validation_data - maybe room to improve and consolidate. maybe. **UPD**: not this time. Also if form knows about utils this would increase coupling.
- maybe move commit popup to HTML (to complete this page). **UPD** maybe not in this PR (but I'll try yet), actually commit picklist is a good and simple model for table-like selector, so nice next target

**UPD**
The refactorings I couldn't pass by:
- string_map can merge with another one - useful sometimes + as a solution for `hidden` field skipping @mbtools have a look - IMHO a cleaner code for the form data handling (although I still think the hidden field should not be skipped)
- form utils can check dirty state without internal attribute setting
- html can render a renderable (a bit on edge but I think it makes sense, especially since we go in smaller ui component direction)
- small improvement in html debug
- small improvement in gui graceful back handling


sbcgua:
@mbtools Marc, do you remember what was the reason for changes in #6139, namely skipping hidden fields when normalizing form data? hidden field can be used to pass some information implicitly. And it is done on the sett_remo page. So I think that the hidden fields should be allowed back. I have another potential solution for my issue in this refactoring, but filtering hidden fields look wrong anyway. What do you think ?

mbtools:
Hidden fields caused issue with `is_dirty` to check if changes need to be saved. Ideally, this logic would be handled by the form (without using form utils).

sbcgua:
Ugh :)

image

One small refactoring ... :)

By the way, I wonder why not make `zcx_abapgit_exception` a `no_check` exception?
Just for discussion. I'll fix this issue here.
But in general. `cx_static` vs `no_check`. Any strong opinions on that? I gravitate to say that static exception definition does not bring a lot of value. In fact, abaplint may be able to catch no_check chains ... if so it would simplify the development just a bit while keeping the code robustness.
@larshp @mbtools

sbcgua:
actually ready for review, I added some updates to the main description.

I'll fix the linter issues after - the fixes will pollute the changes a bit ...

src/ui/core/zcl_abapgit_html.clas.abap
```diff
@@ -381,7 +390,12 @@ CLASS zcl_abapgit_html IMPLEMENTATION.
         TRY.
             lo_html ?= ig_chunk.
           CATCH cx_sy_move_cast_error.
-            ASSERT 1 = 0. " Dev mistake
+            TRY.
+                li_renderable ?= ig_chunk.
+                lo_html ?= li_renderable->render( ).
```
mbtools, line 395: how about this instead of passing on the exception:

```abap
try.
    li_renderable ?= ig_chunk.
    lo_html ?= li_renderable->render( ).
  catch zcx_abapgit_exception into lx_error.
    lo_html = zcl_abapgit_html=>create( |<span class="error">Render error: { lx_error->get_text( ) }</span>| ).
endtry.
```

src/ui/lib/zcl_abapgit_html_form_utils.clas.abap
mbtools, line 1: looks like your pretty print does not align the parameters along `type` ;-)

mbtools (commented):
see above

src/ui/core/zcl_abapgit_html.clas.abap
sbcgua: Interesting idea.

src/ui/lib/zcl_abapgit_html_form_utils.clas.abap
sbcgua: I don't usually use pretty print. I don't like its style and it's not configurable. But I'll have a look.

I think there should be a pre-merge hook to pretty print the code via abaplint. Just as an idea.

larshp: well, first if its something we want to enforce, then I'll add/find a rule for it first, then we can add automation around it if needed

mbtools: If you don't use pp, how to you capitalize the keywords? indent the code? manually?

We do have [guidelines](https://docs.abapgit.org/development-guide/read-first/guidelines.html#pretty-printer). Nothing is enforced, but if code isn't pp, then the next person using pp will get unnecessary diffs. Sure, automation is the way to do it. 

https://github.com/abapGit/abapGit/issues/4810

sbcgua: Hehe, this is a larger topic, for a bear or two at least :))
- yes, I mostly format my code manually, SE80 has a feature to add a keyword as it is typed so it's not super terrible. I actually do use PP sometimes ... for AG only ... I just don't trust the standard PP ... the abaplint has a chance though (cause it is configurable)
- ... and having said that I think there is exactly 0 sense in capitalized keywords ... but ... legacy ...
- guidelines yes, but I think autoformatting is the best and only way to do it properly. Although the best effort to keep it well formatted is of course worth it
- "next person" - yes ... though it's also a wider subject. The type formatting is only part of it. Mixing feature and refactoring, the `data:` style, the parameters indentation ... it's just more complex, not sure there is an ideal solution.

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
```diff
@@ -969,7 +950,9 @@ CLASS zcl_abapgit_gui_page_sett_remo IMPLEMENTATION.
           mo_repo->refresh( ).
         ENDIF.
 
-        rs_handled-state = zcl_abapgit_gui=>c_event_state-go_back_to_bookmark.
+        rs_handled-state = zcl_abapgit_html_form_utils=>create( mo_form )->exit(
+          io_form_data = mo_form_data
+          io_check_changes_versus = initialize_form_data( ) ).
```
mbtools, line 955: Switching tab and "back" will trigger the popup because head-type is different, but we can live with that.

mbtools (commented):
Some comments but works fine. Just needs solution for the exceptions.

src/ui/core/zcl_abapgit_html.clas.abap
sbcgua: done + UT

src/ui/lib/zcl_abapgit_html_form_utils.clas.abap
```diff
@@ -77,8 +81,16 @@ CLASS zcl_abapgit_html_form_utils IMPLEMENTATION.
   METHOD exit.
 
     DATA lv_answer TYPE c LENGTH 1.
+    DATA lo_compare_with LIKE io_check_changes_versus.
+
+    lo_compare_with = io_check_changes_versus.
+    IF lo_compare_with IS NOT BOUND.
```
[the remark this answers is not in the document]
sbcgua: Done

src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap
sbcgua: Yes, also entering values manually and exit does not trigger the change check. It's an issue of different context. Also it was so before I believe. Let's keep it as it is for now.

sbcgua:
Should be OK now. Good idea with the renderable error stub :+1:

sbcgua:
Yet there is room for improvement. Not all have been taken into account. E.g. hotkey popup. And visually it is certainly a popup so should fit the new concept (and close with Esc! =). But. It is initiated in higher-level components (in page actually). How should it block the event handlers of lower-level components? Something to think about. Not in this PR though.

mbtools (approved):
Tested. Works great! 👍

mbtools:
I went back to 1.123.0 to compare. The "back - exit/is_dirty" logic is worse than it was. You were able to switch between different settings (repo/local/remote/...) or between head types (branch/tag/...) and there would not be any popup when clicking "Back" button or pressing F3. That felt better than it is now. 

PS: There's a small bug on "Repo Settings" related to use_lxe checkbox which I will fix separately.

sbcgua:
Hmmm, then probably there was an explicit exception (in some way) to this field. I can have a look. But also one doubt: is tab switching a mere UI event or a setting change? IMO, intuitively, it is a setting change. IMO, I'm not insisting. 
For example, switching to PR and back. I would expect that switching back will revert me to the branch I was where switched to PR. Actually, it was in the initial implementation. Now it is not so. Nevermind. But my point is that in such a case switching from PR to Branch tab would be a setting change, because I don't need extra action on the form, just save it.
So ignoring the tab switching is arguable.

PR #6301, opened 2023-05-23
Move base page and HOC to lib package

sbcgua:
... because they are not pages they are a part of the framework (thus don't belong to the pages package).

Just for order, no code changes


mbtools (approved):
👍

PR #6303, opened 2023-05-27
Pull dialog safeguard

mbtools:
I'm pretty sure, I cancelled a pull dialog the other day but objects got pulled anyway overwriting a lot of uncommitted changes (I think someone else mentioned something like it before but I can't find the reference).

I reviewed the code intensely but can't find any root cause. It might be related to the exception handling.

This change adds a safeguard so no entries are returned if the popup is cancelled and the exception didn't work properly.


larshp (approved):
lgtm

PR #6309, opened 2023-06-01
Fix is_dirty check for Repo Settings

mbtools:
Fix handling of Use LXE checkbox on Repo Settings (checkboxes need to be set with `boolc`)

### DIFF
--- a/src/ui/pages/zcl_abapgit_gui_page_sett_repo.clas.abap
+++ b/src/ui/pages/zcl_abapgit_gui_page_sett_repo.clas.abap
@@ -74,7 +74,7 @@ ENDCLASS.
 
 
 
-CLASS ZCL_ABAPGIT_GUI_PAGE_SETT_REPO IMPLEMENTATION.
+CLASS zcl_abapgit_gui_page_sett_repo IMPLEMENTATION.
 
 
   METHOD constructor.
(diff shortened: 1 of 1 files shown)


sbcgua:
https://github.com/abapGit/abapGit/assets/15635498/4f9cf54a-07d7-42f0-adc1-d7a53d77944c



But I think the reason is different -> form comes empty on exit. I think it is a bug introduced somewhen earlier. That's why I used the merge approach in sett_remo, it's more consistent.

https://github.com/abapGit/abapGit/pull/6300/files#diff-26879d7167b607818e03b1a5b75b126fa79c9e720d1117c4d1954b220def6c2fR944

`mo_form_data->merge( zcl_abapgit_html_form_utils=>create( mo_form )->normalize( ii_event->form_data( ) ) ).`

mbtools:
When `is_dirty` compares the tables, one tables contains `use_lxe` with empty string(0), the other has is a single space for `use_lxe`. 

Other forms like "Local Settings" handle checkboxes the same way with `boolc`. 

https://github.com/abapGit/abapGit/blob/e6fcfd97afd0a0dd4e34a4039ac4a950abca690c/src/ui/pages/zcl_abapgit_gui_page_sett_locl.clas.abap#L326-L337

Of course, there are other ways to address this as part of the framework but I don't have time for such effort.

sbcgua:
Yes, probably these are separate issues indeed. So yes, your fix is good and relevant. It's just that there is another issue too. Maybe to be fixed separately.

PR #6317, opened 2023-06-21
added auto confirmation of transport messages

thorsten-wolf-neptune:
Implementation for avoiding SAP Transport messages that might come up during deserialization (pull) as suggested here:
https://github.com/abapGit/abapGit/issues/6314#issuecomment-1598847796

### DIFF
--- a/src/objects/zcl_abapgit_objects.clas.abap
+++ b/src/objects/zcl_abapgit_objects.clas.abap
@@ -87,12 +87,6 @@ CLASS zcl_abapgit_objects DEFINITION
         VALUE(rv_active) TYPE abap_bool
       RAISING
         zcx_abapgit_exception .
-    CLASS-METHODS confirm_transport_message
-      IMPORTING
-        !iv_msgid                   TYPE symsgid
-        !iv_msgno                   TYPE symsgno
-      RETURNING
-        VALUE(rv_message_confirmed) TYPE abap_bool .
   PROTECTED SECTION.
 
   PRIVATE SECTION.
@@ -116,7 +110,6 @@ CLASS zcl_abapgit_objects DEFINITION
     CLASS-DATA gt_obj_serializer_map TYPE ty_obj_serializer_map .
     CLASS-DATA gt_supported_obj_types TYPE ty_supported_types_tt .
     CLASS-DATA gv_supported_obj_types_loaded TYPE abap_bool .
-    CLASS-DATA gv_confirm_trans_msg_possible TYPE zif_abapgit_definitions=>ty_yes_no .
 
     CLASS-METHODS check_duplicates
       IMPORTING
@@ -216,7 +209,6 @@ CLASS zcl_abapgit_objects DEFINITION
         !iv_filename    TYPE string
       RETURNING
         VALUE(rv_extra) TYPE string.
-    CLASS-METHODS confirm_transport_messages .
 ENDCLASS.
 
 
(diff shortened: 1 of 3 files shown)


larshp:
`symsgty` is released, https://abapedia.org/steampunk-2305-api/symsgty.dtel.html

larshp:
suggest keeping the code in one method, it belongs together, and its not much code, and it will make it easier to move around

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -1268,4 +1278,72 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
...
+             v4 TYPE symsgv,
+           END OF ty_s_message.
+
+    DATA ls_message TYPE ty_s_message.
+
+    FIELD-SYMBOLS: <lt_confirmed_messages> TYPE STANDARD TABLE.
+
+    IF gv_confirm_trans_msg_possible = zif_abapgit_definitions=>c_no.
+      " auto-confirmation of transport messages is not possible in this system.
+      RETURN. ">>>>>>>>>
```
fabianlupa, line 1301: What are these for? Why do they increase in numbers for every return statement?

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -1268,4 +1278,72 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
...
+    IF sy-subrc = 0.
+      gv_confirm_trans_msg_possible = zif_abapgit_definitions=>c_yes.
+    ELSE.
+      gv_confirm_trans_msg_possible = zif_abapgit_definitions=>c_no.
+      RETURN. ">>>>>>>>>>>>>>
+    ENDIF.
+
+    ls_message-id = iv_msgid.
+    ls_message-no = iv_msgno.
+    COLLECT ls_message INTO <lt_confirmed_messages>.
```
fabianlupa, line 1320: Does collect provide any benefit here compared to insert / append?

src/objects/zcl_abapgit_objects.clas.abap
thorsten-wolf-neptune: you mean the >>> ? I like to put them to RETURN / CONTINUE / CHECK to make the stop of processing at this point more obvious to the eye when looking through code.

fabianlupa: They already seem quite obvious to me :P
image

src/objects/zcl_abapgit_objects.clas.abap
thorsten-wolf-neptune: It avoids duplicates in the table and is them same statment used by @mbtools so i just reused that. But insert should also be fine. Should i change? I don't like using append because it does not work for all table types. 
https://github.com/SAP/styleguides/blob/main/clean-abap/CleanABAP.md#prefer-insert-into-table-to-append-to

larshp: I think in this case it does not COLLECT / sum anything, so INSERT is easier to understand, as the intention is just to insert, not sum/collect anything

thorsten-wolf-neptune: ok i change

src/objects/zcl_abapgit_objects.clas.abap
thorsten-wolf-neptune: In your IDE yes but not necessarily for others. But i can remove them if it's not a common thing in this project. Should I?
image

mbtools: we don't use the ">>> in AG, so you can remove them.

mbtools:
A better place for setting the messages is `zcl/if_abapgit_cts_api`

larshp:
👍 looks good to me

mbtools:
When pulling several times, `gt_confirmed_messages` becomes bigger. Suggest storing if the method was called already in a class-variable. 

image

mbtools (approved):
Great. Thanks!

PR #6320, opened 2023-06-24
Handling unsupported objects in deserialize process

mbtools:
With this change, unsupported objects are shown with a different icon and "Object type not supported" message in the pull popup (see example below). 

I'm not aware of any option to disable a checkbox for a ALV row. So it's still possible to select unsupported objects and try to pull them, which will inevitably result in an error. Alternatively, we could remove those entries from the popup completely. Let me know what you prefer.

Also raises an exception in AFF constructor if object type is not supported (as described in [docs](https://docs.abapgit.org/development-guide/serializers/serializers.html#constructor))

Ref https://github.com/abapGit/abapGit/issues/4898#issuecomment-1601647040

Before:

image

image

After:

image

image


sbcgua:
Checkboxes can be disabled via styles. Here is a code fragment, it's for ald alv but styles are supported in SALV as well. So should be no problems to adapt.

```abap
    field-symbols <cellstyle> type lvc_t_styl.
    data ls_cellstyle like line of <cellstyle>.

    if mv_style_field is not initial.
      assign component mv_style_field of structure cs_rec to <cellstyle>.
      ls_cellstyle-fieldname  = 'CHECKBOX'. <<<
      ls_cellstyle-style      = cl_gui_alv_grid=>mc_style_disabled. <<<
      insert ls_cellstyle into table <cellstyle>.
    endif.
```

mbtools:
That disables the column, but need it for certain rows only.

sbcgua:
It's raw based

image

mbtools:
I'm no alv specialist but this looks like it's getting into this mess 
https://blogs.sap.com/2015/06/25/salv-editable-with-single-custom-method/
https://blogs.sap.com/2015/08/07/salv-and-pepper-editing-individual-columns-in-the-salv/
and then it won't work in 757... https://blogs.sap.com/2022/08/01/editable-cl_salv_table-after-release-756/

That's beyond my capabilities

sbcgua:
Essentially not (though indeed it is not obvious at all). The screenshot I posted is not an editable ALV (at least from what I remember about it :). I don't dive in the code. But from what I remember:

- making the SALV field a checkbox makes it clickable
- handling the hotstop event make it also trackable (both of these should be there already)
- adding the style field ... makes it stylable, and disableable in particular

These are really unobvious from the first glance but essentially nothing complex, regular ALV stuff

sbcgua:
zevolving findings works, and I even done it, and the work, and they are indeed complex, and indeed not super compatible. Even an early HANA systems have some specifics. Probably later releases even more. 

But, the above is not something non-standard.

sbcgua:
But, yeah, having said that, I'm also not sure that this dialog should display unsupported objects ... maybe there is another way. So I don't pretend to interfere with the solution, really I didn't studied the issue. Just wanted to mention that SALV + clickable and disablable checkboxes are possible and standard.

mbtools:
Remains as described in initial comment

larshp:
looks good to me

@sbcgua any additional comments?

sbcgua:
Conceptually no objections. The ALV thing can be done separately, I can followup on that.

sbcgua:
> Conceptually no objections. The ALV thing can be done separately, I can followup on that.

Ah no, the example I was hoping for was indeed the older ALV, not SALV. :/ In SALV I can change color and cell type but seems that the styles are not available :( Sorry for misleading

PR #6331, opened 2023-06-30
FIX and Features about Autologon to Repo

Taryck:
Fixed usage of RFC Destination for HTTP client as with PATH Prefix they failed uri is supposed to be at server root.
Enhancement: This allow a default implementation of zif_abapgit_exit~create_http_client to find the proper RFC Destination with **FULL URL matching**.= Both Server AND Path must match the URL to choose the RFC Destination.

Add feature to prevent Password dialog is we need to use ZCL_ABAPGIT_DEFAULT_AUTH_INFO even on DIALOG / Foreground.

### DIFF
--- a/src/exits/zcl_abapgit_exit.clas.abap
+++ b/src/exits/zcl_abapgit_exit.clas.abap
@@ -5,7 +5,68 @@ CLASS zcl_abapgit_exit DEFINITION
   PUBLIC SECTION.
 
     INTERFACES zif_abapgit_exit .
+  types:
+    BEGIN OF ts_rfc_definition,
+        destination             TYPE  rfcdes-rfcdest,
+        servicenr               TYPE rfcdisplay-rfcsysid,
+        server                  TYPE rfcdisplay-rfchost,
+        trace                   TYPE rfcdisplay-rfctrace,
+        language                TYPE rfcdisplay-rfclang,
+        client                  TYPE rfcdisplay-rfcclient,
+        user                    TYPE rfcdisplay-rfcalias,
+        password                TYPE rfcdisplay-rfcexec,
+        description             TYPE rfcdoc-rfcdoc1,
+        proxy_host              TYPE rfcdisplay-rfcgwhost,
+        proxy_service           TYPE rfcdisplay-rfcgwserv,
+        proxy_user              TYPE rfcdisplay-proxyuser,
+        proxy_password          TYPE string,
+        sslapplic               TYPE rfcdisplay-sslapplic,
+        path_prefix             TYPE string,
+        destlock                TYPE rfcdisplay-rfclock,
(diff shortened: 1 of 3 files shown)


src/exits/zcl_abapgit_exit.clas.abap
```diff
@@ -17,23 +79,119 @@
...
+  METHOD find_rfc_dest_from_host_path.
+* Look into preloaded RFC destination for external RFC
+* BOTH server and PATH MUST Match!
+* Need adjustment in zcl_abapgit_http=>create_by_url( )
+    LOOP AT gt_rfc_destinations ASSIGNING FIELD-SYMBOL(<d>)
+      WHERE server = iv_host
+        AND path_prefix = iv_path_name
+        AND ( created_by CS iv_user OR modified_by CS iv_user ).
+      r_destination = <d>-destination.
+      EXIT.
```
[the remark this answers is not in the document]
Taryck: EXIT in within a loop...

src/exits/zcl_abapgit_exit.clas.abap
```diff
@@ -17,23 +79,119 @@ ENDCLASS.
...
+    ENDLOOP.
+  ENDMETHOD.
 
-    lv_class_name = 'ZCL_ABAPGIT_USER_EXIT'.
 
-    IF zcl_abapgit_factory=>get_environment( )->is_merged( ) = abap_true.
```
mbtools, line 29: Don't remove this

src/exits/zcl_abapgit_exit.clas.abap
```diff
@@ -150,30 +308,6 @@ CLASS zcl_abapgit_exit IMPLEMENTATION.
   ENDMETHOD.
 
 
-  METHOD zif_abapgit_exit~change_supported_data_objects.
```
mbtools, line 153: Update your branch from origin and put the deleted methods back

mbtools (changes requested):
Hi @Taryck 

a few things:

- Don't combine several changes into one PR i.e. keep the "no dialog" separate from the RFC client exit changes
- Use pretty printer (https://docs.abapgit.org/development-guide/read-first/guidelines.html)
- Make code compatible with 702
- The change to `create_http_client` is breaking, for example, if someone already has an RFC destination that matches the repo server (but should not be used for logging on via abapGit). I like the idea but don't see how we can add this matching logic to abapGit. I suggest putting the code into your own exit class and if you like publishing it as a separate repo/gist. 
- abaplint finding of `exit in loop` is related to the incompatible 702 syntax (`assigning field-symbol()`). This kind of removes  the loop statement from the parser causing the exit to be not in a loop anymore. Issue will go away when you use 702 syntax.

Issue #6334, opened 2023-07-05
Introduce ABAP Language Version as repository setting in .abapgit.xml 


ThomasPloski:
Dear community,

Based on issue #6154, the information of ABAP Language Version is required per repository. This information should be stored in the .abapgit.xml file.

mbtools:
Closed by #6346. Work continues via #6298

PR #6335, opened 2023-07-05
Introduce ABAP Language Version (ALV) to .abapgit.xml

ThomasPloski:
fixes issue #6334


mbtools:
Code is part of the solution and looks good to me. However, I don't think we should add this as is. Every newly created repo will get diffs when used in older abapGit versions (not sure what happens if you try to pull). 

We should mark the feature to be experimental so we can merge without impacting the community. 

```abap
IF zcl_abapgit_persist_factory=>get_settings( )->read( )->get_experimental_features( ) = abap_true.
...
ENDIF.
```

ThomasPloski:
I will surround the affected coding parts with the mentioned if statement and update the PR.

ThomasPloski:
One step back: I have decided to not touch the current logic in .abapgit.xml handling, so that surrounding with experimental feature is not required, since there is no defaulting while object creation.

mbtools:
looks like these changes are included in #6346

schneidermic0 (approved):
Thanks, @ThomasPloski. LGTM

ThomasPloski:
This PR can be closed since the PR #6346 contains of those changes

PR #6336, opened 2023-07-07
Fix GUI actions for Edge browser control

### LINKED ISSUE #4841: SAP GUI for Windows: Issues with chromium-based browser control (WebView/Edge)

mbtools:
Edge Webview browser control returns upper case action but abapGit requires lower case.

Closes #4841


larshp (approved):
see issue, has been tested

PR #6338, opened 2023-07-08
Fix processing of TABU entries in case of multiple tables

### LINKED ISSUE #6337: Bug: TABU data deletion/insertion for multiple tables exited instead of continued
If the user tries to overwrite TABU entries via Pull function in multiple tables it can happen, that if you have multiple tables and only the entries of the last two tables should be changed, the process is being finished without any change as in zcl_abapgit_data_deserializer in this case the routine is being exited but not continued as the first checked table should not be changed but only the last two.

Will provide a fix.

mraht:
Closes #6337


mbtools (approved):
👍

larshp (approved):

PR #6343, opened 2023-07-10
Fix Code Inspector jump

christianguenter2:
Fixes this dump in Code Inspector result page
image
image
image
image

Side effect from #6336


mbtools (approved):
👍

Issue #6339, opened 2023-07-09
SAP GUI 8.00 SP>=03 / 7.70 SP>=13 chromium browser control not working properly


christianguenter2:
Follow up on #4841 
While chromium based browser control is basically working with SAPGUI 800 SP>=03 / 7.70 SP>=13 there are still some ~minor~ major issues. 

- [x] Hotkeys which trigger `sapevent` are not working on some pages (e.g. Repository List and Repository) while others are working (e.g. Stage)
- [ ] External links are broken. E.g. clicking on this link jumps to https://github.comchristianguenter2abapgit.git/
![image](https://github.com/abapGit/abapGit/assets/17437789/50533222-36a0-490a-b58c-671a1dce9174)
![image](https://github.com/abapGit/abapGit/assets/17437789/4859f808-42db-45ea-9b80-d2685df24cec)
- [ ] diff links are broken
![image](https://github.com/abapGit/abapGit/assets/17437789/d8cbf951-cfff-4a73-b3e0-ac2dc8e0079d)
![image](https://github.com/abapGit/abapGit/assets/17437789/09f9a1b8-5625-454e-b48a-03169b1929fb)
![image](https://github.com/abapGit/abapGit/assets/17437789/d8bfadb0-f44a-44c0-abb6-1c0c60eb71e2)

christianguenter2:
Seems there's a new special way how edge control handles `SAPEVENT`...
![image](https://github.com/abapGit/abapGit/assets/17437789/5fbd95bc-abc6-40d4-b010-85eaf5d23c55)

christianguenter2:
Links look like this
![image](https://github.com/abapGit/abapGit/assets/17437789/e6b5f90e-c41a-4ba4-a562-597c28979b77)

And are transformed to this when sent back to ABAP
![image](https://github.com/abapGit/abapGit/assets/17437789/7b44eb18-8669-4417-ba84-9a3b101c70e7)

mbtools:
The edge solution must be compatible with previous apps. a) all is upper case which is wrong and b) characters are removed (I suspect there's more than slash missing). This is some SAP nonsense that should be fixed by SAP.

christianguenter2:
I try to draw SAPs attention:
https://blogs.sap.com/2023/01/20/sap-gui-for-windows-8.00-is-coming-on-27th-of-january-2023-new-features-lifecycle-information/comment-page-1/#comment-680656

christianguenter2:
News from SAP
[...]

Issue #4661, opened 2021-03-23
CL_GUI_HTML_VIEWER and SAP GUI


mbtools:
There are various issues related the HTML viewer and it's event handling in connection with different SAP GUI types.

This collection of resources should help resolve issues:

3024355 - Webgui HTML Viewer: Restrictions compared to SAP GUI for Windows
2973428 - Reverse Tabnabbing vulnerability within SAP NetWeaver Application Server ABAP (Applications based on SAP GUI for HTML)
820706 - ITS HTML Viewer: Parameters of the event SAPEVENT
191908 - Collective note: HTML Viewer Control

From these notes: 

Method | Support in SAP GUI for HTML | abapGit usage
---------- | ---------------------------- | ----------------
close_document | starting with note 2973428 not supported any longer | only used when exiting AG
free | supported |
get_current_url | not supported | only used when viewing external pages in AG
go_back | not supported | used on all forms in AG
load_data | supported |
sapevent | supported | see below
set_registered_events | supported |
set_visible | supported |
show_url | supported |

For abapGit usage of `cl_gui_html_viewer`, see `zcl_abapgit_html_viewer_gui`.

### Use of the event SAPEVENT in SAPGUI for HTML

To initialize Event SAPEVENT in SAP GUI for HTML, all SAPEVENT definitions within an HTML document must adhere to the following formats,

in an HTML form with the POST Method: 
```html
<FORM name="myform" method="post" action="SAPEVENT:myaction?myparameter">
...
</FORM>
```

in an HTML form with the GET Method: 
```html
<FORM name="myform" method="get" action="SAPEVENT:myaction">
...
</FORM>
```

in HTML ANCHOR:
```html
<A href="SAPEVENT:myaction?myparameters">
```

in INPUT ACTION:
```html
<INPUT type="..." formaction="SAPEVENT:myaction?myparameters">
```

`<myaction>` is the user-defined action code, the maximum length is 70 characters.
`<myparameters>` are the user-defined parameters (character string), which must not contain any `?`, `!` and `/` because these are ITS-specific control characters.

### To-do for abapGit 

- Review action usage in abapGit forms
- Review usage of special characters in form output (any user input - also filenames - must be escaped properly)
- Investigate how go_back can be replaced for HTML GUI

Please add or edit

mbtools:
just for reference

PR #6345, opened 2023-07-11
Fix links for edge browser control

christianguenter2:
Latest edge browser control handles links and other url parameters quirky.  Slashes (/) are cut out when send back to ABAP and therefore the links are corrupt. To mitigate this issue this change introduces a new concept called ~short~ url parameter repository where we store ~shortened/ ~encoded ~urls~ parameters (shortening happens via hash function). Instead of the origin urls we send the ~shortened~ encoded ones to the GUI. If the GUI invokes a link the ~short url~ encoded parameter is translated back to the origin ~url and called~ and further processed.

An ~short~ encoded url looks like this: 
SAPEVENT://s_a_p_e_v_e_n_t/url?url=C715004CBA979C4877FD1959C48BF267838C32A9
instead of
sapevent://s_a_p_e_v_e_n_t/url?url=https://github.com/sbcgua/abap_w3mi_poller.git

Other example:
before
sapevent://s_a_p_e_v_e_n_t/go_fill_diff?KEY=000000000110&PATH=%2fsrc%2f&FILENAME=zcl_test_root_class.clas.abap
after
sapevent://s_a_p_e_v_e_n_t/go_fill_diff?6BB30ACCB81E6D4BE0CE40972DBF73916E27D7EC

#6339


fabianlupa:
Is this behavior documented somewhere? I wonder if this is a workaround for a bug or some kind of misused feature.

mbtools:
I had compiled this https://github.com/abapGit/abapGit/issues/4661 before for reference. 

Did you try escaping the `/` to `%2f` in the parameter part?

christianguenter2:
> Is this behavior documented somewhere? I wonder if this is a workaround for a bug or some kind of misused feature.

Not that I'm aware of

christianguenter2:
> Did you try escaping the `/` to `%2f` in the parameter part?
Yes. I tried several escaping mechanisms. Without success. But maybe I missed the obvious.

christianguenter2:
@mbtools issue also occurs with escaped links. See third task in #6339

mbtools:
Is this really what we want? I see this as a SAP bug and rather not support Edge control until it's fixed.

mbtools:
The expectation is certainly that no backend changes are necessary (to_upper/to_lower are already hacks we should not have needed). It is probably causing problems with standard SAP solutions or other customers, too. If there's no movement (no comments in the SAP notes), I can create a new case next week.

christianguenter2:
Thx. Yeah, expecation vs. reality.

christianguenter2:
Closing. Should be fixed by SAP.

PR #6392, opened 2023-08-10
Fix potential crash in background mode

ConjuringCoffee:
A runtime error can occur if `zcx_abapgit_exception` is raised before `li_log` is instatiated. I've moved the creation of the log to the beginning of the loop to fix this.


larshp (approved):
yea, as its used in the CATCH part

PR #6398, opened 2023-08-10
Fix unit test in `zcl_abapgit_objects`

mbtools:
Not sure what caused this but I'm getting errors when running UTs for `zcl_abapgit_objects`:

image

I changed the mock class to be in line with other object serializers which avoids the errors.


larshp (approved):
had the same error, just ignored it :)

PR #6422, opened 2023-08-21
Fix typo

albertmink:
Is it just my ABAP system that expects this RECEIVING type ?!

ref #6346


albertmink:
Things are bad. This need some more analysis.

albertmink:
will create an issue instead

albertmink:
will create an issue instead

Issue #6298, opened 2023-05-21
Preparation for handling ABAP language version


mbtools:
Tasks related to ABAP Language Version (#6154):

- [x] Consolidate `READ/INSERT/DELETE REPORT` statements
   (#6299)
- [x] Consolidate `CALL FUNCTION 'READ/UPDATE_PROGDIR'` 
  (#6464)
- [x] Add `ABAP_LANGUAGE_VERSION` to `.abapgit.xml` 
  (#6346)
- [x] Pull: Add logic to validate ABAP Language Version of repository against system/software component
  (#6346)
- [x] Enhance `zcl_abapgit_abap_language_vers` 
  (#6484)
- [ ] New: Add ABAP Language Version to New Online/Offline Repo 
- [x] Settings: Add ABAP Language Version to Repo Settings
  (#6488)
- [x] Object Classes: Remove ABAP Language Version from serialize methods
  (#6375)
  (#6476)
- [x] Object Classes: Add logic to set ABAP Language Version in deserialize method
  (#6476)

PR #6424, opened 2023-08-21
[INTF] Write the ABAP language version from the package

albertmink:
is part of #6426


albertmink:
@mbtools could you review this PR, as this contributes to #6298?
If this it the way how it goes, I can continue with Classes, Function Groups and some more.

mbtools:
Serializing to "standard" is "fake". I think the preferred solution is to not serialize the field at all (clear it). However this will impact existing repos so needs to be coordinated. @larshp ?

mbtools:
Replaced by #6476

PR #6429, opened 2023-08-24
Optimize tadir read

### LINKED ISSUE #6428: Optimization of `ZCL_ABAPGIT_TADIR->ZIF_ABAPGIT_TADIR~READ`
The method ZCL_ABAPGIT_TADIR->ZIF_ABAPGIT_TADIR~READ is slow for huge top packages like the SAP standard package "BASIS" with more than 400.000 objects.

The private method BUILD takes about 1 minute to select all relevant TADIR entries. Additional it does not respect the repo filter in method ZCL_ABAPGIT_SERIALIZE=>ADD_OBJECTS.

image

image

I optimized it. It takes now only about one second without the filter. 

image

I implemented additional the filter, because the private method CHECK_EXISTS also takes a long time for this top package. Maybe here would help a parallelization.

I hope that the chosen objects for the unit test with the filter exists in all systems. It exists in A4H.

image

fmabap:
Closes #6428


larshp:
guess the expected TADIR entries doesnt exist in open-abap

anyhow, suggest removing the assumptions, move the test to `/test/`, and use osql test doubles for TADIR, example in https://github.com/open-abap/open-abap-core/blob/main/src/unit/osql/cl_osql_test_environment.clas.testclasses.abap

test/zcl_abapgit_tadir_test.clas.testclasses.abap
```diff
@@ -0,0 +1,169 @@
+CLASS ltcl_build DEFINITION DEFERRED.
+
+CLASS ltcl_build DEFINITION FOR TESTING RISK LEVEL HARMLESS DURATION SHORT FINAL.
```
larshp, line 3: its not deferred, its defined in the next statement, MEDIUM runtime will make it execute later than the tests in /src/

```suggestion
CLASS ltcl_build DEFINITION FOR TESTING RISK LEVEL HARMLESS DURATION MEDIUM FINAL.
```

larshp:
https://github.com/open-abap/open-abap-core/pull/761

src/zif_abapgit_definitions.intf.abap
```diff
@@ -3,6 +3,13 @@ INTERFACE zif_abapgit_definitions
 
   TYPES:
     ty_string_tt TYPE STANDARD TABLE OF string WITH DEFAULT KEY .
+  TYPES:
+    BEGIN OF ty_obj,
+      obj_type TYPE tadir-object,
+      obj_name TYPE tadir-obj_name,
+    END OF ty_obj.
+  TYPES:
+    ty_obj_tt TYPE STANDARD TABLE OF ty_obj WITH DEFAULT KEY.
```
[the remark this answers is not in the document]
fmabap: I changed it from binary search to sorted table with non-unique key obj_type and hased unique key obj_type_obj_name

test/zcl_abapgit_tadir_test.clas.testclasses.abap
fmabap: I adjusted it, but now the unit test failed and I don't know why. In Eclipse in the src folder was it ok.
File \test\zcl_abapgit_tadir_test.clas.testclasses.abap

fmabap: image

larshp: fixed with https://github.com/abapGit/abapGit/pull/6429#issuecomment-1691678003

larshp:
keep the scope of this PR to optimize the TADIR read? as per https://github.com/abapGit/abapGit/issues/6428

larshp:
https://github.com/abaplint/transpiler/pull/1304

fmabap:
The only scope of this PR is to optimize the TADIR read as per https://github.com/abapGit/abapGit/issues/6428

fmabap:
> > The only scope of this PR is to optimize the TADIR read as per #6428
> 
> then I dont expect any changes to `src/objects/core/zcl_abapgit_serialize.clas.abap`, https://github.com/abapGit/abapGit/pull/6429/files#diff-e79d9e936a1a8f68ab662c89a2374f7e84b86b8b3686e3677971ad362ae23ef9

The changes in this class zcl_abapgit_serialize are required to pass the filter to the TADIR class.

larshp:
how can the performance problem be reproduced in the current codebase? And if so, which parts are part of that performance optimization, vs which parts is the new unused functionality being introduced?

fmabap:
> how can the performance problem be reproduced in the current codebase? And if so, which parts are part of that performance optimization, vs which parts is the new unused functionality being introduced?

You would need a big repository to reproduce directly in the system. I created the demo branch https://github.com/fmabap/abapgit/tree/demo_%236428_issue to show the issue with a unit test on the SAP Standard top package BASIS with 400.000 objects and multiple namespaces.
image

The main issue was the check if the namespace was alread determined. There was a check for each object with a namespace in the table with the 400.000 entries if the namespace exists, else it would be added.

image

I solved it be adding an additional table that contains only the namespaces. There should be not more than 100 namespaces in real live.

image

image

We use the filter functionality daily in our project with the Stage by Transport functionality.

image

Passing the filter to the tadir class avoids the CHECK_EXISTS for the not relevant objects. 

image

fmabap:
@larshp  I added more infos to the fail of the unit tests.
image

As mentioned it works fine in Eclipse
image

It looks like that the transpiler has an issue with these read tables
image

May be it is because of the special type of the it_filter table that @mbtools requested.

image

src/objects/sap/zif_abapgit_sap_namespace.intf.abap
```diff
@@ -3,14 +3,22 @@ INTERFACE zif_abapgit_sap_namespace
...
+      iv_namespace  TYPE trnspace-namespace
     RETURNING
       VALUE(rv_yes) TYPE abap_bool.
 
+  METHODS split_by_name
+    IMPORTING
+      iv_obj_with_namespace    TYPE tadir-obj_name
+    EXPORTING
+      ev_namespace             TYPE trnspace-namespace
+      ev_obj_without_namespace TYPE tadir-obj_name
```
larshp, line 21: returning a structure will make it possible to do chaining in 702

fmabap: I adjusted it.

larshp:
suggest moving the refactoring of calls to `RS_NAME_SPLIT_NAMESPACE` to a separate PR

fmabap:
> suggest moving the refactoring of calls to `RS_NAME_SPLIT_NAMESPACE` to a separate PR

I created the PR #6443 for it.

fmabap:
@larshp the unit test for the namespace filter works after your fix, but the unit test for the filter of the local package $SWF_RUN_CNT still fails on github, but in eclipse it is fine.

image

The packages are selected here 

image

The TADIR entry for them are created here

image

Can you please check, why this unit test does not work on github?

fmabap:
@larshp 
I added additional unit tests to find the issue of the abap transpiler.

It looks like that issue is this while loop.

image

I added unit tests that did the selection without the while they work also on github

image

Can you please fix it?

fmabap:
@larshp 
The PR #6445 shows the transpile error in the new test class abapGit\test\zcl_abapgit_sap_package_test.clas.testclasses.abap without the new logic.

fmabap:
All checks passed 🎉🎉🎉 => ready for review

fmabap:
This PR has been splitted into #6461 and #6462

PR #6461, opened 2023-09-01
Optimize add namespace

fmabap:
#6428


fmabap:
> this change adds the namespaces to a separate table, which is then used for reading, which is the optimization?
Yes. So we have a fast read on a small table with max. 100 entries in real life instead of 400.000 entries like in the test with the BASIS package explained here in detail.
https://github.com/abapGit/abapGit/pull/6429#issuecomment-1693354417

larshp (approved):

PR #6463, opened 2023-09-01
Replace `is_sap_cloud_platform` with `is_cloud`

mbtools:
Replace deprecated method (see `cl_cos_utilities`)

PS: `get_system_language_filter` is just re-order


mbtools:
sorry, i misread a comment about this method.

Issue #6481, opened 2023-09-11
Warning about Edge browser control is displayed even when Internet Explorer browser control is used


kjetil-kilhavn:
In one of our systems (sandbox) the warning about Edge browser control is displayed regardless of which browser control I use. 
The development system in the same system line behaves as expected and only displays the warning when I use the Edge (Chromium-based) browser control.

After chasing ghosts such as system parameters and security settings I discovered that the reason is a repository with quotation marks in the display name - which causes a Javascript syntax error in the "repoCatalog" array.  
The syntax error message was: Expected '}'

Repository display name: ABAP2XLSX Demonstrate loss of formatting when using "Format as table"

mbtools:
Closed by #6482

PR #6482, opened 2023-09-11
Escape display name for javascript repo catalog

kjetil-kilhavn:
Escape display name for javascript repository catalog. Without this escaping a display name containing double quote characters caused Javascript syntax error, which manifested itself through the warning about Edge browser control being displayed although the Internet Explorer browser control was being used.

Solves #6481


larshp (approved):
LGTM

PR #6487, opened 2023-09-12
Prepare `FUGR` for ABAP Language Version

mbtools:
Set consistent language version for function group and includes

### DIFF
--- a/src/objects/zcl_abapgit_object_fugr.clas.abap
+++ b/src/objects/zcl_abapgit_object_fugr.clas.abap
@@ -75,6 +75,7 @@ CLASS zcl_abapgit_object_fugr DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
       IMPORTING
         !it_functions TYPE ty_function_tt
         !ii_log       TYPE REF TO zif_abapgit_log
+        !iv_version   TYPE uccheck
         !iv_package   TYPE devclass
         !iv_transport TYPE trkorr
       RAISING
@@ -101,6 +102,7 @@ CLASS zcl_abapgit_object_fugr DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
     METHODS deserialize_xml
       IMPORTING
         !ii_xml       TYPE REF TO zif_abapgit_xml_input
+        !iv_version   TYPE uccheck
         !iv_package   TYPE devclass
         !iv_transport TYPE trkorr
       RAISING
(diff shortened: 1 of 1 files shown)


larshp (approved):
LGTM

src/objects/zcl_abapgit_object_fugr.clas.abap
```diff
@@ -75,6 +75,7 @@ CLASS zcl_abapgit_object_fugr DEFINITION PUBLIC INHERITING FROM zcl_abapgit_obje
       IMPORTING
         !it_functions TYPE ty_function_tt
         !ii_log       TYPE REF TO zif_abapgit_log
+        !iv_version   TYPE uccheck
```
schneidermic0, line 78: Do we need this parameter? Function modules (as all sub other sub objects) don't specify their "own" ABAP language version. They have the same ABAP language version as their function group.

src/objects/zcl_abapgit_object_fugr.clas.abap
```diff
@@ -1204,12 +1204,16 @@ CLASS zcl_abapgit_object_fugr IMPLEMENTATION.
...
+          lv_abap_version TYPE trdir-uccheck,
           lt_functions    TYPE ty_function_tt,
           lt_dynpros      TYPE ty_dynpro_tt,
           ls_cua          TYPE ty_cua.
 
+    lv_abap_version = get_abap_version( io_xml ).
+
     deserialize_xml(
       ii_xml       = io_xml
+      iv_version   = lv_abap_version
```
schneidermic0, line 1216: Is there a reason why you called the parameter just "version"? I even saw in your some other PR (still draft) you renamed "version" to "state",

When I see a parameter called "version" related to API's for ABAP repository objects, I assume it's the flag for the active/inactive version. If "version" is the ABAP Language version, this might be misleading.

Therefore, I would add something to the parameter name to indicate it is the ABAP Language version (e.g., `iv_abap_version`, `iv_abap_language_version`, ...)

What do you think?

schneidermic0 (commented):
Hey @mbtools,

Thanks for taking up this topic. 👍 
I just went over the PR and have some small comments. Maybe, you want to check them.

src/objects/zcl_abapgit_object_fugr.clas.abap
mbtools: I tried to stick with `version` (`iv_version`) when it's related to `INSERT/READ REPORT` ie. the classic "Unicode" flag for source code. It's also shorter and avoids having to realign parameters :wink:.

Best to see the final suggested state in #6476

schneidermic0: > I tried to stick with version (iv_version) when it's related to INSERT/READ REPORT 

OK, I see. I still would prefer to make it more explicit to differentiate. For sure, this is up to you. :)
Feel free to set this as resolved.

PR #6494, opened 2023-09-14
ECTD: Fix diff after import based on sort order

### LINKED ISSUE #6489: ECTD diff after import based on sort order

mbtools:
Clears `SORTLNR` nodes which were causing diffs. These are automatically filled during deserialize. 

Works fine with the test repo. @fabianlupa would be nice to test with a larger `ECTD` case that has more parameter variants.

Repo: https://github.com/abapGit-tests/ECATT
Branch: test_sortlnr

Closes #6489

### DIFF
--- a/src/objects/zcl_abapgit_object_ecatt_super.clas.abap
+++ b/src/objects/zcl_abapgit_object_ecatt_super.clas.abap
@@ -120,6 +120,12 @@ CLASS zcl_abapgit_object_ecatt_super DEFINITION
         CHANGING
           ci_document TYPE REF TO if_ixml_document,
 
+      clear_element_collection
+        IMPORTING
+          iv_name     TYPE csequence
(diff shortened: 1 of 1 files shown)


fabianlupa:
Thanks! I added a more complex test case in https://github.com/abapGit-tests/ECATT/commit/f6517c5f1f61819c2bbb45c65c18e4812445172a Maybe you can also try pulling that to see if there are any diffs? On my end this issue seems to be fixed by this.

fabianlupa (approved):

PR #6496, opened 2023-09-15
SAJC,SAJT: Job Catalog Entry and Job Template

mbtools:
Add support based on generic serializer

Test repos:
https://github.com/abapGit-tests/SAJC
https://github.com/abapGit-tests/SAJT

Ref #5912


larshp (changes requested):
these 2 have AFF definitions, so start these as json instead of having to refactor later?

https://github.com/SAP/abap-file-formats/tree/main/file-formats/sajc
https://github.com/SAP/abap-file-formats/tree/main/file-formats/sajt

mbtools:
Oh, the aff list has grown quite a bit. That makes this one pointless.

PR #6498, opened 2023-09-15
Move up ON_EVENT exit call

### LINKED ISSUE #6497: Exit ON_EVENT does not allow overriding page specific actions
From https://docs.abapgit.org/user-guide/reference/exits.html#on-event:

> This exit allows you to extend abapGit with new features that are not suitable for abapGit itself. For example, you can link to a new page from a wall message. __Another use case is redirecting menu items to a custom page rather than standard abapGit__, for example using a company-specific solution to replace "Advanced > Run Code Inspector" (see #4722).

The highlighted part currently only seems to work specifically for general menu items, not for actions a level deeper on pages. I am attempting to override the navigation from the repo list to the repo page. However, the exit is called in `ZCL_ABAPGIT_GUI_ROUTER->ZIF_ABAPGIT_EVENT_HANDLER~ON_EVENT`. Which is _late_. `ZCL_ABAPGIT_GUI->HANDLE_ACTION` for this use case has 3 event handlers to go through with `ZCL_ABAPGIT_GUI_ROUTER` being the last one in the list. The action never reaches the exit as it is handled by the "default" implementation beforehand.

image

Technically having the general page routing last seems correct. I'd say the exit call should be moved though from:

https://github.com/abapGit/abapGit/blob/b1daa712268f14e2a0282e2640529603fa9bbc05/src/ui/routing/zcl_abapgit_gui_router.clas.abap#L674

To before the loop:

https://github.com/abapGit/abapGit/blob/b1daa712268f14e2a0282e2640529603fa9bbc05/src/ui/core/zcl_abapgit_gui.clas.abap#L285

Unless this is intentionally not supported and only new custom events should be supported as well as overriding specifically actions handled by the GUI router.

fabianlupa:
Fixes #6497

### DIFF
--- a/src/ui/core/zcl_abapgit_gui.clas.abap
+++ b/src/ui/core/zcl_abapgit_gui.clas.abap
@@ -282,12 +282,16 @@ CLASS ZCL_ABAPGIT_GUI IMPLEMENTATION.
         it_postdata     = it_postdata.
 
     TRY.
-        LOOP AT mt_event_handlers INTO li_handler.
-          ls_handled = li_handler->on_event( li_event ).
-          IF ls_handled-state IS NOT INITIAL AND ls_handled-state <> c_event_state-not_handled. " is handled
-            EXIT.
-          ENDIF.
-        ENDLOOP.
+        ls_handled = zcl_abapgit_exit=>get_instance( )->on_event( li_event ).
+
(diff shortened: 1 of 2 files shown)


mbtools (approved):
👍

PR #6499, opened 2023-09-15
Add option to remove all objects but keep repository

mbtools:
"Remove Objects" is like "Uninstall" but keeping the repository in abapGit. After "Remove Objects" you can pull again to get a fresh clone of the repo. 

New Advanced Menu:
- Remove Repository (same as previous "Remove")
- Remove Objects (same as "Uninstall" but keeps the repository in abapGit)
- Uninstall (same as before)

image

To-do: Update docs


fabianlupa:
After testing https://github.com/abapGit-tests/CLAS_exceptions I see why that would be useful :)

:+1:

mbtools:
yeah, reinstalled that repo probably 30x... then I had enough. Isn't that how innovation works?

larshp (approved):

PR #6509, opened 2023-09-21
Fix link hints for radio buttons with edge control

christianguenter2:
Link hints for radio buttons were not working with edge browser control. This changes fixes that.

abapgit


mbtools (approved):
good 👍

PR #6512, opened 2023-09-27
Encapsulate locks for background mode

ConjuringCoffee:
If you have an application interacting with abapGit running in the background, then you want to ensure it doesn't run at the same time as the regular abapGit background job. Today, it is necessary to duplicate the enqueue and dequeue from method `RUN` of `ZCL_ABAPGIT_BACKGROUND` to achieve this. This could cause problems for custom applications if the enqueue type of the background mode was ever changed.

One such example would be [`ZABAPGIT_SCHEDULER` from Marc Bernard's tools](https://github.com/Marc-Bernard-Tools/ABAP-Tools-for-abapGit/blob/main/src/zabapgit_scheduler.prog.abap).

This pull request aims to solve that problem by encapsulating the `ENQUEUE` and `DEQUEUE` of the background mode into public methods.

### DIFF
--- a/src/background/zcl_abapgit_background.clas.abap
+++ b/src/background/zcl_abapgit_background.clas.abap
@@ -14,15 +14,23 @@ CLASS zcl_abapgit_background DEFINITION
     CLASS-METHODS run
       RAISING
         zcx_abapgit_exception .
+
     CLASS-METHODS list_methods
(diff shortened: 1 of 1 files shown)


mbtools (approved):
Looks good! Are you going to add a PR for my scheduler, too? 😉

PR #6516, opened 2023-09-29
Fix unit tests in `zcl_abapgit_abap_language_vers`

mbtools:
Limit tests for cloud package only if it's available

Ref #6484


larshp:
image

larshp (approved):

PR #6519, opened 2023-09-29
SFxx: Fix missing lock checks and unlocking

mbtools:
After #6517

Objects of the switch framework didn't check for existing locks and were not unlocked after deserializing.


larshp (approved):
lgtm

PR #6540, opened 2023-10-13
Fix syntax errors in latest abapGit

christianguenter2:
#6537
grafik


mbtools (approved):
we're going over the speed limit. thx! 😀

PR #6553, opened 2023-10-20
Refac: use constants

albertmink:
Facilitate future changes when eventually dropping the experimental feature flag (ALAV and AFF).


mbtools (approved):
👍

PR #6563, opened 2023-10-24
Fix "lines between methods"

mbtools:
Fix some cases causing diffs on lower releases

Examples:

image

image


larshp (approved):

mbtools:
we could make `https://rules.abaplint.org/newline_between_methods/` a target rule: 

```json
{
  "count": 2,
  "logic": "exact"
}
```

PR #6572, opened 2023-10-25
Pretty-print a couple files

mbtools:


larshp (approved):
fine, but some of this SE24 will break again

PR #6601, opened 2023-10-31
Improve documentation of ZCL_ABAPGIT_OBJECT_COMMON_AFF

wurzka:
The current lack of this documentation makes it unclear what these methods do and how overwriting in the framework affects the operations. 

This is improved by this pull request.


mbtools:
Thanks @wurzka! Doc in the code is nice, but we should also update https://docs.abapgit.org/development-guide/serializers/ to include AFF. I created https://github.com/abapGit/docs.abapgit.org/issues/180

mbtools (approved):

PR #6616, opened 2023-11-02
Clean-up `select_add_order_by`

mbtools:
Just a start


mbtools:
looks like a false positive in `zcl_abapgit_object_intf` since the itab is sorted.

larshp (approved):

PR #6638, opened 2023-11-09
Consolidate offline and online repo view

### LINKED ISSUE #6129: Show local state for offline repos
This is a question related to #6128, raised by @mbtools 

Intention is to have more visibility on offline repo status (in particular useful for deployments):
- was it changed after the last sync ?
- last change user
- maybe file structure
- maybe serialized code (though this probably has very little value since it's easy to drill down in the object which is almost the same)
- unfortunately the real diff is not possible since there is nothing to compare with

It looks like this
image
image

Now question is, if this feature should be switchable or be the only default (so an improvement over the current view) ?
image

The new approach is certainly not worse or better than the old one, **except one thing** - speed of rendering. Offline repos do not serialize to be displayed, just TADIR is being read. (`ZCL_ABAPGIT_REPO_CONTENT_LIST->BUILD_REPO_ITEMS_LOCAL_ONLY`). Serialization takes time for larger packages.

**Question: is the speed important in this scenario ?** 

For me: probably not so much, I'd gravitate to making the new approach a default and the only one. Also it is a potential code simplification (though it is secondary). Are there any opinions, cases, experiences ?

mbtools:
The object list of offline and online repos was determined in different ways. Offline had no serializing which had side effects like missing `.abapgit.xml` in the view. 

This change consolidates the logic so offline repos behave the same as online. When you open an offline repo, objects will be serialized. `.abapgit.xml` and `.apack-manifest.xml` will be visible immediately.

Before:

image

After:

image

Furthermore, this moves the `has_remote_source` method to `zcl_abapgit_repo`. Only `reset_remote` remains in `zcl_abapgit_repo_offline`. 

Closes #6129

### DIFF
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -35,6 +35,8 @@ CLASS zcl_abapgit_repo DEFINITION
       FOR zif_abapgit_repo~refresh .
     ALIASES set_dot_abapgit
       FOR zif_abapgit_repo~set_dot_abapgit .
+    ALIASES has_remote_source
+      FOR zif_abapgit_repo~has_remote_source .
 
     METHODS bind_listener
       IMPORTING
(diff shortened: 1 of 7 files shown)


larshp (approved):
lgtm

PR #6640, opened 2023-11-10
Fix deletion/uninstall when using local settings

mbtools:
If the local options "ignore subpackages" or "only local objects" are turned on, then deleting all objects or uninstalling the repository will also delete all subpackages and objects included in the subpackages and objects that are not local to the system.

Reproduce issue:

1. Create repo for https://github.com/abapGit-tests/DEVC_main_sub, branch `with_programs`
2. Pull all objects
3. Go to local settings and turn on "ignore subpackages" (sub-package and sub-program show as diffs which is ok)
4. Select "Advanced > Remove Objects" or "Uninstall"
5. Popup shows "4 objects" although the sub-package/program should be ignored
6. Confirm
7. Sub-package and program are deleted

The fix centralizes the tadir object selection and ensures that local settings "ignore subpackages" or "only local objects" are respected when selecting objects.

### DIFF
--- a/src/repo/zcl_abapgit_repo.clas.abap
+++ b/src/repo/zcl_abapgit_repo.clas.abap
@@ -856,26 +856,10 @@ CLASS zcl_abapgit_repo IMPLEMENTATION.
 
   METHOD zif_abapgit_repo~get_tadir_objects.
 
-    DATA:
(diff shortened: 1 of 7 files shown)


larshp (approved):
lgtm

PR #6646, opened 2023-11-13
ZCL_ABAPGIT_SERIALIZE / DETERMINE_MAX_PROCESSES refactoring

christianguenter2:
- simplify conditional logic
- add more unit tests
- mock exit

### DIFF
--- a/src/zif_abapgit_environment.intf.abap
+++ b/src/zif_abapgit_environment.intf.abap
@@ -34,4 +34,10 @@ INTERFACE zif_abapgit_environment
   METHODS is_variant_maintenance
     RETURNING
       VALUE(rv_is_variant_maintenance) TYPE abap_bool.
+  METHODS init_parallel_processing
+    IMPORTING
+      iv_group                      TYPE rzlli_apcl
+    RETURNING
+      VALUE(rv_free_work_processes) TYPE i.
+
 ENDINTERFACE.
--- a/src/zif_abapgit_environment.intf.abap
+++ b/src/zif_abapgit_environment.intf.abap
@@ -36,7 +36,7 @@ INTERFACE zif_abapgit_environment
       VALUE(rv_is_variant_maintenance) TYPE abap_bool.
   METHODS init_parallel_processing
     IMPORTING
-      iv_group                      TYPE rzlli_apcl
+      iv_group                      TYPE clike
     RETURNING
       VALUE(rv_free_work_processes) TYPE i.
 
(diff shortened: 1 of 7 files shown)


src/zif_abapgit_environment.intf.abap
```diff
@@ -34,4 +34,10 @@ INTERFACE zif_abapgit_environment
   METHODS is_variant_maintenance
     RETURNING
       VALUE(rv_is_variant_maintenance) TYPE abap_bool.
+  METHODS init_parallel_processing
+    IMPORTING
+      iv_group                      TYPE rzlli_apcl
```
larshp, line 39: so the INTF is ABAP cloud compatible, 

```suggestion
      iv_group                      TYPE clike
```

christianguenter2: Thx. Fixed.
How can I detect these kind of problems in advance?

mbtools:
the UTs fail because I'm using an exit:

image

```abap
  METHOD zif_abapgit_user_exit~change_max_parallel_processes.
    cv_max_processes = 4.
  ENDMETHOD.

  METHOD zif_abapgit_user_exit~change_rfc_server_group.
    cv_group = 'abapgit'.
  ENDMETHOD.
```

Maybe add a check...

```abap
  METHOD determine_max_processes_free.

    CHECK mo_cut->mv_group = 'parallel_generators'. "<<<

    given_parallel_proc_disabled( abap_false ).
    ...
```

christianguenter2:
> the UTs fail because I'm using an exit:
> 
> image
> 
> ```abap
>   METHOD zif_abapgit_user_exit~change_max_parallel_processes.
>     cv_max_processes = 4.
>   ENDMETHOD.
> 
>   METHOD zif_abapgit_user_exit~change_rfc_server_group.
>     cv_group = 'abapgit'.
>   ENDMETHOD.
> ```
> 
> Maybe add a check...
> 
> ```abap
>   METHOD determine_max_processes_free.
> 
>     CHECK mo_cut->mv_group = 'parallel_generators'. "<<<
> 
>     given_parallel_proc_disabled( abap_false ).
>     ...
> ```

Good catch, thx.

Hm, wouldn't be mocking/doubling the exit the right thing to do?

fabianlupa:
Yes I had the same experience in the original PR. I'd suggest disabling calls to exit implementations in test contexts (factory method could probably check for sy-repid or if a test class is accessible). Bonus would be mocking in tests.
I am more concerned of someone implementing an exit and someone else executing the "harmless" unit tests in the same system.

mbtools:
mocking would work. it's just more complicated since it's not forseen yet.

christianguenter2:
Ok, let's start with mocking the exit. Disabling calls to exit implementation in test context isn't as easy as it seems. E.g. `SY-CPROG` depends on various factors (SAPGUI/ADT/NW release...). But I agree, that's something we should think about.

mbtools (approved):
Works great. Thanks

PR #6650, opened 2023-11-16
Sort abaplint.json

mbtools:
- Make it easier to find a rule
- Add `dynpro_checks`, `form_tables_obsolete`, `message_exists`

image


larshp (approved):
I would like to have the `global` settings first, but anyhow

PR #6662, opened 2023-11-24
disable superfluous check for offline repos

### LINKED ISSUE #6660: Repo: Switch to Offline not possible
Given any online repo, goto Remote Settings, Switch to Offline and save

grafik
grafik
No message appears and when I try to navigate back I get this message and the changes are not saved
grafik
After 'Switch to Online' it looks like this
grafik
The validation is from here
https://github.com/abapGit/abapGit/blob/e9d8c1d985158360a67f8dcb9b0660902e512e96/src/ui/pages/zcl_abapgit_gui_page_sett_remo.clas.abap#L849-L853
grafik

I assume it's a side effect from #6635? How has the check to be adjusted? We can we remove it?

christianguenter2:
fixes #6660


mbtools (approved):
Thanks 👍

PR #6663, opened 2023-11-24
Add "Ignore" option for ABAP Language Version

mbtools:
Introduce new mode that will ignore the ABAP Language Version when serializing objects ("Lars Mode").

Ref #6154


larshp (approved):
lgtm

PR #6679, opened 2023-11-29
Fix running multiple instances of abapGit

mbtools:
The second start can result in a lock due to checksum migration (`ezabapgit`). However, the update/lock is unnecessary if nothing has changed.

image


larshp (approved):
nice

PR #6681, opened 2023-11-30
Check for ABAP language version during serialize

mbtools:
If the repository is set to a specific ABAP language version (standard, key user, cloud), objects that do not match the setting will raise an error message.

If the ABAP language version is undefined or ignored, there's no check/error.

Also more tests and fixes.

Ref #6154 

Examples:

image

image


larshp (approved):
lets go

PR #6709, opened 2023-12-19
CLAS: Avoid popups in lower releases

mbtools:
When deserializing exception classes in lower releases (for example 702), you get a popup asking for the package of the exception class. This fix sets the correct package and avoids the popup.


larshp (approved):
lgtm

PR #6711, opened 2023-12-19
FUGR: Fix check if include belong to function grp

mbtools:
Needs to compare complete `tadir` key to distinguish between regular and exit function groups


larshp (approved):
ahh, okay

PR #6716, opened 2023-12-21
Fix abapGit popups external call

christianguenter2:
If abapGit popups are called outside abapGit toolbar buttons are sometimes missing, because status `102` or `110` of `SAPMSVIM` is used.
image

E.g. `zcl_abapgit_ui_factory=>get_popups( )->popup_to_select_from_list`

before
image

after
image


mbtools (approved):
👍

PR #6737, opened 2024-01-07
Check class/interface for version is part of repo

mbtools:
- The class or interface used for the version constant must be part of the repository
- Upper-case name of version constant
- Minor text adjustments

Example: Try using a constant from another package in abapGit

image


larshp (approved):
might be a bit slow to read all the TADIR for the repo, but guess its okay

PR #6742, opened 2024-01-08
Refactor ZCL_ABAPGIT_OO_FACTORY

christianguenter2:
- restrict instantiation
- friendship to factory
- align method naming to other factories
- remove unused gi_object_oriented_object 
- use factory instead of direct instantiation in zcl_abapgit_object_clas


mbtools:
I agree with the changes in general but I have a use case subclassing `oo_class/interface` that won't be possible anymore. Can we keep those `CREATE PUBLIC`, please?

christianguenter2:
Sure, made them public again.

mbtools (approved):
👍

PR #6745, opened 2024-01-09
DDLS: deletes remaining TADIR entry during delete

stockbal:
If a DDLS object is locally deleted - e.g. during a pull operation - the corresponding DDLS TADIR entry is not deleted.  
Calling the function `WB_TREE_UPDATE_OBJECTLIST` will remedy this problem.

This is the same approach as implemented in the Object delete logic of ADT
image

Because `corr_insert( )` may recreate the TADIR entry, the function module needs to be called at the end of the delete method.


larshp (approved):
lgtm

PR #6752, opened 2024-01-15
Create package popup: Fix prefill package name

christianguenter2:
regression from #6656

before
grafik

after
grafik


larshp (approved):
thanks 👍

PR #6756, opened 2024-01-18
BG: Ensure auth buffer is cleared even in error case

ConjuringCoffee:
I noticed this during the review of #6736.

I'm not sure _exactly_ why the auth buffer must be cleared, but this change ensures that it is cleared even if there is an error when processing or refreshing the repository.


mbtools (approved):
👍

PR #6787, opened 2024-02-01
Ignore macOS "Desktop Services Store" files

BeckerWdf:
In the Apple macOS operating system, .DS_Store is a file that stores custom attributes of its containing folder, such as folder view options, icon positions, and other visual information.
See: https://en.wikipedia.org/wiki/.DS_Store

We don't want to track them in our git repo.


larshp (approved):
thanks :+1

PR #6793, opened 2024-02-02
Fix ABAP Language Version Handling for Repository Setting "Any" for AFF Object Types

### LINKED ISSUE #6791: Wrong ABAP language version is stored for object types handled by AFF common handler
If an abapGit repository is pulled into a system and the repository's setting for the ABAP language version is set to "Any" (*), it seems that the particular value is passed to AFF object handlers and stored in the data base.

As soon as the object shall be changed again you face locking issues because ABAP language version "*"  is not supported. 

If ABAP language version is set to "Ignore" (-), it seems to work (by accident), because AFF framework ignores this value as no ABAP language version is set.

schneidermic0:
Fixes #6791


schneidermic0:
@mbtools Thanks for the review. 

I have tested it with a repository containing object types CHKO, CHKC and CHKV. For these it worked fine, now.

Issue #6798, opened 2024-02-05
Bug Report: Command Palette Alignment Issue on 'Repository List' Page.


francisco-milan:
Dear abapGit Team,

I wanted to report a bug related to the alignment in the Command Palette (F1) on the "Repository List" page. As you can see in the image below, the items "Only Favorites" and "Display" are not aligned with the rest of the elements in the list. 

![image](https://github.com/abapGit/abapGit/assets/36721657/61eb9ba0-d433-495a-9420-059981592ffb)

This misalignment is a result of intentionally putting a white space in the Toolbar between the check mark icons and the corresponding text.

![image](https://github.com/abapGit/abapGit/assets/36721657/69d215e1-2b42-452c-9abe-36ec63a4bf18)

To improve the alignment, I initially proposed a solution by adjusting two lines of ABAP code within the Method "RENDER_FILTER_BAR", part of the Class "ZCL_ABAPGIT_GUI_PAGE_REPO_OVER".

![image](https://github.com/abapGit/abapGit/assets/36721657/c321ec83-be78-467d-a283-38b9a5b79957)

Instead of: (Line 23)
      iv_txt   = |<i id="icon-filter-favorite" class="icon icon-check { lv_icon_class }"></i> Only Favorites|
I suggested using: (Line 24)
      iv_txt  &nbsp; = &vert;&lt;i id=&quot;icon-filter-favorite&quot; class=&quot;icon icon-check &lbrace; lv_icon_class &rbrace;&quot;&gt;&lt;/i&gt;&vert; && &grave; &grave; && &vert;Only Favorites&vert;

And for the "Detail" item, instead of: (Line 28)
      iv_txt   = '<i id="icon-filter-detail" class="icon icon-check"></i> Detail'
I recommended using: (Line 29)
      iv_txt &nbsp;  = '&lt;i id="icon-filter-detail" class="icon icon-check"&gt;&lt;/i&gt; ' && &grave; &grave; && 'Detail'

And here's the result:

![image](https://github.com/abapGit/abapGit/assets/36721657/e010860b-f6eb-4d8a-ad44-93c97ef05886)
![image](https://github.com/abapGit/abapGit/assets/36721657/5c9bca9b-193d-4328-ae99-37ae00abbadd)

[...]

PR #6799, opened 2024-02-06
Fix bug in Command Palette

francisco-milan:
Fix issue in Command Palette ("Only Favorites" and "Detail") following @mbtools' recommendation of trimming the text in JavaScript.
Details of this issue can be found in #6798.


mbtools (approved):
👍

PR #6808, opened 2024-02-09
Remove "Submit Query" from Command Palette.

### LINKED ISSUE #6807: Form action "Submit Query" in Command Palette
Hi abapGit Team,

  Today I was checking the Command Palette on the "Repository List" page and I noticed something interesting. It seems like the "Submit Query" action doesn't quite fit in there since it's more of a form thing.

image

I've got a simple fix in mind. We could just modify this line a little bit:

https://github.com/abapGit/abapGit/blob/ddd6bdb3326f0196193fb3d09d481369323d4267/src/ui/zabapgit_js_common.w3mi.data.js#L2376

**Proposed change:**
```
[].slice.call(document.querySelectorAll("input[type='submit'][value]:not([value='Submit Query'])"))
```

This would filter out the "Submit Query" action from the Command Palette, keeping things tidy and focused on what it's meant for.

image

What do you think? If it sounds good, I can create a new PR to apply this change.

Best regards,

Francisco Milan.

francisco-milan:
This PR is to remove the "Submit Query" action from the Command Palette ("Repository List" page). 
Closes #6807.

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -2373,7 +2373,7 @@ function enumerateUiActions() {
   });
 
   // forms
-  [].slice.call(document.querySelectorAll("input[type='submit']"))
+  [].slice.call(document.querySelectorAll("input[type='submit'][value]:not([value='Submit Query'])"))
     .forEach(function(input) {
       items.push({
         action: function() {


sbcgua:
I'd recommend marking `unlistable` commands in html and not hard-fix it in js. There might be more commands like that, now or in the future, or the text might change. So just add a css class (even without css code - just as a marker). Smth like `not-a-command` or `x-command-palette` or similar (but readable). Then apply the fix that you're suggesting but with that class instead of hardcoding the command name. It would be more correct semantically.

PR #6816, opened 2024-02-15
Change transport type when setting original system

### LINKED ISSUE #6726: Incorrect transport type after initial pull with "original system" set
"Works great with initial pull into a new system but releasing the transport request/task after gives an error becuase it's not of the correct type since now the objects should be in a repair transport becuase the original system has been changed after the objects were created. "

image

image

_Originally posted by @einareiriksson in https://github.com/abapGit/abapGit/issues/6715#issuecomment-1874316475_

mbtools:
Follow-up to #6715

After setting the original system at the end of deserializing objects, the type of the transport is now switched to "Repair" which prevents errors when releasing the transport.

Closes #6726

### DIFF
--- a/src/cts/zcl_abapgit_cts_api.clas.abap
+++ b/src/cts/zcl_abapgit_cts_api.clas.abap
@@ -82,7 +82,7 @@ ENDCLASS.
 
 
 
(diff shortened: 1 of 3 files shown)


larshp (approved):
lgtm

PR #6818, opened 2024-02-16
Fix bug in Personal settings page

francisco-milan:
This PR is to fix the bug reported in #6817.


francisco-milan:
Closes #6817

mbtools (approved):
thanks 👍

PR #6651, opened 2023-11-17
Repo List: Make labels clickable

mbtools:
Clicking a label in the list of repos filters the list accordingly

image

PR #6821, opened 2024-02-20
Remove duplicate labels from Command Palette.

francisco-milan:


mbtools:
The duplicate labels are a side effect of #6651. Your code assumes that same title means same action. It's true for these repo labels but not necessarily in general. For example, there could be two "View" items with different links or parameters. We should remove duplicates only if title and action match.

francisco-milan:
Dear abapGit Team,

Thank you for your prompt response and valuable feedback regarding the issue with duplicate entries in the Command Palette.

I appreciate the insights provided by @sbcgua and @mbtools regarding the underlying cause of the problem and the need for a more semantic solution. Upon reviewing your suggestions, I agree that a more comprehensive approach is necessary to address this issue effectively.

I acknowledge the importance of marking labels in table rows as unlisted, as suggested by @sbcgua. This approach aligns with the goal of ensuring a more robust and semantic solution to prevent the occurrence of duplicate entries in the Command Palette.

As I am relatively new to CSS, I will need to conduct some investigation to implement this solution effectively. Any guidance or assistance from your end in this regard would be greatly appreciated. In the meantime, I propose applying the code I provided as an immediate (temporary) solution to avoid seeing the duplicates in the Command Palette. The CSS alternative suggested by @sbcgua can then be applied later as a permanent solution.

I will work on implementing this solution and provide an updated pull request accordingly. Your patience and collaboration in resolving this issue are greatly appreciated.

Please feel free to reach out if you have any further questions or require additional information.

Thank you for your attention to this matter.

Best regards,

Francisco Milan

mbtools:
As suggested by @sbcgua in #6822, marking labels in the repo list as unlisted so they can be ignored by the command palette. 

Can you please test, @francisco-milan?

francisco-milan:
Hi Marc,

I've tested the changes you applied, and I'm happy to report that everything is working smoothly now. The issue with duplicates in the Command Palette has been resolved, and the labels in the repository list are no longer appearing as duplicates.

image

Thank you for implementing the suggested fix, and please let me know if there's anything else I can assist with.

Best regards,

Francisco Milan.

PR #6824, opened 2024-02-22
Fix Filter Application Bug

mkaesemann:
Moved the filtering of ignored items before the application of the filter when getting remote files to avoid exceptions that might occur if a repo contains ignored files that have no corresponding representation in the SAP system


mbtools (approved):
Lgtm

PR #6853, opened 2024-03-14
SICF: Include alternate service name

### LINKED ISSUE #6847: Different SICF paths for the same service in the source and target systems
**abapGit 1.127.0 - Developer Version**

I think this topic is related to the following issue: https://github.com/abapGit/abapGit/issues/4902

I have created some custom OData services in the source system. I would like to move a complete package that also contains these OData services to another system (both are S/4HANA).
Most of the objects from the package have been moved without any problems. However, I've noticed some problems related to these OData services:
1. Paths for these services in SICF are different after import, even though they are correct (not nested) in the `.sicf.xml` files. See the screenshots below for details.
Source system:
Source System Path
Source System URL (REMOTE XML in abapGit)
Target system:
Target System Path
Because of this, abapGit still shows me differences between the local state and the remote repository. However, when I try to pull again, I get a list of errors telling me that these services and some nodes already exist. So it sees differences that are there, but imports objects differently.
2. This is more of a suggestion than a serious issue. The service name does not change when you import SICF services. It still contains the name of the old system and the client at the beginning of the service name. I think this should be mapped so that it's not a mystery to SAP Basis how these names got into their system.

If you need any more information from me, I will update.

mbtools:
Symptom: 

Deserializing an SICF service leads to a diff showing a new object with a different filename compared to the repository.

Cause:

The service has an alternative name in lower case (`icfservice-icfaltnme_orig`). When deserializing the service, it is created with an upper-case alternate name.

Fix:

Included the lower-case alternative name in the serialized file in order to create the same service definition when deserializing the object.

Closes #6847

### DIFF
--- a/src/objects/zcl_abapgit_object_sicf.clas.abap
+++ b/src/objects/zcl_abapgit_object_sicf.clas.abap
@@ -669,7 +669,10 @@ CLASS zcl_abapgit_object_sicf IMPLEMENTATION.
     CLEAR ls_icfservice-icf_user.
     CLEAR ls_icfservice-icf_cclnt.
     CLEAR ls_icfservice-icf_mclnt.
-    CLEAR ls_icfservice-icfaltnme_orig.
+    " If the original name is different (lower vs upper case), it needs to be serialized
+    IF ls_icfservice-icfaltnme = ls_icfservice-icfaltnme_orig.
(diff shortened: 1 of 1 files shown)


mbtools:
Tested:

image

PR #6862, opened 2024-03-16
Code inspector pages improvements

sbcgua:
User-wise - visual and UX improvement of codi pages (code inspector and syntax check):
- better look in general
- table formatting
- sorting
- filtering
- err/warn counters
- the visualization code is plus-minus rewritten completely

image

Invisible internals:
- significant progress on html_table component:
  - sorting support and event helper
  - default css style
  - support for rendering data attributes - which helps with CSS in particular, but can be useful elsewhere as well (e.g. JS logic)
  - relevant unit tests
- html: `div` wrapper, `data_attr` helper added
- HOC: menu and title providers are takes from the components unless passed explicitely

Offtopic: minor fix of stage refresh placing. For good. Was interfering the changes in codi page.

TODO: will write some docs about table component (in doc repo)

P.S. I wanted also to add JS search like in stage. But I'm lazy. Next time. Maybe. Or maybe elsewhere.

### DIFF
--- a/src/ui/core/zcl_abapgit_html.clas.abap
+++ b/src/ui/core/zcl_abapgit_html.clas.abap
@@ -29,6 +29,11 @@ CLASS zcl_abapgit_html DEFINITION
         iv_checked     TYPE abap_bool OPTIONAL
       RETURNING
         VALUE(rv_html) TYPE string .
+    CLASS-METHODS data_attr
+      IMPORTING
+        iv_str         TYPE string OPTIONAL
+      RETURNING
+        VALUE(rs_data_attr) TYPE zif_abapgit_html=>ty_data_attr .
   PROTECTED SECTION.
   PRIVATE SECTION.
 
@@ -122,6 +127,16 @@ CLASS ZCL_ABAPGIT_HTML IMPLEMENTATION.
     IF iv_initial_chunk IS NOT INITIAL.
       ri_instance->add( iv_initial_chunk ).
     ENDIF.
+  ENDMETHOD.
+
+
+  METHOD data_attr.
+
+    SPLIT iv_str AT '=' INTO rs_data_attr-name rs_data_attr-value.
+    IF rs_data_attr-name IS INITIAL.
+      CLEAR rs_data_attr.
+    ENDIF.
+
   ENDMETHOD.
 
 
@@ -523,6 +538,7 @@ CLASS ZCL_ABAPGIT_HTML IMPLEMENTATION.
       ii_content = ii_content
       iv_id    = iv_id
       iv_class = iv_class
+      is_data_attr = is_data_attr
       iv_hint  = iv_hint ).
     ri_self = me.
   ENDMETHOD.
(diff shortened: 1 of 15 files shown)


larshp:
tested ok on 754,

image

src/ui/core/zcl_abapgit_html.clas.abap
```diff
@@ -29,6 +29,11 @@ CLASS zcl_abapgit_html DEFINITION
         iv_checked     TYPE abap_bool OPTIONAL
       RETURNING
         VALUE(rv_html) TYPE string .
+    CLASS-METHODS data_attr
```
larshp, line 32: suggest renaming this method to `parse_data_attr `

sbcgua: done

larshp (approved):
thanks, lgtm

sbcgua:
https://github.com/abapGit/docs.abapgit.org/pull/217

larshp:
@mbtools do you want to review this PR, or have any comments?

mbtools:
Looks great. The improved HTML Table will be useful in other places. Thanks, Alexander!

I'm not sure about moving the repo refresh into the staging page. For example, it will now refresh going from "Flow" to "Stage". Maybe add a `force_refresh` parameter to the constructor so there's a choice. 

PS: Some users have very large repos so we should avoid forced refreshes (https://github.com/abapGit/abapGit/issues/5137).

sbcgua:
> ... moving the repo refresh into the staging page ... @mbtools 

Yeah ... that was a controversial change. Let me explain. The stage page is called from: 1) codi 2) router 3) flow. In cases 1 and 2, refreshing the repo outside of stage page is a wrong semantics. Refresh is a part of staging business logic - I mean "before showing the stage page we want to ensure that the repo is syncronized" - this is because of "stage" and not because of the calling logic. So, calling refresh in 1 and 2 was just completely wrong. Missing refresh in 3 seemed a mistake to me, so I just ignored it. Now your comment clarified that this was by design - OK. At the same time, I'm not sure that stage should always refresh, I even think it should not. But it is an entirely separate discussion and I didn't want to change irrelevant logic, so I just changed "completely wrong" code to "tolerable" :))))

Anyway, I added the param to the constructor as you suggested. The default is `true`. The Flow passes explicit `false`. So now the final behavior should be like before the change.

sbcgua:
> The improved HTML Table will be useful in other places ... @mbtools 

Yes. Maybe we can start planning to use it elsewhere. But yet accurately, I have some more ideas in mind.

1) currently sorting logic is outside of the component. And I think it is OK. But, for simpler usages it would be nice to have it encapsulated, like in ALV. So maybe there is a room for some kind of "managed" mode, which also handles sorting itself. Maybe.
2) Also if I have time, I might write a unified JS filter.
3) And maybe html pull dialog ... I don't want to promise, but there are some annoying issues with the existing pull process.

mbtools:
All good 👍

larshp:
🚢

PR #6871, opened 2024-03-20
Fix Unexpected Functionality for "Maximum Length of Comment" Field

francisco-milan:
This PR fixes the issue documented in #6872

### DIFF
--- a/src/ui/lib/zcl_abapgit_html_form_utils.clas.abap
+++ b/src/ui/lib/zcl_abapgit_html_form_utils.clas.abap
@@ -289,7 +289,7 @@ CLASS zcl_abapgit_html_form_utils IMPLEMENTATION.
             CATCH cx_root.
               ro_validation_log->set(
                 iv_key = <ls_field>-name
-                iv_val = |Input Value Out of Range. Please enter a number with less than 32 digits| ).
+                iv_val = |Input Out of Range: Please enter a valid entry with fewer than 32 digits| ).
               CONTINUE.
           ENDTRY.
           IF <ls_field>-min <> cl_abap_math=>min_int4 AND lv_number < <ls_field>-min.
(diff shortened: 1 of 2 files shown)


src/ui/lib/zcl_abapgit_html_form_utils.clas.abap
```diff
@@ -247,7 +247,7 @@ CLASS zcl_abapgit_html_form_utils IMPLEMENTATION.
     DATA:
       lt_fields TYPE zif_abapgit_html_form=>ty_fields,
       lv_value  TYPE string,
-      lv_number TYPE i.
+      lv_number TYPE int8.
```
mbtools, line 250: won't work in 702

mbtools (changes requested):
50 is the recommended max for the title (https://www.git-scm.com/docs/git-commit#_discussion) so that is the default value for the setting.

Users should be able to pick other values. min = 1 is correct, but max should be limited to what fits into a single line. That's about 100 characters. Maybe someone likes higher values and has them already set in abapGit. Let's set max = 250. 

Example where the commit title "should fit":

image

francisco-milan:
Test results after implementing the requested changes:

**Scenario 1**. "Maximum Length of Comment" set to a value <= -10,000,000,000,000,000,000,000,000,000,000 (32 digits or more).
Here I'm proposing to change the original error message, so instead of "Maximum Length of Comment is not numeric" we can display something like: "Input Value Out of Range. Please enter a numeric value with less than 32 digits". 

image

**Scenario 2**. "Maximum Length of Comment" set to a value >= -9,999,999,999,999,999,999,999,999,999,999 AND < 1 (up to 31 digits).

image

**Scenario 3**. "Maximum Length of Comment" set to a value >= 1 AND <= 250.

image

**Scenario 4**. "Maximum Length of Comment" set to a value >= 251 AND <= 9,999,999,999,999,999,999,999,999,999,999 (up to 31 digits).

image

**Scenario 5**. "Maximum Length of Comment" set to a value >= 10,000,000,000,000,000,000,000,000,000,000 (32 digits or more). Also here I would like to use the revised version of the error message instead of the original ("Maximum Length of Comment is not numeric").

image

If you have any questions or comments, please let me know.

Best regards,

Francisco Milan.

francisco-milan:
Closes #6872

mbtools (approved):
👍

mbtools (approved):
👍

Issue #6878, opened 2024-03-22
Inaccurate Hint for "Maximum Length of Comment" Field


francisco-milan:
Hi abapGit Team,

I wanted to bring to your attention a side effect that I observed following the implementation of PR #6871. It concerns the hint displayed for the "Maximum Length of Comment" field. Previously, the hint indicated "At most 50 characters." However, as a result of the changes introduced in PR #6871 we need to update the text of the hint to reflect the new limit.

Current Hint: "At most 50 characters"

![image](https://github.com/abapGit/abapGit/assets/36721657/76366d00-3d68-464f-8ce2-fc3b5d4a8118)

Proposed Hint: "At most 255 characters"

![image](https://github.com/abapGit/abapGit/assets/36721657/1c2812f3-7f01-4739-b6cb-555c759b0171)

I have already created the PR with the necessary adjustments, and its number is #6877. If you have any questions or comments regarding this change, please don't hesitate to reach out.

Thank you for your attention to this matter.

Best regards,

Francisco Milan.

francisco-milan:
Dear abapGit Team,

I wanted to bring to your attention a recent change made in issue #6877 where the variable **lv_commitmsg_comment_length** was replaced with the constant **lc_commitmsg_comment_max_len**. I'd like to emphasize the importance of retaining both the constant and the variable. Specifically, the variable lv_commitmsg_comment_length is essential for dynamically adjusting the length of the "Comment" field on the "Commit" page. Currently, this field is fixed at 255 characters, which may not always align with our system preferences.

To illustrate the necessity of this adjustment, I changed the length of the "Comment" field to 15 characters as an example (see the 2 images below). However, without the restoration of the variable lv_commitmsg_comment_length, such adjustments would have no effect.

![image](https://github.com/abapGit/abapGit/assets/36721657/14040ce0-8f9a-41f6-9d19-ca5c52f4f35a)

![image](https://github.com/abapGit/abapGit/assets/36721657/ad506bd4-0228-45fc-a88e-348948be276f)

Therefore, I've initiated PR #6886 to restore the variable lv_commitmsg_comment_length. This restoration will ensure flexibility within abapGit settings, allowing us to tailor the "Comment" field length according to our specific requirements.

Your consideration of this request is greatly appreciated.

Best regards,

Francisco Milan.

PR #6886, opened 2024-03-30
Restore Variable lv_commitmsg_comment_length

francisco-milan:
Restoring variable lv_commitmsg_comment_length, see comments added to #6878.


larshp (approved):
thanks

PR #6892, opened 2024-04-02
Refactor: i18n filename logic 

### LINKED ISSUE #6865: Refactor: move i18n filename logic to `zcl_abapgit_filename_logic`
Deriving object from filename should be moved from `zcl_abapgit_objects_files` to `zcl_abapgit_filename_logic`:

https://github.com/abapGit/abapGit/blob/57ea8ba7fc50610b9ebfc413a0caadba0dc518d5/src/objects/core/zcl_abapgit_objects_files.clas.abap#L405

Ref https://github.com/abapGit/abapGit/pull/6830

mbtools:
Closes #6865

Move the logic for `i18n` files to `zcl_abapgit_filename_logic`. 

@albertmink could you please do a quick test that the interface translation logic still works properly after this change?

PS: As a follow-up step, I will replace `zcl_abapgit_objects_files=>get_i18n_properties_file` with `read_i18n_files` so that `po` and `properties` files are handled the same way.


albertmink:
Will test later this week

src/objects/texts/zcl_abapgit_properties_file.clas.abap
```diff
@@ -10,7 +10,14 @@ CLASS zcl_abapgit_properties_file DEFINITION
       c_properties_feature TYPE string VALUE 'TRANSL'.
 
     METHODS constructor
-      IMPORTING iv_lang TYPE laiso.
+      IMPORTING
+        iv_lang TYPE laiso.
+
+    METHODS parse
```
albertmink, line 16: AFAIK: `parse` is not used as logic is pushed down to CLAS object_intf

src/objects/core/zcl_abapgit_objects_files.clas.abap
```diff
@@ -397,22 +432,33 @@ CLASS zcl_abapgit_objects_files IMPLEMENTATION.
...
+        IMPORTING
+          ev_lang     = lv_lang
+          ev_ext      = lv_ext ).
 
       CASE lv_ext.
         WHEN 'po'.
           CREATE OBJECT lo_po EXPORTING iv_lang = lv_lang.
           lo_po->parse( <ls_file>-data ).
           APPEND lo_po TO rt_i18n_files.
+        WHEN 'properties'.
```
albertmink, line 458: See other comment
Screenshot 2024-04-04 at 10 12 36

albertmink (changes requested):
Thanks, I could pull/change/push the properties files.

There is some copy/paste from `po` logic that is not needed, as a lot of the logic is pushed to the objects itself.  Maybe later in development this is needed.

Overall it is very hard to read the diffs (maybe I shouldn't review the lines, but the functionality). Methods that are moved around and hence, not shown properly in the diffs.

mbtools:
I was interested in the functional test, just to confirm that I didn't mess anything up.

The next PR should then move the json-path logic to the `parse` method of `zcl_abapgit_properties_file` so you don't have to repeat it for every object type.

albertmink (approved):
LGTM

PR #6897, opened 2024-04-03
Where used page

sbcgua:
ref: #6868

## Main feature

where used command in Advanced options of the repo - find all external (to the repo package) usages of objects in this package (and it's sub-packages). It's useful for checking potential dumps in the code which uses package code after update deployments.

image

image

logic is implemented in `zcl_abapgit_where_used_tools` which is more or less copy of https://github.com/sbcgua/crossdeps/blob/master/src/zcl_dependency_model_wu.clas.abap

## Secondary improvements

- `html->render( iv_no_line_breaks )` - which does not apply indentation (can be useful sometimes, though generally a workaround)
- `html` allows specifying multiple `data-attrs`
- `chunk_lib->get_item_link` allows passing object directly (rather than repo item)
- `html_table` supports 2-level headers (the upper is called groups). Though without default styling - it's tricky. Consequently support `data-gid` in all relevant cells for styling. "Used in" shade demostrates the feature
- `html_table` supports `inline` sorting. It is ugly (imho) but convenient for simple tables (which are probably the majority) because does not need handling sorting in the higher component (which is also ugly). In general, I'm not super happy with the architecture but it's definitely far more convenient than building tables "manually" each time :) I think now it is OK to implement tables in other places (@mbtools), I'm personally thinking of the pool dialog and maybe repo-view (because it needs redesign) but no promises

Docs for new table features: https://github.com/abapGit/docs.abapgit.org/pull/225

### DIFF
--- a/src/inspect/zcl_abapgit_where_used_tools.clas.abap
+++ b/src/inspect/zcl_abapgit_where_used_tools.clas.abap
@@ -431,5 +431,12 @@ CLASS ZCL_ABAPGIT_WHERE_USED_TOOLS IMPLEMENTATION.
       ir_package_scope = lt_package_scope
       it_tadir         = lt_tadir ).
 
+    SORT rt_objs.
+    DELETE ADJACENT DUPLICATES FROM rt_objs.
+    " Duplicates happen e.g. because where-used is found by method.
+    " However here this functionality aggregates them to the object
+    " These are not true duplicates, so if ever the method name (or any other duplicate cause)
+    " will be extracted, this sort can be removed
+
   ENDMETHOD.
 ENDCLASS.
--- a/src/inspect/zcl_abapgit_where_used_tools.clas.abap
+++ b/src/inspect/zcl_abapgit_where_used_tools.clas.abap
@@ -1,0 +1,368 @@
+class ZCL_ABAPGIT_WHERE_USED_TOOLS definition
+  public
+  final
+  create public .
+
+  public section.
+
+    types ty_devc_range type range of tadir-devclass.
+    types:
+      begin of ty_dependency,
+        package type devclass,
+        obj_type type tadir-object,
+        obj_prog_type type trdir-subc,
+        obj_name type tadir-obj_name,
+        obj_cls  type rsfindlst-object_cls,
+        dep_package type devclass,
+        dep_obj_type type tadir-object,
+        dep_obj_name type tadir-obj_name,
+        dep_used_cls type rsfindlst-used_cls,
+        dep_used_obj type rsfindlst-used_obj,
+        cnt type i,
+      end of ty_dependency.
+    types:
+      tty_dependency type standard table of ty_dependency with default key.
+
+    " the initial version of this utility is also available as a standalone tool
+    " here: https://github.com/sbcgua/crossdeps
+    methods select_external_usages
+      importing
+        i_package        type tadir-devclass
+        ir_package_scope type ty_devc_range optional
+      returning
+        value(rt_objs) type tty_dependency
+      RAISING
+        zcx_abapgit_exception.
+
+  protected section.
+  private section.
+
+    types:
+      begin of ty_obj_signature,
+        package  type devclass,
+        obj_type type tadir-object,
+        obj_name type tadir-obj_name,
+      end of ty_obj_signature.
+
+    types tty_where_used type standard table of rsfindlst with default key.
+    types ty_seu_obj type standard table of seu_obj with default key.
+    types:
+      begin of ty_dev_object,
+        type type seu_stype,
+        tadir type trobjtype,
+      end of ty_dev_object.
+
+    data mt_object_packages type hashed table of ty_obj_signature with unique key obj_type obj_name.
+    data mt_dev_obj_cache type hashed table of ty_dev_object with unique key type.
+
+    methods get_where_used
+      importing
+        iv_obj_type      type euobj-id
+        iv_obj_name      type tadir-obj_name
+        it_scope         type ty_seu_obj optional
+        ir_package_scope type ty_devc_range optional
+      returning
+        value(rt_findings) type tty_where_used
+      RAISING
+        zcx_abapgit_exception.
+
+    methods get_obj_package
+      importing
+        iv_obj_type      type tadir-object
+        iv_obj_name      type tadir-obj_name
+      returning
+        value(rv_package) type tadir-devclass.
+
+    methods get_func_package
+      importing
+        iv_func_name      type tadir-obj_name
+      returning
+        value(rv_package) type tadir-devclass.
+
+    methods build_package_scope
+      importing
+        it_tadir type standard table
+        ir_package_scope type ty_devc_range
+      returning
+        value(rt_package_scope) type ty_devc_range.
+
+    methods collect_where_used
+      importing
+        it_tadir type standard table
+        ir_package_scope type ty_devc_range
+      returning
+        value(rt_objs) type tty_dependency
+      RAISING
+        zcx_abapgit_exception.
+
+    methods convert_list
+      importing
+        iv_package  type ty_dependency-dep_package
+        iv_obj_type type ty_dependency-dep_obj_type
+        iv_obj_name type ty_dependency-dep_obj_name
+        it_where_used type tty_where_used
+      returning
+        value(rt_objs) type tty_dependency.
+
+    methods decode_obj_type
+      importing
+        iv_type type rsfindlst-object_cls
+      returning
+        value(rv_type) type ty_dev_object-tadir.
+
+ENDCLASS.
+
+
+
+CLASS ZCL_ABAPGIT_WHERE_USED_TOOLS IMPLEMENTATION.
+
+
+  method build_package_scope.
+
+    field-symbols <tadir> type zif_abapgit_definitions=>ty_tadir.
+    field-symbols <pkg> like line of rt_package_scope.
+
+    rt_package_scope = ir_package_scope.
+    loop at it_tadir assigning <tadir>.
+      check <tadir>-object = 'DEVC'.
+      append initial line to rt_package_scope assigning <pkg>.
+      <pkg>-sign   = 'E'.
+      <pkg>-option = 'EQ'.
+      <pkg>-low    = <tadir>-obj_name.
+    endloop.
+
+  endmethod.
+
+
+  method collect_where_used.
+
+    data li_progress type ref to zif_abapgit_progress.
+    data lt_where_used type tty_where_used.
+    data lt_objs_portion like rt_objs.
+
+    field-symbols <tadir> type zif_abapgit_definitions=>ty_tadir.
+
+    li_progress = zcl_abapgit_progress=>get_instance( lines( it_tadir ) ).
+
+    loop at it_tadir assigning <tadir>.
+      check <tadir>-object <> 'DEVC'.
+
+      li_progress->show(
+        iv_current = sy-tabix
+        iv_text    = |{ <tadir>-object } { <tadir>-obj_name }| ).
+
+      lt_where_used = get_where_used(
+        iv_obj_type = |{ <tadir>-object }|
+        iv_obj_name = <tadir>-obj_name
+        ir_package_scope = ir_package_scope ).
+
+      lt_objs_portion = convert_list(
+        iv_package    = <tadir>-devclass
+        iv_obj_type   = <tadir>-object
+        iv_obj_name   = <tadir>-obj_name
+        it_where_used = lt_where_used ).
+
+      append lines of lt_objs_portion to rt_objs.
+
+    endloop.
+
+    li_progress->off( ).
+
+  endmethod.
+
+
+  method convert_list.
+
+    " See also CL_FINB_GN_BBI=>GET_CROSSREF
+
+    field-symbols <dep> like line of rt_objs.
+    field-symbols <use> like line of it_where_used.
+
+    loop at it_where_used assigning <use>.
+
+      append initial line to rt_objs assigning <dep>.
+      <dep>-cnt = 1.
+      <dep>-dep_package  = iv_package.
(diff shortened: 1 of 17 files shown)


src/inspect/zcl_abapgit_where_used_tools.clas.abap
```diff
@@ -0,0 +1,375 @@
...
+      END OF ty_dependency.
+    TYPES:
+      tty_dependency TYPE STANDARD TABLE OF ty_dependency WITH DEFAULT KEY.
+
+    CLASS-METHODS new
+      RETURNING
+        VALUE(ro_instance) TYPE REF TO zcl_abapgit_where_used_tools.
+
+    " the initial version of this utility is also available as a standalone tool
+    " here: https://github.com/sbcgua/crossdeps
```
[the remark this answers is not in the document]
larshp: no problem, its good to keep reference

src/inspect/zcl_abapgit_where_used_tools.clas.abap
```diff
@@ -0,0 +1,375 @@
...
+          IF <dep>-package IS NOT INITIAL.
+            <dep>-obj_type = 'INTF'.
+          ENDIF.
+        ENDIF.
+      ENDIF.
+
+      IF <dep>-package IS INITIAL.
+        <dep>-package = '????'.
+
+        IF <dep>-obj_type = 'PROG'. " Maybe it is an include
```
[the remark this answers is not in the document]
larshp: its a mess, trial and error is the way I'd approach it, ship it and there will be fixes over time

src/inspect/zcl_abapgit_where_used_tools.clas.abap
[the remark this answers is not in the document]
sbcgua: tnx

src/inspect/zcl_abapgit_where_used_tools.clas.abap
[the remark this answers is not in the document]
mbtools: This has a good coverage: [`zcl_abaplint_deps_find`](https://github.com/abaplint/abaplint-sci-client/blob/main/src/deps/zcl_abaplint_deps_find.clas.abap)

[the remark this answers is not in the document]
sbcgua: Aha, looks interesting, thank you. FUGR seems to be findable this way for the first glance. It is an improvement

pokrakam:
You can check `cl_wb_manager->if_wb_manager~request_tool_access` which leads to the actual object checks in `cl_wb_source_dispatcher->if_wb_decision_class~get_tool_for_request`.

sbcgua:
progress ...

image

src/ui/lib/zcl_abapgit_html_table.clas.testclasses.abap
```diff
@@ -226,4 +227,71 @@
...
+
+    li_html_exp->add(
+      '<table id="with-sort">' )->add(
+      '<thead>' )->add(
+      '<tr>' )->add(
+      '<th></th>' )->add(
+      '<th colspan="2">Group</th>' )->add(
+      '</tr>' )->add(
+
+      '<tr>' )->add(
```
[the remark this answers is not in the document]
sbcgua: Although it adds readability ...

image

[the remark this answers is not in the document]
mbtools: reads much better with `add` in the front

```abap
li_html->add( '<table>'
  )->add( '<thead>'
  )->add( '<tr>'
  )->add( '<th>hello</th>'
  )->add( '<th>world</th>'
  )->add( '</tr>' ).

```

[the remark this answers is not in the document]
sbcgua: Yes, this can be arranged actually, but another problem is max length of 120 symbols and that orphan `add` in the middle  does not fit.

Actually, I think this linter rule is mostly good ... but easy (in-code) exceptions would be so great ...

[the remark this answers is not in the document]
sbcgua: Ah, you mean to break chaining in the middle. I didn't think about it, a good idea actually, thanks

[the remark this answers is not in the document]
sbcgua: I wonder if abap will ever support multiline strings like in js ...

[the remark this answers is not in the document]
mbtools: Nah, multi line strings are confusing because you don't know if the line break is part of the string of not.

sbcgua:
Ready for review. Updated the description too.

P.S. except the linter errors for field symbols. @larshp what do you think about this and #6898

sbcgua:
linter comments fixed

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
```diff
@@ -0,0 +1,222 @@
...
+
+    mi_table = zcl_abapgit_html_table=>create(
+      ii_renderer              = me
+      is_initial_sorting_state = ls_sorting_state ).
+    mi_table->define_column_group(
+        iv_group_title = 'Repo object'
+        iv_group_id    = '' " No need
+      )->define_column(
+        iv_column_id    = 'dep_package'
+        iv_column_title = 'Pkg'
```
mbtools, line 96: No abbreviations necessary... "Package", "Type", "Name" (same as in [repo view](https://github.com/abapGit/abapGit/blob/ecdd17162a2c5e14d4feae99ef2f6f9b157db7b7/src/ui/pages/zcl_abapgit_gui_page_repo_view.clas.abap#L1040))

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
```diff
@@ -0,0 +1,222 @@
...
+    register_handlers( ).
+    init_table_component( ).
+
+    CREATE OBJECT ri_html TYPE zcl_abapgit_html.
+
+    ri_html->div(
+      iv_class   = 'wu-header'
+      iv_content = |Where used for package {
+        zcl_abapgit_gui_chunk_lib=>render_package_name( mv_package )->render( iv_no_line_breaks = abap_true )
+        } and it's subpackages| ).
```
mbtools, line 180: Depending on repo setting:

"Where-used of objects in package XYZ in other packages" 
or 
"Where-used of objects in package XYZ and its subpackages in other packages"

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
```diff
@@ -0,0 +1,222 @@
...
+
+    lt_where_used = zcl_abapgit_where_used_tools=>new( )->select_external_usages( mv_package ).
```
mbtools, line 182: Should depend on "Ignore sub-package" setting

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
```diff
@@ -0,0 +1,222 @@
...
+        iv_column_title = 'Pkg'
+      )->define_column(
+        iv_column_id    = 'dep_obj_type'
+        iv_column_title = 'Type'
+      )->define_column(
+        iv_column_id    = 'dep_obj_name'
+        iv_column_title = 'Obj name'
+      )->define_column(
+        iv_column_id    = 'dep_used_obj'
+        iv_column_title = 'Used obj' ).
```
mbtools, line 105: Maybe "Used Data Type" is better

mbtools (commented):
Some small things only

Does it list objects that are not used anywhere?

Overall, this needs a bit of docs to clarify what it does and does not find. 

AG supports a lot of object types. Which object types are supported by this feature?

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
sbcgua: This one ?

image

hmmm, that means requiring the full repo object in this page rather than just the package name ... Interesting. What 's the use of this option? I can imagine ignoring specific subpackages but all ... hmm. Let me think about it. There is something unnatural about this flag.

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
sbcgua: Hmm, are you sure there are only data types? (I'm just not sure exactly how it works, but in a much larger repo this field contains transactions too)

sbcgua:
> Does it list objects that are not used anywhere?

No

> AG supports a lot of object types. Which object types are supported by this feature?

All (that are supported by standard where-used or rather `RS_EU_CROSSREF` FM

> Overall, this needs a bit of docs to clarify what it does and does not find.

Yeah. looks like. But there is no too much to describe. I wonder if a (?) sign somewhere in the corner would be enough? Just to extend docs to the scale of a paragraph (from the current description). What do you think ?

E.g.

```
This tool cycles through all objects in the repo and runs the standard where-used function against it. The result is displayed here less the usages inside the repo itself. The tool can be used to detect potential regressions in the code which uses the repo and external to it (e.g. when deploying updates to a library-like repo).
```

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
sbcgua: And if sub-packages are omitted, should the usages of THE PACKAGE in THE SUBPACKAGES be displayed? Hmmm. very strange flag ... :)

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
mbtools: It looks like the ABAP-internal representation that you also see in debugger. I would go for "ABAP Type" as in the [docs](https://help.sap.com/doc/abapdocu_latest_index_htm/latest/en-US/index.htm?file=abentype_hierarchy.htm)

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
mbtools: The flag existed "forever". It's useful if you have one repo per package, no matter the package hierarchy or if you want to serialize a package in the middle of the hierarchy.

mbtools:
I wouldn't put a lot of text, but a link to the docs.abapgit.org is good. We have the "?" icon for dialog pages (as part of the form). Makes sense to have it also for regular pages. 

There's a few new things in the PR that might not exist in 702. I will check it Friday.

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
sbcgua: done

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
sbcgua: "forever" does not except that it is strange and suboptimal. The excluded code should be outside of the package or in a specific sub-package(s) but not "all" subpackages - this is a very weird and wrong case. This feature should be redesigned imho. But OK, it does is not applicable to this PR. I'll see how to fix it. It's just really not straight forward especially in terms of displaying connections with exclusions

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
sbcgua: image

I'd keep as is ...

sbcgua:
> There's a few new things in the PR that might not exist in 702. I will check it Friday.

Yeap, please check if possible. It works on 731.

sbcgua:
> ... a link to the docs.abapgit.org is good

Fine, but what a good place for this paragraph? It's too small for a dedicated page. Any ideas?

mbtools:
Could be a mouseover hint like we have in forms.

src/ui/pages/codi/zcl_abapgit_gui_page_whereused.clas.abap
mbtools: `lo_repo->get_tadir_objects( )` takes all settings into consideration and gives you the right objects

sbcgua:
all done

mbtools:
Good news: It works on 702!

I tried it with the AG repo and it took about 20 sec. 

image

A couple render issues:
- The table is wider than the AG frame (right side is sticking out beyond the "refresh" button)
- The type column headings wrap 

When sorting the columns, it does a complete refresh. Shouldn't it just re-render?

sbcgua:
overhang ... yeah, I've seen this too already on some real data ... the content is too long and unbreakable ... I'm thinking what to do with this, most probably will rearrange the "Used obj" thing somewhere to title or second line in object name. 

Sorting: good point. I'll check.

sbcgua:
done

sbcgua:
> Capitalize menu items

Not sure why it's needed but done

> Duplicates

Better to have a case. I didn't notice duplicates in my examples. Please check your ones under debugger, might be they are not duplicates - maybe different objects used or something, then they are useful and not duplicates

mbtools:
The duplicates occur when an object is used in multiple methods of a class:

image

You can see it, for example, if you install https://github.com/abapGit/CI and check the where-used of abapGit itself.

larshp:
I think, consider just shipping it, even though there are duplicates, open a separate issue for that. Makes it easy to contribute a fix, plus this is all isolated functionality

mbtools (approved):
Let's go! 👍

PR #6905, opened 2024-04-08
Change message if local packages got lost

mbtools:
Local packages might get lost during an upgrade (see #6784 for details). This changes the message raised by abapGit when serializing a repository based on such lost packages.

Before:

image

After:

image

The general solution is to simply pull the missing packages again from the remote.


larshp (approved):
lgtm

PR #6906, opened 2024-04-08
SOTS: Remove from TADIR selection

### LINKED ISSUE #4480: SOTS/SOTR: Refactor

mbtools:
All OTR longtexts (`SOTS`) are now serialized as part of the object they belong to (see where-used of `ZCL_ABAPGIT_SOTS_HANDLER`).

It is therefore not necessary anymore to serialize `SOTS` separately. The `SOTS` object type is now ignored by abapGit and corresponding files can be deleted from repositories (assuming you updated objects like enhancements and classes which contain the longtexts).

PS: Ignoring `SOTS` closes several issues with this object type (Closes #4480).


larshp (approved):
lgtm

Issue #6917, opened 2024-04-23
No option to set default git user details


pokrakam:
There is a default Git user (name + email) available in `zcl_abapgit_persistence_user` and in the config XML. 
I'd expect to see it in the Personal Settings config page, but can't find any option anywhere to set this. Not high priority.

mbtools:
AG stores the user/email of the last commit in these fields:

https://github.com/abapGit/abapGit/blob/a18b9c7f810a5ff3b904e2575583021ce851824b/src/ui/routing/zcl_abapgit_services_git.clas.abap#L61-L64

So it become the default for the *next* commit 😃

mbtools:
Might be something for @francisco-milan ?

francisco-milan:
Sure, you can assign this issue to me, I'll be glad to help.

mbtools:
Closed by #6920 (changing authentication popup is seperate issue/pr)

PR #6920, opened 2024-04-29
Set Default Git User Details 

francisco-milan:
PR to implement new functionality, as per #6917

### DIFF
--- a/src/ui/zcl_abapgit_password_dialog.clas.abap
+++ b/src/ui/zcl_abapgit_password_dialog.clas.abap
@@ -25,8 +25,11 @@ CLASS zcl_abapgit_password_dialog IMPLEMENTATION.
   METHOD popup.
 
     DATA: lx_error TYPE REF TO cx_sy_dyn_call_illegal_form.
-    DATA(default_git_uname) = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_uname( ).
-    DATA(default_git_email) = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_email( ).
+    DATA default_git_uname TYPE string.
+    DATA default_git_email TYPE string.
+
+    default_git_uname = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_uname( ).
+    default_git_email = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_email( ).
 
     IF default_git_uname IS NOT INITIAL.
       cv_user = default_git_uname.
--- a/src/ui/zcl_abapgit_password_dialog.clas.abap
+++ b/src/ui/zcl_abapgit_password_dialog.clas.abap
@@ -25,6 +25,14 @@ CLASS zcl_abapgit_password_dialog IMPLEMENTATION.
   METHOD popup.
 
     DATA: lx_error TYPE REF TO cx_sy_dyn_call_illegal_form.
+    DATA(default_git_uname) = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_uname( ).
+    DATA(default_git_email) = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_email( ).
+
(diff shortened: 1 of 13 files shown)


francisco-milan:
Closes #6917

larshp:
image

src/ui/zcl_abapgit_password_dialog.clas.abap
```diff
@@ -25,6 +25,17 @@ CLASS zcl_abapgit_password_dialog IMPLEMENTATION.
...
+    DATA lv_default_git_uname TYPE string.
+    DATA lv_default_git_email TYPE string.
+
+    lv_default_git_uname = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_uname( ).
+    lv_default_git_email = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_email( ).
+
+    IF lv_default_git_uname IS NOT INITIAL.
+      cv_user = lv_default_git_uname.
+    ELSEIF lv_default_git_email IS NOT INITIAL.
+      cv_user = lv_default_git_email.
```
larshp, line 37: this looks suspicious, both uname and email goes into `cv_user` ?

larshp: perhaps just leave the password popup as it, and have this PR surface the configuration settings only? there are various other exits for the authentication(which might be different than git users)

pokrakam: Looks to me that we just default the email address as user name if not supplied. Sounds sensible to me.

larshp: will the default override what the user has entered as username?

pokrakam: No, see line 34 just before your quoted lines
`IF lv_default_git_uname IS NOT INITIAL.`

larshp: yea, but that default then overrides the repo login? plus affects proxy authentication? I suggest leaving this PR to just surfacing the configuration in the settings page, and leaving the popup as is

image

pokrakam: Mmh, ok I get what you mean now. Agree the login is/should be completely separate to the name & email that is added to a commit.

larshp (changes requested):
please revert the password dialog changes, see comments above

src/ui/zcl_abapgit_password_dialog.clas.abap
mbtools: Isn't the point of the default to pre-fill the password popup? 

It should happen only if no user has been given yet. Default user having priority over default email is fine for me. 

```abap
  IF cv_user IS INITIAL.
    lv_default_git_uname = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_uname( ).
    lv_default_git_email = zcl_abapgit_persist_factory=>get_settings( )->read( )->get_default_git_email( ).

    IF lv_default_git_uname IS NOT INITIAL.
      cv_user = lv_default_git_uname.
    ELSEIF lv_default_git_email IS NOT INITIAL.
      cv_user = lv_default_git_email.
    ENDIF.
  ENDIF.
```

larshp: git is git, the popup is authentication, which already have settings and functionality

it might be okay to change the defaults, but I'd still suggest that for a different PR, so we can properly focus on reviewing it, the authentication is central functionality. Surfacing the settings is low criticality, that should be easy to merge

francisco-milan: Hi @larshp,

    The changes have been reverted, as per your request. Let me know in case you want me to create a new PR to pre-fill the password popup. 

Best regards,

Francisco Milan.

larshp (approved):
thanks

PR #6926, opened 2024-05-10
Suppress popup when using background API

### LINKED ISSUE #6925: Popup when pulling using background API
When pulling using `zcl_abapgit_background_pull`, there is still a popup to confirm activation of objects. 
This goes against the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Workaround fix underway.

pokrakam:
Closes #6925

Slightly workaround-y solution, but better than passing a flag or inspecting the call stack in https://github.com/abapGit/abapGit/blob/29baa5a15274518fd1dfb55b97029445f2e18bcb/src/objects/core/zcl_abapgit_objects_activation.clas.abap#L276


mbtools (approved):
Thanks! 👍

PR #6946, opened 2024-06-12
Fix yank mode - copy to clipboad

christianguenter2:
regression from #6944

`ii_event->query( )->get( 'CLIPBOARD' ).` 
vs 
`ii_event->form_data( )->get( 'CLIPBOARD' ).`


larshp (approved):

mbtools:
Thanks!

PR #6958, opened 2024-06-23
Read-only support for SXSD - Classic BAdI definition

pokrakam:
Read-only as we can only create the newer Enhancement Spots in customer systems. Deserialiser will issue appropriate error.

### DIFF
--- a/src/objects/zcl_abapgit_object_sxsd.clas.abap
+++ b/src/objects/zcl_abapgit_object_sxsd.clas.abap
@@ -63,7 +63,7 @@ CLASS zcl_abapgit_object_sxsd IMPLEMENTATION.
   ENDMETHOD.
 
   METHOD zif_abapgit_object~exists.
-    rv_bool = xsdbool( ms_badi_attr IS INITIAL ).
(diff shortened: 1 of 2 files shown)


fabianlupa:
Are you sure about not being able to create them? I am pretty sure I created one in a partner namespace on 7.55 about 1.5 years ago.

fabianlupa:
Just checked, you need to press this fancy hidden button instead of the normal create button in SE18 ;)

image

pokrakam:
Good find. I was just going by the message, which I checked and was there since 7.00
image

Not sure when I'll have time to do the deserialisation, and given that we shouldn't be creating Classic BADIs since <7.02 I'd suggest merging this for now. Unfortunately since it's read-only, I can't set up a test.

larshp (approved):
Suggest merging, if someone needs deserialization it can be done in a different PR

pokrakam:
FWIW, I tried to see how much effort the creation would be and found that the "Create Classic BADI" functionality seems to be broken on my A4H 2022 system. It generates a half-baked interface that kindof exists, until you try to use it then get the message it doesn't exist. 
I didn't pursue it any further, might have another look on another system when I find time.

PR #6960, opened 2024-06-28
TABL: Clear DBINCLEXCL and DBSYSSELx fields

### LINKED ISSUE #6959: DB Table index include/exclude flag: clear if no DBs specified
There is one typical diff I experience during deployments between different releases. `DD12V-DBINCLEXCL`. I have a suspicion that it is a bug in older abap releases. Although it does not cause direct issues but in AG it causes diffs. I'm not a big expert in how abap objects are stored so before fixing this would be great to collect some opinions for more experienced.

So the evidences are:

1) a table has in an index. Not limited to any DB

image

2) Just in case, if I click "For selected" then, inside there is a radio button with "Exclude" option by default. **Yet, the activated setting is "Index on ALL"**

image

3) And in DD12V the field DBINCLEXCL = 'E' (although it does not do anything). Problem is that in newer releases this field is force cleared on activation (which is more consistent)

image

Solution idea: on serialization, if DBSTATE is empty, clear DBINCLEXCL 

image

sbcgua:
closes #6959

image

### DIFF
--- a/src/objects/tabl/zcl_abapgit_object_tabl.clas.abap
+++ b/src/objects/tabl/zcl_abapgit_object_tabl.clas.abap
@@ -84,7 +84,7 @@ ENDCLASS.
 
 
 
-CLASS zcl_abapgit_object_tabl IMPLEMENTATION.
+CLASS ZCL_ABAPGIT_OBJECT_TABL IMPLEMENTATION.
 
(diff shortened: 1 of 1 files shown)


mbtools (approved):
👍

PR #6961, opened 2024-07-01
SXSD: Fix check if BADI exists

mbtools:
- Fix UT in `zcl_abapgit_objects`, bug in #6958
- Fix spacing between methods


larshp (approved):
lgtm

PR #6966, opened 2024-07-02
CDBO: Fix missing parameters in `constructor`

pokrakam:
Optional parameters not implemented can cause cx_.. exception or dump if called with files object. 
Also did some minor PP/tidying.


mbtools (approved):
Thanks 👍

PR #6978, opened 2024-07-10
Clear variables when getting `i18n` files

albertmink:
Not sure how you prefer:
(a) clear variables before passing to method
(b) clear variables inside METH

This PR implements (b).

Problem:
A non-cleared variable `lv_ext`
https://github.com/abapGit/abapGit/blob/1269360946d0625b2b671bc3171b01d0d6a96c9d/src/objects/core/zcl_abapgit_objects_files.clas.abap#L407


albertmink:
Contributes to #6980

mbtools:
I prefer (b). But in this case, the `clear` should be inside of `zcl_abapgit_filename_logic=>i18n_file_to_object` 😉

wurzka (approved):
LGTM :)

PR #6979, opened 2024-07-10
Simplify the JSON2ABAP

albertmink:
No need to call a fully customized AJSON handler. Just a plain AJSON with snake_case.

Contributes to #6980


mbtools (approved):
👍

PR #6981, opened 2024-07-10
Allow staging for empty repositories

### LINKED ISSUE #6968: Small Usability problem when creating new online repo
When I create a new online repo and I have a already stuff in the package, I arrive on the page with a list of the local objects but have no stage link/button there. I have to go back to the repo list and need to use the stage/button there for the initial stage/commit/push.

Current standalone version

mbtools:
Makes staging available if a repo contains zero files.

Closes #6968


larshp (approved):
lgtm

Issue #6980, opened 2024-07-10
[AFF] Get INTF with translation pulled


albertmink:
Issue that tracks effort of pulling an INTF object in ABAP file format with translation.

Test repository
https://github.com/albertmink/testing_intf_aff

Need experimental features: AFF, TRANS

albertmink:
Current state after pull:
abapGit tells that translation en-GB is not in the system
<img width="846" alt="Screenshot 2024-07-23 at 10 12 06" src="https://github.com/user-attachments/assets/3d7d75a6-2713-4a5c-8a71-faaf43a65fd8">

However, data is in the table
<img width="514" alt="Screenshot 2024-07-23 at 10 11 05" src="https://github.com/user-attachments/assets/c07ead81-eaa3-4a11-8bf8-715379c5afba">

Need work on the serialization part, I assume.

albertmink:
Pulling objects incl. translation finished.

PR #6986, opened 2024-07-11
Extend logic to detect BCP47 languages

albertmink:
Now, the logic can understand BCP47 languages.

After: #6978 
Ref #6980

### DIFF
--- a/src/objects/core/zcl_abapgit_filename_logic.clas.abap
+++ b/src/objects/core/zcl_abapgit_filename_logic.clas.abap
@@ -183,6 +183,12 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
   METHOD i18n_file_to_object.
 
     DATA lo_dot TYPE REF TO zcl_abapgit_dot_abapgit.
+    DATA lt_filename_elements TYPE string_table.
+    DATA lv_index TYPE syst_tabix.
+    DATA lv_langu_bcp47 TYPE string.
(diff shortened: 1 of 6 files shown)


larshp:
in https://github.com/abapGit/abapGit/blob/main/src/objects/core/zcl_abapgit_objects_files.clas.testclasses.abap#L57 the language "cz" is not valid, try changing it to "cs"

larshp:
also fails in system, 
image

albertmink:
Need help.
From my understanding the error is in `zcl_abapgit_convert=>language_bcp47_to_sap1`. However, afaik the unit test in the PR calls the METH with 'en' and 'de'. Both are already covered by unit tests of the `zcl_abapgit_convert`, so why does it fail here? What am I missing?
Screenshot 2024-07-16 at 09 59 17

larshp:
is all unit tests running okay in a real system? then I'll have a look at open-abap

albertmink:
Yes indeed.

larshp:
I still get these errors when running on https://hub.docker.com/r/sapse/abap-cloud-developer-trial, 

image

larshp:
try changing "cz" to "cs", https://github.com/abapGit/abapGit/pull/6986#issuecomment-2227297986

image

larshp (approved):
lets go

albertmink:
Wait, I'd like to add a meaningful UT. Maybe this works also ABAP standalone ;)

Issue #6985, opened 2024-07-11
Q: How shall `zcl_abapgit_convert=>language_sap2_to_sap1` work?


albertmink:
The METH `zcl_abapgit_convert=>language_sap2_to_sap1` doesn't convert `6N` (aka `en-GB` aka `둮`) as it selects form `I_LANGUAGE`, which in my system contains entries
<img width="341" alt="Screenshot 2024-07-11 at 13 04 15" src="https://github.com/abapGit/abapGit/assets/38354196/1a7955f5-a0a8-485d-b4f7-601a5c6bfecb">
Method even raises and does not call statement of Line 325, which would returns the  expected value `둮`
<img width="381" alt="Screenshot 2024-07-11 at 13 16 04" src="https://github.com/abapGit/abapGit/assets/38354196/ef9ed744-ce83-42fd-8851-7ed30274eb3c">

Why is that and is the current behaviour right?

mbtools:
The first part is intended for ABAP Cloud, where cl_i18n_languages is not released. So the answer needs to come from SAP how to handle the missing languages in cloud.

albertmink:
Ok, what I understand is that abapGit cannot convert languages in ABAP Cloud based on `cl_i18n_languages`. I learned that
```ABAP
data(converter) = xco_cp_language=>format->iso_639.
data(spras) = converter->to_language( 'EN' )->value. " is 'E'
```
would be an ABAP Cloud alternative. However, it raises when converting non-installed languages, e.g. `6N`.

So my question is: What is the expected behaviour in abapGit (in ABAP Cloud scenario) when converting non-installed languages?

mbtools:
The result of importing objects with AG should be the same as importing them via transport. 

AFAIK, a transport with texts in several languages would import all of them no matter if the language is installed in the system or not. 

If AG ignored some languages that are not installed, there would be diffs as well. Assuming a complete repo is translated that would mean diffs on all objects. 

It's best to import all languages. But then we need some kind of fallback for that ABAP cloud conversion method.

albertmink:
Thanks for clarification.

albertmink:
Maybe provide a table within abapGit as for bcp47 introduced in #6915
Supported Languages [Supportedlanguages.xls](https://github.com/user-attachments/files/16334608/Supportedlanguages.xls) as of SAP Note 73606

mbtools:
[...]

PR #6995, opened 2024-07-23
Filenames for translation are case sensitive

albertmink:


albertmink:
Tests in dev system where I_LANGUAGE contains 6N are green
Screenshot 2024-07-23 at 16 35 57

ref #6985

wurzka (approved):
Looks good to me 👍

mbtools (approved):
👍

PR #6999, opened 2024-07-25
fix: syst_langu does not exist on older releases

### LINKED ISSUE #6998: syst_langu does not exist on older releases
https://github.com/abapGit/abapGit/blob/a120ce66b0343d84f7a48bdc7d9f5969572b1125/src/objects/zcl_abapgit_object_intf.clas.locals_imp.abap#L710

AndreaBorgia-Abo:


mbtools (approved):
👍 I added `syst_langu` to a rule so it doesn't happen again

PR #7013, opened 2024-08-07
DSFI: Updated object handler to match the latest AFF definition

fvalves:
As part of the discussion to set up the ABAP File Format of object type DSFI, the ACDS file has been dropped. The object handler must be updated to match this decision.


mbtools (approved):
LGTM 👍

PR #7028, opened 2024-08-26
AFF names translation files `texts`

albertmink:
Specification and examples, see https://github.com/SAP/abap-file-formats/blob/main/docs/specification.md#language

### DIFF
--- a/src/objects/core/zcl_abapgit_filename_logic.clas.abap
+++ b/src/objects/core/zcl_abapgit_filename_logic.clas.abap
@@ -370,16 +370,23 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
 
 
   METHOD object_to_i18n_file.
-    DATA: lv_langu_sap1 TYPE sy-langu,
+    DATA: lv_langu_sap1  TYPE sy-langu,
           lv_langu_bcp47 TYPE string.
 
     lv_langu_sap1 = zcl_abapgit_convert=>language_sap2_to_sap1( to_upper( iv_lang ) ).
     lv_langu_bcp47 = zcl_abapgit_convert=>language_sap1_to_bcp47( lv_langu_sap1 ).
 
-    rv_filename = object_to_file(
-      is_item  = is_item
-      iv_extra = |i18n.{ lv_langu_bcp47 }|
-      iv_ext   = iv_ext ).
+    IF iv_ext = `properties`.
+      rv_filename = object_to_file(
+        is_item  = is_item
+        iv_extra = |texts.{ lv_langu_bcp47 }|
+        iv_ext   = iv_ext ).
+    ELSE.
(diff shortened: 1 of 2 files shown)


src/objects/core/zcl_abapgit_filename_logic.clas.abap
```diff
@@ -370,16 +370,23 @@ CLASS zcl_abapgit_filename_logic IMPLEMENTATION.
...
     lv_langu_bcp47 = zcl_abapgit_convert=>language_sap1_to_bcp47( lv_langu_sap1 ).
 
-    rv_filename = object_to_file(
-      is_item  = is_item
-      iv_extra = |i18n.{ lv_langu_bcp47 }|
-      iv_ext   = iv_ext ).
+    IF iv_ext = `properties`.
+      rv_filename = object_to_file(
+        is_item  = is_item
+        iv_extra = |texts.{ lv_langu_bcp47 }|
```
wurzka, line 382: Is this correct? [Here](https://github.com/SAP/abap-file-formats/blob/main/docs/properties.md), we specified that the file shall be named `<object_name>.<object_type>.i18n.<language>.properties`.

Texts is only for text elements of classes.

src/objects/core/zcl_abapgit_filename_logic.clas.testclasses.abap
```diff
@@ -505,7 +505,7 @@ CLASS ltcl_run_checks IMPLEMENTATION.
       iv_ext  = 'properties' ).
 
     cl_abap_unit_assert=>assert_equals(
-      exp = 'zif_abap.intf.i18n.en-GB.properties'
+      exp = 'zif_abap.intf.texts.en-GB.properties'
```
wurzka, line 508: I think interfaces don't even have text elements according to https://github.com/SAP/abap-file-formats/tree/main/file-formats/intf. So I would have expected that `zif_abap.intf.i18n.en-GB.properties` is correct?!

wurzka (commented):
I am confused with `texts` and `i18n`.
We have specified the filename with `i18n` for translation files, see https://github.com/SAP/abap-file-formats/blob/main/docs/properties.md.
However, `texts` is reserved for text elements of classes: https://github.com/SAP/abap-file-formats/tree/main/file-formats/clas

mbtools:
Aha, I'm confused, too. AG puts all main language texts in the object definition file (XML) and only separates translations into a separate file (with the experimental LXE/po files). 

Looks like AFF is doing it differently now. The JSON already includes a text i.e. the description of the object. Why do you need another file just for text elements? Can't these become part of the JSON?

albertmink:
🙈 ahhhh my bad. I got confused by our own documentation. Yeah, translation is `i18n` and texts/textpools are `texts`. So the PR is obsolete.

👉 closing PR

PR #7031, opened 2024-08-29
Comparing GET TIME STAMP and sy-datum and sy-uzeit only works if the user is in UTC

fvalves:


mbtools (changes requested):
The test works just fine on my system with `EST` timezone (timestamp matched `sy-datum/uzeit`):

image

But it fails with the suggested change:

image

Please check under exactly which conditions the test needs to be changed.

fvalves:
Thanks for the feedback, @mbtools ! I'll do more research to identify what causes the difference in behavior.

mbtools (approved):
Works now. Thanks!

Issue #6975, opened 2024-07-09
Combining packages with different Software Components results in package properties being changed 


AriannaMussoBarcucci:
Hello, I have encountered a confusing behavior while using the latest version of the zabapgit_standalone report when combining subpackages with different software components and ABAP language versions.

In my initial system, I have a structure package called Z_TEST_SUPERPACKAGE with Software Component = ZLOCAL and language = ABAP for Cloud Development

Inside this top structure package I created two development packages:

- Z_TEST_TIER1, with Software Component = ZLOCAL and language = ABAP for Cloud Development
- $Z_TEST_TIER2, with Software Component = LOCAL and language = ABAP standard

Via report ZABAPGIT_STANDALONE I then connected the Z_TEST_SUPERPACKAGE to a github repo, and pushed all the content.
I then pulled the content of the repo in a different system.
 
And here is the interesting behavior: the first package, which has same name prefix, same Software Component and same ABAP language as the structure package, is pulled in correctly.
**The second package (which has a different name suffix , different Software Component and different ABAP language) gets modified**: when pulled into the new system, the name is changed from $Z_TEST_TIER2 to Z_TEST_TIER2, the Software Component is changed from LOCAL to ZLOCAL, and the ABAP language is now blank (and when clicking on the dropdown menu in ABAP Developer Tools from Eclipse, the only available option is now ABAP language for cloud development). 
 
Is it possible that the zabapgit_standalone report reads the name prefix and software component from the top package, and then assigns it to all the subpackages, effectively changing the original Software Component?

mbtools:
In general, a scenario combining transportable and non-transportable packages ($-packages) in one repo is not supported and should have led to an error when creating the repo in abapGit. It looks like the package checks aren't strict enough.

I would consider such a combination in one repo not good practice anyway. Create separate repos for Z_TEST_TIER1 and $Z_TEST_TIER2.





All packages must have the same

AriannaMussoBarcucci:
Hi Marc,

Thank you for the quick reply!
Ok, this answer my question, I will take your recommendation and use two separate repos for the two packages.

PR #7037, opened 2024-09-03
Add package check when creating repository

mbtools:
Trying to create a repository for a package hierarchy that contain a mix of transportable and non-transportable packages is not supported and now leads to an error message.

Ref https://github.com/abapGit/abapGit/issues/6975#issuecomment-2218268991


larshp (approved):
lgtm

PR #7048, opened 2024-09-12
Enable CDS types

fvalves:
CDS types, with object type DRTY, had been previously implemented, but they had not yet been properly flagged as data dictionary objects.


mbtools (approved):
Thanks

PS: We need to run our [ci](https://ci.abapgit.org) more frequently and on a newer release to catch such issues 😉

PR #7126, opened 2025-01-10
Patch: fix 'invalid filename' error if filename is empty

christianguenter2:
If an object is erroneously (root cause unknown) detected as changed locally like this
image

Patching is not possible and raises `invalid filename` error from here
https://github.com/abapGit/abapGit/blob/3daf4d3f5ccdbd62d505ad39d22cd9cd6e61cbef/src/ui/pages/zcl_abapgit_gui_page_patch.clas.abap#L413
image

image

image

This change mitigates the error and patching is possible again
image


mbtools:
The state is based on sha1 comparison. I'm certain there's a diff. The problem is the SAP function used to calc the diff lines. I already included a `compute_diff_extra` in [`zcl_abapgit_diff`](https://github.com/abapGit/abapGit/blob/main/src/utils/zcl_abapgit_diff.clas.abap) that handles some cases. It looks like there's more and it would be good to find out what's missing. 🤷

mbtools (approved):
👍

christianguenter2:
@mbtools 
You're right. I just checked the object in `COMPUTE_DIFF_EXTRA` and did a diff in the debugger. There are some deltas. Seems related to tabs/whitespace.
image

image

image

mbtools:
`RS_CMP_COMPUTE_DELTA` converts tabs to 2 spaces... 😠 

image

mbtools:
These replacements impact `lt_delta` which is the source for the display of the diffs. This means there are no tabs to visualize. Usually they would show when selecting "View > Show Hidden Characters".

mbtools (approved):

PR #7132, opened 2025-01-16
Exit: Enhance any toolbar

### LINKED ISSUE #7121: Add User Exits to allow custom actions in repo toolbars
Similar to #6249, having user exits in place to modify the toolbars to add custom actions would be nice. For example, having a button that launches a custom tool for a given company.

I think it would be as simple as adding code to the  ZCL_ABAPGIT_GUI_PAGE_REPO_VIEW->BUILD_MAIN_TOOLBAR and BUILD_ADVANCED_TOOLBAR methods, allowing end users to adjust the toolbar objects.

Users would also need to also implement code in the ON_EVENT user-exit to handle these new buttons.

mbtools:
This exit allows you to add your own actions to any of the toolbars in abapGit. It's also possible to rename or remove items using the exit. Each toolbar has a unique id that you can use to identify it in the exit.

Closes #7121

### DIFF
--- a/src/exits/zcl_abapgit_exit.clas.abap
+++ b/src/exits/zcl_abapgit_exit.clas.abap
@@ -328,6 +328,18 @@ CLASS zcl_abapgit_exit IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD zif_abapgit_exit~enhance_any_toolbar.
+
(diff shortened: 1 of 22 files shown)


larshp (approved):
lgtm

PR #7134, opened 2025-01-17
Enable ABAP Language Version support for DDLS

fabianlupa:
Only for systems where baseinfo is supported and only for systems where the field is called `ABAP_LANGUAGE_VERSION`. I assume implementing fallbacks for each isn't worth it because the lower releases don't have ABAP Language Version support themselves.

### DIFF
--- a/src/objects/zcl_abapgit_object_ddls.clas.abap
+++ b/src/objects/zcl_abapgit_object_ddls.clas.abap
@@ -276,11 +276,12 @@ CLASS zcl_abapgit_object_ddls IMPLEMENTATION.
       lx_error         TYPE REF TO cx_root.
 
     FIELD-SYMBOLS:
-      <lg_data>             TYPE any,
-      <lg_data_baseinfo>    TYPE any,
-      <lg_source>           TYPE any,
-      <lg_baseinfo_string>  TYPE any,
-      <lg_baseinfo_ddlname> TYPE any.
+      <lg_data>                  TYPE any,
+      <lg_data_baseinfo>         TYPE any,
+      <lg_source>                TYPE any,
+      <lg_baseinfo_string>       TYPE any,
+      <lg_baseinfo_ddlname>      TYPE any,
+      <lg_abap_language_version> TYPE any.
 
     TRY.
         CREATE DATA lr_data TYPE ('DDDDLSRCV').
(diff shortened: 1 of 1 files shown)


fabianlupa:
This change seems to be compatible with regards to the serialized objects as "Standard ABAP" with value "X" is not stored in the table but standard ABAP has " " which means the initial value will not be added to the XML.

fabianlupa:
I guess we might need some [developer documentation](https://docs.abapgit.org/development-guide/serializers/serializers.html) for this. Some assumptions I am making:

1. On default settings the ABAP Language Version should be serialized in every object of the repo individually and can be set for each object individually. Just by chance Standard ABAP has an initial value and therefore does not show up in the XMLs (backwards compatible).
2. The experimental feature settings make it so it is never serialized even with values differing from Standard ABAP or to make the value be inherited from a language level set at repo level.
3. SAP will implement the same in their fork used in SAP BTP ABAP Environment and SAP S/4HANA Public Cloud so these object handlers are compatible and there are no diffs between the environments?

fabianlupa:
I created similar PRs for DDLX and DCLS. I also need BDEF, DOMA, DTEL, G4BA, SRVB, SRVD, TABL and TOBJ but would wait until the implementation here is ok. What does "planned" mean in the supported object type list? https://docs.abapgit.org/user-guide/reference/supported.html

mbtools:
It's been a while but it looks like you got it right. There's plenty in these docs: https://docs.abapgit.org/user-guide/reference/abap-language-version.html#supported-object-types but we could certainly add a bit to the dev docs (which are also missing AFF). 

The intention is certainly compatibility with BTP ie Eclipse abapGit but that's up to SAP as we know. 

"Planned" means I identified these object types as ones that support ALaV and therefore need to be included in abapGit this way. We were just waiting for helpful hands like yours 👍 

Test repos and CI should be enhanced too so we can test these scenarios more easily... I don't need any of it so prio low from my end.

fabianlupa:
> The intention is certainly compatibility with BTP ie Eclipse abapGit but that's up to SAP as we know.

Doesn't look like it at the moment.

https://github.com/fabianlupa/alav-cloud-test/blob/c0f3c23930b653abf74d13e9adaf459bab7eda83/src/zcl_fl_test_git.clas.xml#L12

CLAS in BTP ABAP Environment comes out with a hardcoded `UNICODE = X`, meaning language version "Standard ABAP", which is not the language version of the object in the system. I dared having a look at the object handler in the fork which does have logic to set it to `5` but it's commented out.... Anyone we can ping about that?

mbtools:
For source code object types, standard is X. For others its initial (see `zcl_abapgit_abap_lang_vers=> get_default_abap_language_vers).

mbtools:
For the BTP abapGit bug, try going via Michael Schneider or Albert Mink. They should know who's in charge of that.

fabianlupa:
My use case currently is mainly to have the strict syntax checks apply on premise. And to not lose them by pulling and getting reset to Standard ABAP. For cloning into cloud systems I'll probably have to setup a second branch anyway for cloud specific objects types (iam objects for example). The current behavior without anything set or experimental enabled should work fine for that I think.

PR #7140, opened 2025-02-03
Execute WDCC after import method

fdupret:
When importing a new Fiori catalog, it cannot be found in `/UI2/FLPD_CONF` until some cache refresh is triggered.
This cache refresh is performed automatically when importing a transport request containing a Fiori catalog (see table `OBJM` for object type `WDCC`). This pull request intends to mimic this behavior.

Solution inspired by `zcl_abapgit_object_chdo` and `zcl_abapgit_objects_generic`.

### DIFF
--- a/src/objects/zcl_abapgit_object_wdcc.clas.abap
+++ b/src/objects/zcl_abapgit_object_wdcc.clas.abap
@@ -8,6 +8,9 @@ CLASS zcl_abapgit_object_wdcc DEFINITION
 
     INTERFACES zif_abapgit_object .
   PROTECTED SECTION.
+    METHODS after_import
+      RAISING
+        zcx_abapgit_exception .
   PRIVATE SECTION.
 ENDCLASS.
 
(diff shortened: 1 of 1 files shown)


src/objects/zcl_abapgit_object_wdcc.clas.abap
```diff
@@ -230,10 +233,33 @@ CLASS zcl_abapgit_object_wdcc IMPLEMENTATION.
...
+    ls_cts_object_entry-obj_name = ms_item-obj_name.
+    INSERT ls_cts_object_entry INTO TABLE lt_cts_object_entry.
+
+    CALL FUNCTION 'WDR_CFG_AFTER_IMPORT'
+      EXPORTING
+        iv_tarclient  = sy-mandt
+        iv_is_upgrade = abap_false
+      TABLES
+        tt_e071       = lt_cts_object_entry
+        tt_e071k      = lt_cts_key.
```
mbtools, line 259: unfortunately, error messages are not exported. not sure if it's worth trying to import them from memory with `IMPORT LT_LOG FROM MEMORY ID 'APPEND_LOG'` 🤷

fdupret: I looked into it and I'm not sure it's worth the effort:

1. Passing incorrect values to the function module will yield in a silent fail. 
2. The logs are not reset between each call, resulting in duplicate entries in abapgit log:
image

Let me know what you think.
WIP here: https://github.com/fdupret/abapGit/tree/wdcc-serialize-with-log

mbtools: not worth it 👍

PR #7144, opened 2025-02-05
Fix saving inconsistent default transport

mbtools:
Avoid issue when abapGit saves inconsistent default transport. Also removes exception from constructor.


larshp (approved):
lgtm

PR #7150, opened 2025-02-18
Staging issue when folder name is in uppercase 

fidley:
I have an older repo, where the code is stored in the folder /CODE/ on github. Because of adaptation in the code of abapgit I cannot stage anymore changes to repo. I don't want to change now the folder to lowercase as, in this case it will show to all people that all objects were deleted and new objects were created (instead of showing the changes only). 

image

The 3 small changes made it work for me.

### DIFF
--- a/src/utils/zcl_abapgit_path.clas.abap
+++ b/src/utils/zcl_abapgit_path.clas.abap
@@ -27,6 +27,8 @@ CLASS zcl_abapgit_path DEFINITION
       IMPORTING iv_path            TYPE string
       RETURNING VALUE(rv_filename) TYPE string.
 
+  PROTECTED SECTION.
+  PRIVATE SECTION.
 ENDCLASS.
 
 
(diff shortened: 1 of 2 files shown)


mbtools:
Can you please add a unit test ?

fidley:
I can, but it will take a while. I start today holidays and I would need to install the contributor version for it.

src/utils/zcl_abapgit_path.clas.abap
```diff
@@ -125,7 +125,7 @@ CLASS ZCL_ABAPGIT_PATH IMPLEMENTATION.
       ev_filename = iv_fullpath.
     ENDIF.
 
-    ev_filename = cl_http_utility=>unescape_url( escaped = ev_filename ).
+    ev_filename = cl_http_utility=>unescape_url( escaped = to_lower( ev_filename ) ).
```
mbtools, line 128: The class might be used elsewhere and should not be changed. 

Change the filename to lower in src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap

mbtools (changes requested):

mbtools (approved):

mbtools:
Sorry, Łukasz. This was blocking. I made the necessary changes

fidley:
> Sorry, Łukasz. This was blocking. I made the necessary changes

No Marc, I am sorry. I forgot about this after my holidays. 
Thanks for adaptation!

Issue #7142, opened 2025-02-04
Non-unicode System: Dump ASSERTION_FAILED in method class_constructor of class zcl_abapgit_gui_event


listth:
Hi, 
we're facing a short dump when starting ZABAPGIT_STANDALONE (latest build).
System is a non-unicode ECC 6 EHP 7 (12) / Basis 740 (0032):

![Image](https://github.com/user-attachments/assets/aff5e005-91e0-469b-a8aa-60f95f8e05db)

In fact it's the call for CL_ABAP_CONV_IN_CE->CONVERT in method kHGwldKtJfHZXmphMQgfQYNhjRNkGw=>convert, called from xstring_to_string_utf8, which fails:  

![Image](https://github.com/user-attachments/assets/f85a2453-a912-4b92-8c0a-c7a02acce9bc)

![Image](https://github.com/user-attachments/assets/599a09a7-7d8d-42b9-ab31-5b7e06089ddb)

Any help is highly appreciated.

Thanks, Thomas

mbtools:
Please post the complete call stack

listth:
What I found out in the meantime:
- according to report RSCPINST, system is on codepage 8500
- C2A0 is not available on 8500, so I replaced the method call "gv_non_breaking_space = zcl_abapgit_convert=>xstring_to_string_utf8( lc_nbsp )." and add the value directly: "gv_non_breaking_space = |X'A0'|."
So far, this seems to work, now we're still struggeling in the import with special characters like ä, ü and ö. 
I replaced them by ae etc. in the xml files and import went fine. But if possible it would be good to have a solution also for these cases. For the test import we had only a few objects, but next one has too many objects to edit the files manually. 

As requested, here is the callstack:

![Image](https://github.com/user-attachments/assets/761c59a9-b812-4885-a8dc-51df320d6d6b)

mbtools:
We can put a fallback into the class constructor. 

The import of xml (or other data) is trickier to fix since we don't have non-Unicode systems. The dev version has many unit tests. Are any failing on the 8500 system?

listth:
First of all, many thanks for the fix.
I've changed the code accordingly and tested it in the unicode system and it works as expected.

Regarding the unit tests: I'll try to install dev version this evening and send update

listth:
Dev version imported into system, but pull stops with 2 errors:

![Image](https://github.com/user-attachments/assets/801b0657-8b1c-463e-94a5-aae044cb7ddb)

[...]

PR #7151, opened 2025-02-22
Fix unit tests for non-Unicode systems

mbtools:
Some tests didn't pass on non-Unicode systems because xstring to string does not include BOM. This should take care of the issues mentioned in https://github.com/abapGit/abapGit/issues/7142#issuecomment-2643044612.

[listth](https://github.com/listth), do you have time to test if the tests pass on your non-Unicode system (together with #7146)?


mbtools:
@listth Do you still have access to the non-Unicode system to test this?

larshp (approved):
thanks

PR #7164, opened 2025-03-24
add actionbars to command palette (F1)

christianguenter2:
side effect from #7132

Before
image

After
image


larshp (approved):
thanks, looks good to me

PR #7165, opened 2025-03-24
Replace `sotr_head*` update with API

mbtools:
Ref #7131


larshp (approved):
thanks

would probably like/need the SOTR handler class behind an interface, but I can do that later

PR #7172, opened 2025-03-29
Fix syntax errors in lower releases (table keys)

mbtools:
Unfortunately, lower releases don't like the secondary keys with these where clauses.

@christianguenter2 Is there a pragma to avoid the warnings in newer releases?

Regression #7169

image


christianguenter2:
Ah, damn. Thx for fixing. Yes there's a pragma for that. Preparing a PR.

Issue #7148, opened 2025-02-13
Too less parallel sessions related to #7074


fmabap:
Hi,

this issue is related to #7074.

We can have with the new implementation max. 6 parallel tasks (depending on how much GUI Windows are open) instead of 32 before.

![Image](https://github.com/user-attachments/assets/62d85be4-b6f8-46d7-90c1-ca49c947a9cf)

We have one repo with all 12496 Customer Objects. So the serialization takes too long without the parallelization.

Maybe you can have a look at the standard class CL_ABAP_PARALLEL how the max. available parallel sessions are determined there.

Kind regards
Frank

mbtools:
What's your setting for rdisp/max_alt_modes?

fmabap:
Current value is 6

mbtools:
This is the max online GUI sessions per user which abapgit now respects. You should see a higher parallel degree if you increase this value.

fmabap:
I did some testing.

I uncomment the limitation to the gui sessions and I got also the error with the max. number of session.

I filtered the 12000 objects in LT_TADIR, so that only TABL and DTEL remain. These are still over 5000 objects.

When I serialize these 5000 objects then the error does not occur. So I assume it is a combination of the parallelization and one special object type that tries to create an additional gui session.

I received some times in the ON_END_OF_TASK method the sy-subrc 4 ( system_failure ). LV_MESS contained the max. session err. but I could not find out which task_id it was.

fmabap:
I sorted now LT_TADIR. If I delete the following object types from LT_TADIR then it works.

object = 'IEXT' or object = 'ECTD' or object = 'ENHO' or object = 'ENQU'

I will do some further analyses maybe tomorrow.

fmabap:
I found the reason.

The error occurs at the Object Type ECTD (ECATT).

Method CL_APL_ECATT_CUSTOMIZING->READ_USER_SETTINGS

The SAP Standard calls here the method cl_gui_frontend_services=>get_sapgui_workdir. This leads to the max. session error in parallel mode (why ever...).

![Image](https://github.com/user-attachments/assets/ff759332-cb7d-43e2-b86d-723e09e66c7b)

In sequential mode does this error not occur.

So a workarond could be to run ECTD always in sequential mode.

[...]

PR #7175, opened 2025-03-29
Fix max parallel session

mbtools:
If no user sessions are available, then parallel processing needs to be disabled.

Ref https://github.com/abapGit/abapGit/issues/7148#issuecomment-2658482441


mbtools:
Merged into #7176

PR #7176, opened 2025-03-30
Fix "Maximum number of GUI sessions reached"

### LINKED ISSUE #7148: Too less parallel sessions related to #7074

mbtools:
ECATT Test Config and Data object types (`ECTC`/`ECTD`) do not support parallel serializer since they trigger user sessions by calling `cl_gui_frontend_services=>get_sapgui_workdir`. This can lead to "Maximum number of GUI sessions reached" error. 

This change processed these two object types in sequentially.

Closes #7148

### DIFF
--- a/src/objects/core/zcl_abapgit_serialize.clas.abap
+++ b/src/objects/core/zcl_abapgit_serialize.clas.abap
@@ -128,6 +128,12 @@ CLASS zcl_abapgit_serialize DEFINITION
       RETURNING
         VALUE(rv_result) TYPE abap_bool.
 
+    METHODS is_no_parallel
+      IMPORTING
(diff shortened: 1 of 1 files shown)


larshp (approved):
thanks, looks good to me

abap-weasel:
This also fixed an old issue #4833 for me. Thanks guys!

PR #7181, opened 2025-03-31
Fix AFF unit tests for lower releases

christianguenter2:
fixes these failing unit tests on NW<=757
image

image


larshp (approved):
thanks

PR #7182, opened 2025-03-31
Remove http-agent from factory

mbtools:
Using the http-agent class as singleton has side-effects (http-headers). Creating the instance from the http-agent class works just fine.

If you used this factory method somewhere else, replace it with `zcl_abapgit_http_agent=>create( )`.


larshp (approved):
thanks

Issue #5580, opened 2022-05-21
Remove Aliases


mbtools:
https://github.com/abapGit/abapGit/search?q=aliases&type=code

- [x] Aliases in `zcl_abapgit_ajson*` (exception since code is imported)
- [x] Aliases in `zcl_abapgit_html`
- [x] Aliases in `zcl_abapgit_object_xxxx`
Most of the time `ALIASES mo_files FOR zif_abapgit_object~mo_files.` can be removed since it's not used
- [x] Aliases in `zcl_abapgit_object_common_aff`
- [ ] Aliases in `zcl_abapgit_repo`
- [ ] Aliases in `zcl_abapgit_repo_online`
- [x] Aliases in `zcl_abapgit_repo_srv`
- [x] Set abaplint rule
https://github.com/abapGit/abapGit/blob/6a70f0332caf5224e205d155045f4c411ffaad1c/abaplint.json#L64

mbtools:
After the try in #7189, we will keep the `aliases` in the repo classes

PR #7189, opened 2025-04-04
Repo Refactoring: Use interfaces instead of classes

christianguenter2:
Replaces ~some~ all usages of  `zcl_abapgit_repo` and `zcl_abapgit_repo_online` with the corresponding interfaces. Should ease testing a bit.

**interface `ZIF_ABAPGIT_EXIT` is changed**


mbtools:
Good work. We want to get rid of the aliases https://github.com/abapGit/abapGit/issues/5580. This should be the chance to do just that.

christianguenter2:
Ok, turned out that getting rid of the `ALIASES` is a big endeavor 😅
So, consider this PR as a draft, as it needs more refinement.

mbtools:
Oh my. That was not my intention. I thought changing the calls using the aliases could be adjusted without replacing classes with interfaces everywhere. `mo_repo_online->set_url` would become `mo_repo_online->zif_abapgit_repo_online~set_url`. 

The `mi_repo(_online)` self references are a nice trick. It keeps the method calls simple without the long interface names. Maybe it should be `mi_me` which is shorter and makes it clear that it's not some other object? (In other project without prefixes, I will use `_me`).

We will have to see what impact the class to interface change has on other projects. I will do some testing

christianguenter2:
No worries, it's was also fun 😊
Thanks for you feedback. Good idea.

I introduced
```
    DATA mi_me TYPE REF TO zif_abapgit_repo .
```
instead of 
```
    DATA mi_repo TYPE REF TO zif_abapgit_repo .
```
in `zcl_abapgit_repo`

and 

```
    DATA mi_me TYPE REF TO zif_abapgit_repo_online .
    DATA mi_super TYPE REF TO zif_abapgit_repo.
```
instead of
```
    DATA mi_repo_online TYPE REF TO zif_abapgit_repo_online .
```
in `zcl_abapgit_repo_online`. 

I think that's an improvement as it's now more obvious that we point to ourselves/parent.

larshp:
consider if/how/when some of these changes can be merged, partial changes are okay, in order not to get into a lot of conflicts

mbtools:
I started a test... To be honest, it's a pain making many adjustments. The exists are one thing but doable. However, when it comes to code using the unofficial [API](https://docs.abapgit.org/development-guide/api/api.html#read), i.e. the repo classes, it's no fun. 

This is breaking a lot, CI for example:

image

I know it's similar to the dozen or so tools I built around AG.

You have to mention the interface but sometimes the super interface. Or you have to use the trick like the self-reference in the online repo with "me" and "super". Just not cool compared to the way it is.

I got a new appreciation for aliases esp. if several interfaces are involved. so hmmm

christianguenter2:
Yeah, I agree. Too many things breaking and the `mi_me` and `mi_super` looks suspiciously hacky. 
Though overall I think it's the right way to move more things toward the repo interfaces.
I'd try to reintroduce the aliases and therefore make the PR less breaking.

larshp:
its fine

but ALIASES is a smell of bad design IMHO, with https://github.com/abapGit/abapGit/issues/6618 there might be 3 repository types

mbtools:
- Class to Interface transition: 👍 
- Refactor + other repo types: 
  I would rather go for one repo class containing a reference to one of several repo providers (#3799, #2216). 
- Aliases + impact on others:
  The community isn't too keen on having things changed (we once renamed a parameter causing uproar). Let's keep the aliases.

larshp:
> Aliases + impact on others:
> The community isn't too keen on having things changed (we once renamed a parameter causing uproar). Let's keep the aliases.

everyone is free to setup their code and processes in a manner that discovers problems early, we need to keep moving. Anyone can also choose to stay behind

mbtools:
With the aliases it's good. Minimal changes to some renamed parameters. 

Final `lo_repo` 😉 

image

Approved from my side 👍

larshp:
image

PR #7218, opened 2025-04-21
Add support for ABAP Language Version for generic

mbtools:
Phase 3 of #6689

Adds support for object types using generic serializer class using a new field mapping rule which clears/sets the ABAP Language Version field.


larshp (approved):
thanks, looks good to me

PR #7225, opened 2025-04-25
STVI: Fix unstable diff when shdtvsvciu or shdfvguicu are not initial

leem53:
I discovered the problem with unstable diffs under HANA also exists for STVI objects


larshp (approved):
thanks, looks good to me

PR #7248, opened 2025-05-22
Always update class descriptions on deserialize

### LINKED ISSUE #7241: Non-english original language class descriptions pulled as english translation
1. Have a repository with a class with original language <> EN, for example DE, use DE as the logon language
2. Create a global class and commit it
3. Uninstall the repo
4. Pull the repo
5. Diffs where SEOCLASSTX contains duplicates of the german texts as english?!

Image

Minimal demo repo: https://github.com/fabianlupa/clas-descr-test

fabianlupa:
As a way to resolve #7241 when it happens.

This should not have any effect on the "only serialize main language" option, right?


mbtools:
i don't think "only serialize main language" is working correctly here anyway.

mbtools (approved):
Passes all CI test👍

PR #7251, opened 2025-05-22
Require CLAS and INTF for HTTP in deserialization

### LINKED ISSUE #7249: HTTP: Two pull required if repo contains any INTF
If a repo contains any INTF object the ABAP deserialization step for CLAS is prioritized after HTTP causing HTTP deserialization to fail as the implementing class does not reference the required interface yet. An exception is raised with an invalid handler message.

Image

Image

https://github.com/abapGit/abapGit/blob/7a84acdc1bee543aca2d3fbd949827ba4ece280b/src/objects/core/zcl_abapgit_file_deserialize.clas.abap#L197

Should HTTP be added to LATE? Or should it require CLAS in the priority determination?

Test repo: https://github.com/abapGit-tests/HTTP_order

fabianlupa:
Fixes #7249


mbtools (approved):
👍

PR #7259, opened 2025-05-25
Suppress PO comments

sbcgua:
The feature to suppress comments in PO files. Addresses the issue noted e.g. [here](https://github.com/abapGit/abapGit/issues/2539#issuecomment-1520251413)

At Local settings

image

The enabled flag removes the green line

image

## Refactorings

1) zif_abapgit_object~map_filename_to_object now change filename only (!) - actually I'd suggest refactoring this method further - will open an issue to discuss
  - + changes in DEVC and SICF
2) filename_logic
  - minor semantic changes in get_lang_and_ext
  - changes in object_to_file to meet changes in zif_abapgit_object~map_filename_to_object
3) zcl_abapgit_lxe_texts
  - changed selection of LXE lang pairs
4) zif_abapgit_i18n_file~lang_suffix - new method to output complete lang suffix (they are different in PO and PROPS - the props implementaion broke things for po)


sbcgua:
some subtle bugs ... checking yet

sbcgua:
Anyone knows why language mapping db is a hardcoded part of AG ... ?

image

sbcgua:
> we dont know a way to do it that is compatible with ABAP Cloud

but could it have been made abap-cloud-specific then? Also there are several version of the same language ... (e.g. German on the screenshot) ... 

well, it's a bit messy now, e.g. for some languages PO files now have 4 letter language suffix and for some just 2 ...

Also the `zif_abapgit_object~map_object_to_filename` (was is relatively new, right?) redefines the name of the file (package.xml) ... including PO files, so now there is a dump with name duplication. it's fixable of course ... maybe it is even an old bug ... just that it need some fixes

sbcgua:
P.P.S. `map_object_to_filename` receives language so in theory it can create the complete proper name inside, but that means that it must know about all kinds of extra files which is bad. It just doesn't by design. An extra reason to refactor this thing.

sbcgua:
Is there a way to check tables/FMs available in cloud?
- FM LXE_T002_CHECK_LANGUAGE?
- FM LXE_T002_GET_LANGUAGES
- Tables lxe_t002x and t002 ?

Looks like the CL_I18N_LANGUAGES was not the best option to do the lang conversion and LXE internally uses these FMs and tables. in particular lxe_t002x and t002x are a bit different ...

larshp:
> Is there a way to check tables/FMs available in cloud?

https://developers.sap.com/tutorials/abap-environment-trial-onboarding..html

But its SAP...

larshp (approved):
thanks, lgtm

PR #7282, opened 2025-06-06
PO UTF8 BOM and stage page duplication rule

sbcgua:
Minor fixes to PO related logic:

1) remove UTF8 BOM marker - looks like they are not supported by native gettext tools, and they are unnecessary in general in this case
2) stage page logic that checks duplication excludes package.devc.xml but now also relevant .po files (by the way this is certainly a "business logic" si ideally should be moved to the "model" eventually)


mbtools (approved):
👍

PR #7288, opened 2025-06-09
Root package in where used

sbcgua:
In case of deep packages it might be more informative to show the root package of the Where-used package. So here is a switch for this (by default the behavior is to show parent package as it was before)

image


larshp (approved):
thanks, looks good to me

PR #7289, opened 2025-06-11
Fix dump in data loss check

mbtools:
Fixes dump if table is not changed but moved to another package

image

image


larshp (approved):
lgtm

PR #7290, opened 2025-06-11
Fix maintaining background settings

mbtools:
Regression #7229


kjetil-kilhavn:
Tested. The settings page now displays the correct setting value both initially and after changing the mode and saving.

PR #7297, opened 2025-06-13
Remove package check when creating offline repo

### LINKED ISSUE #7253: Automatic package creation does not work for offline repos
According to the docs the feature of creating a package automatically instead of having to press the _Create Package_ button should also work for offline repositories. It doesn't though, I get the following message.

Image

Docs:

Image

https://github.com/abapGit/docs.abapgit.org/blob/e9453ccda5f70af3ce6e9ec852c17ad2b25d8b15/src/user-guide/projects/offline/install.md?plain=1#L15

Are the docs wrong or was this broken at some point?

mbtools:
Check is not necessary. The package will be create during pull (after zip import).

Closes #7253


larshp (approved):
lgtm

PR #7300, opened 2025-06-16
Refactor: Move status calc to interface

mbtools:
Separates status calculation from repo layer and consistency check. No changes to calculation logic.


larshp (approved):
thanks

PR #7302, opened 2025-06-22
Exclude objects from i18n

### LINKED ISSUE #7293: i18n: add the possibility to exclude objects from translation
The case: a large package with subpackages, which also has contributed (co-packed) libs (e.g. ajson or whatever of this sort). The libs have technical nature, used internally thus do not need any translation. In fact, one can also imagine a "native" subpackage with utilities which never face the user. In such a case it can be reasonable to exclude these objects or packages from translation (to reduce number of files and "red flags" of missing translations). Or, more accurately, force "master language only" upon these particular objects.

current thoughs:

- it should be a list of paths, like Exclude paths feature

Image

- it should probably be the repo setting (so a part of .abapgit, rather then local in-system setting)

Looks like an easy thing to implement, but would appreciate opinions.

sbcgua:
closes #7293

image

image

**Final solution**:
- build a simplified path, from tadir item (from serialization - it has package path!): `/src/pkg/subpkg/obj.type`, so not xmls, the type is kind of file extension
- check this path versus list of patterns
- the pattern without `/` or `*` at start are prepended with `*/` (kind of relative path - though I'm not 100% as it becomes implicit... yet convenient)

### DIFF
--- a/src/ui/pages/sett/zcl_abapgit_gui_page_sett_repo.clas.abap
+++ b/src/ui/pages/sett/zcl_abapgit_gui_page_sett_repo.clas.abap
@@ -35,7 +35,7 @@ CLASS zcl_abapgit_gui_page_sett_repo DEFINITION
         main_language    TYPE string VALUE 'main_language',
         i18n_langs       TYPE string VALUE 'i18n_langs',
         use_lxe          TYPE string VALUE 'use_lxe',
-        master_lang_only_objs TYPE string VALUE 'master_lang_only_objs',
+        wo_transaltion   TYPE string VALUE 'wo_translation',
         starting_folder  TYPE string VALUE 'starting_folder',
         folder_logic     TYPE string VALUE 'folder_logic',
         ignore           TYPE string VALUE 'ignore',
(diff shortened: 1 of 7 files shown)


sbcgua:
Works. But needs a fresh minded review. E.g. I wonder if a stricter validation is required (now it's missing in fact) and also a bit unsure about implicit addition of `*/` although it is convenient ... but feels bugprone.

src/ui/pages/sett/zcl_abapgit_gui_page_sett_repo.clas.abap
```diff
@@ -160,6 +161,14 @@ CLASS zcl_abapgit_gui_page_sett_repo IMPLEMENTATION.
       iv_name        = c_id-use_lxe
       iv_label       = 'Use LXE Approach for Translations'
       iv_hint        = 'It''s mandatory to specify the list of languages above in addition to this setting'
+    )->textarea(
+      iv_name        = c_id-wo_transaltion
+      iv_label       = 'Objects (wildcard) to keep in master language only (without translation)'
```
mbtools: Let's stick with "main language" on the UI

sbcgua: done

PR #7358, opened 2025-08-07
Ignore empty translation table

sbcgua:
Subtle bug which raised error if translation block appeared empty.

Looks like no one uses PO files :P Well ... first to blame myself, I also haven't do it productively till this year...


mbtools (approved):
👍

PR #7363, opened 2025-08-08
remove REPT "DUMMY KEY FOR DDIC FLAG COPY"

### LINKED ISSUE #7314: LXE and REPT: DUMMY KEY FOR DDIC FLAG COPY
Anyone knows what this is ? (maybe makes sense to force-remove it from translations)

Image

It has a very strange behavior. It presents in the original system but disapears when deployed, even to a system with the same abap version.

sbcgua:
closes #7314


sbcgua:
not quite agree with some linter rules ...

sbcgua (commented):
linter fix

mbtools (approved):
👍

PR #7443, opened 2025-10-25
cloud: remove use of DIRTREE

larshp:
function module `WB_TREE_ACTUALIZE` takes input TYPE ANY, so this should be okay

tested ok, no crash


larshp:
merging

PR #7449, opened 2025-11-04
Use user defaults for new repos

pokrakam:
When setting a default user/email in global settings -> Personal -> Git Default Values, I expect it to be used for new repos. This does not happen, it ignores and uses my SAP user profile name/email. Fixed.
In the config XML it's in: 
```xml
<SETTINGS>
  <DEFAULT_GIT_UNAME>
  <DEFAULT_GIT_EMAIL>
```
but did not appear to be used.

### DIFF
--- a/src/persist/zcl_abapgit_persistence_user.clas.abap
+++ b/src/persist/zcl_abapgit_persistence_user.clas.abap
@@ -24,7 +24,6 @@ CLASS zcl_abapgit_persistence_user DEFINITION
       ty_repo_configs TYPE STANDARD TABLE OF ty_repo_config WITH DEFAULT KEY .
     TYPES:
       BEGIN OF ty_user,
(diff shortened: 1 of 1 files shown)


pokrakam:
Mmh, the integration test failures don't seem to have anything to do with my changes 😕

src/persist/zcl_abapgit_persistence_user.clas.abap
mbtools, line 27: can we remove the old field?

mbtools (commented):
Yes, test failure is unrelated. Old fields looks to be superfluous now

src/persist/zcl_abapgit_persistence_user.clas.abap
pokrakam: Good point it looks unused, removed.

Issue #7438, opened 2025-10-23
Pragma for regex


mbtools:
There are a bit over 100 regex used which cause warnings in higher releases.

To remain compatible, we can't change to PCRE. 

Should we add `##REGEX_POSIX`?

<img width="1400" height="299" alt="Image" src="https://github.com/user-attachments/assets/b7918d5d-2951-447a-8458-5ea561ecfd2a" />

christianguenter2:
Hm, doesn't that change one warning for another? I get this in lower releases (e.g. 740).
<img width="2014" height="73" alt="Image" src="https://github.com/user-attachments/assets/c308b805-ea20-4aad-90f4-3ab38af8d8eb" />

fabianlupa:
We already have syntax warnings in lower releases if I recall correctly (ABAP Doc I think for example?). To me it seems preferable to get rid of the warnings in higher releases where most of the active development happens (presumably).

christianguenter2:
Yes there are:
<img width="2164" height="155" alt="Image" src="https://github.com/user-attachments/assets/ae1be7b0-f32b-4109-9671-8f1b07ae3d66" />
I agree with @fabianlupa s reasoning.

larshp:
> We already have syntax warnings in lower releases if I recall correctly (ABAP Doc I think for example?). To me it seems preferable to get rid of the warnings in higher releases where most of the active development happens (presumably).

yea, we should put this into https://docs.abapgit.org/development-guide/ as a design decision, aim for no warnings on high releases, no syntax errors on low

mbtools:
Let's ask copilot to do this in a quiet moment: code and doc changes

PR #7450, opened 2025-11-11
Refactor regex usage to include ##REGEX_POSIX pragma

mbtools:
I used Cursor with the Composer 1 model to work on #7438 with the following prompt:

```
scan all abap files and add a ##REGEX_POSIX pragma to FIND and REPLACE statements 
or FIND and REPLACE functions that use a REGEX.  The pragma should be placed at the 
end of the statement before the period.
```

It did well for the statements but ignored all functions 🤷 . I had to add the second sentence since it put the pragma in front of the statement at first. Overall it used 54% of the 200K context window and took about 2 min.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,14 +24,14 @@
     ]
   },
   "devDependencies": {
-    "@abaplint/cli": "^2.113.233",
+    "@abaplint/cli": "^2.113.236",
     "@abaplint/database-sqlite": "^2.11.78",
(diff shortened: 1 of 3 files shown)


mbtools:
When prompted to also add pragma to built-in find and replace functions, I had to give it some examples before it worked. Interestingly, this time the agent created a script to make the changes instead of adjusting the files directly.

It does not include usage of `cl_abap_regex` but it's a good start.

PR #7453, opened 2025-11-18
Performance zlib stream

mbtools:
This should make it faster for large amounts of data. 

Ref #7431


larshp:
nice, do you have any performance numbers to share?

but overall looks good, only reading and not updating seems good

mbtools:
I don't have a test case. This is just good ol' ABAP tuning. If something is called millions of times, every millisecond counts.

PR #7462, opened 2025-12-04
bugfix: delete `SUSH` after `G4BA`

larshp:


larshp:
I could not get it working setting `DELFLAG` with fm `TADIR_INTERFACE` 😔

PR #7464, opened 2025-12-04
DDLS: Respect user setting when display CDS views

mbtools:
If "Show in ADT" is disabled, CDS views are shown in SAP GUI (if available).

Here's an example (7.58):

image


larshp:
nice

PR #7467, opened 2025-12-10
ENHO: Add check for identical enhancements in hook class

larshp:
some ENHO can be inconsistent, resulting in a crash when adding identical filenames to `mo_files`, this adds a guard and a proper error message


mbtools (approved):
👍

PR #7468, opened 2025-12-10
fix crash in XML output

larshp:
avoiding short dump XML_CONVERSION_NO_NUMBER	CX_SY_CONVERSION_NO_NUMBER in CALL TRANSFORMATION


src/xml/zcl_abapgit_xml_output.clas.abap
```diff
@@ -67,6 +67,7 @@ CLASS ZCL_ABAPGIT_XML_OUTPUT IMPLEMENTATION.
 
     CALL TRANSFORMATION id
       OPTIONS initial_components = 'suppress'
+      value_handling = 'move'
```
mbtools, line 70: I guess this option is available in 702 but I will check

mbtools: works!

larshp:
thanks

PR #7471, opened 2025-12-12
SOTS: fix crash

larshp:


mbtools:
There's a header in the database but not details. That's a DB inconsistency, no? It should error in such case.

larshp:
now raises exception

mbtools (approved):
Thx 👍

PR #7215, opened 2025-04-16
New error handling for not supported object types

christianguenter2:
This change introduces a new specialized exception class `zcx_abapgit_type_not_supported` inherited from `zcx_abapgit_exception` replacing the latter in serializer classes.
Additionally all checking whether an object type is supported or not is moved to the constructors of serializer classes, reducing some redundancies and boilerplate 'not supported handling'. Plus make it more obvious if an object type is supported or not: If the serializer can be instantiated it is supported.

PR #7472, opened 2025-12-15
bugfix: propagate exceptions from object serializer constructors

larshp:


mbtools:
The point of https://github.com/abapGit/abapGit/pull/7215 was to *not* raise `zcx_abapgit_exception` but introduce a separate exception (because it need different handling). Whichever object class raise `zcx_abapgit_exception` in its `constructor` needs to be changed. Or we catch all exceptions in `create_object` and map them to `zcx_abapgit_type_not_supported`. 

PS: This should have been [documented](https://docs.abapgit.org/development-guide/serializers/serializers.html#constructor) but we missed that after #7215.

larshp:
`DOCT`, https://github.com/abapGit/abapGit/blob/main/src/objects/zcl_abapgit_object_doct.clas.abap#L6

mbtools:
There are a few other object types that also use `zcx_abapgit_exception` in the `constructor` definition. 

I would only change `create_object`:

```
CATCH cx_sy_create_object_error. " before
CATCH cx_sy_create_object_error zcx_abapgit_exception. " after
```

larshp:
closing, see https://github.com/abapGit/abapGit/pull/7484 instead

PR #7474, opened 2025-12-16
`IOBJ`: read inactive if no active version exists

larshp:
https://docs.abapgit.org/development-guide/serializers/overview.html#overview

### DIFF
--- a/src/objects/zcl_abapgit_object_tran.clas.testclasses.abap
+++ b/src/objects/zcl_abapgit_object_tran.clas.testclasses.abap
@@ -76,7 +76,7 @@ CLASS ltcl_split_parameters IMPLEMENTATION.
     then_param_shd_be(
       iv_line  = 1
       iv_field = 'VIEWNAME'
-      iv_value ='ZMM_BF_SPEZ_ROLE' ).
+      iv_value = 'ZMM_BF_SPEZ_ROLE' ).
 
     then_param_shd_be(
       iv_line  = 2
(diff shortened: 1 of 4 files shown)


src/objects/zcl_abapgit_object_iobj.clas.abap
```diff
@@ -449,9 +455,17 @@ CLASS zcl_abapgit_object_iobj IMPLEMENTATION.
 
     lv_iobjnam = ms_item-obj_name.
 
+    IF zif_abapgit_object~is_active( ) = abap_true.
+      lv_version = 'A'.
+    ELSE.
+      " if its not active then look for content version
+      lv_version = 'D'.
```
mbtools, line 462: That's not correct. BW does have three versions. Active (A), inactive (I), and content (D). Content is delivered by SAP or partners as templates. Only when the content/template is activated, it becomes part of the customer system and relevant for abapGit (A). (I) and (D) need to be ignored.

mbtools (changes requested):

src/objects/zcl_abapgit_object_iobj.clas.abap
larshp: so it only exists() if its in version `A` ?

mbtools: Correction: serialize only A-version. Exists if either A or I-version is present. D-Version is a different tadir object type.

larshp: its the IOBJ type in my example, anyhow, just want it to not crash

mbtools: Just ignore D

larshp:
closing, some other day

Issue #3162, opened 2020-01-19
Support for more BW Objects


captainabap:
Hi,

Abapgit is great. Thank you. I use it for my private ABAP Projects. But for my work, I could need support for the following BW Objects:
- ADSO - Advanced DSOs
- AREA - InfoAreas
- DTPA - DAta Transfer Processes
- ROUT - Routines
- RSDS - DataSources
- TRFN - Transformatio

Is this possible? SAP is still struggling with importing transports of various BW Objects. So it can be very challenging. 

Best regards,
Jörg

christianguenter2:
Sure, we can work on it. But I guess all the main abapGit contriubutors are not working in the BW area, so we need the help of the community. Recently @pawelwiejkut started with the BW objects IOBJ and ODSO. Maybe you can try to implement the serializers for the mentioned object types and we can assist?

pawelwiejkut:
Hello all,
@captainabap for now, I'm working on TRFN, and AREA support so please leave this two to me if you want to contribute. Next I can help with other objects, but personally I don't have access to the BW on/4HANA systems so it will be hard with new objects like ADSO.

captainabap:
Hello, 
thank you for your fast feedback. @pawelwiejkut: It's great news, that you are already  working on AREA and TRFN. I can try ADSO, but it will take some time. 
How is the correct order of the different Object types controlled. E.g InfoObjects must be imported before ADSOs. And TRFN must be imported after ADSOs, right?
And have you also covered AMDP Routines in the TRFN serializing?

pawelwiejkut:
@captainabap, haha it was also one of my questions, so let me answer. Prioritization of objects deserialization is done here: https://github.com/larshp/abapGit/blob/master/src/zcl_abapgit_objects.clas.abap#L1040

larshp:
suggest adding a note explaining it in https://docs.abapgit.org/development/serializers.html

pawelwiejkut:
@captainabap Hard to say will AMDP be supported in the first version of TRFN  (can't test on non-HANA). Btw. I have looked at my notes and I saw that for ADSO `cl_rso_adso_api=>create` can be a good start point, but of course, the choice is yours.

captainabap:
@pawelwiejkut Thank you! This ADSO class looks promissing.

goalkeeper84:
HI Guys,
may I also can contribute. I have a BW4 and a 7.5 available to test.

goalkeeper84:
[...]

PR #7477, opened 2025-12-16
`ELEM` support

larshp:
could not find any APIs


mbtools:
That's BW, part of BW queries. For several reasons (https://github.com/abapGit/abapGit/issues/3162#issuecomment-658361843), it's  a can of worms. Supporting these kind of objects raises expectations that abapGit will not be able to handle in the current state. Do we really need ELEM?

larshp:
closing

PR #7484, opened 2025-12-22
fix error handling in create_object()

larshp:


src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -399,7 +399,7 @@ CLASS ZCL_ABAPGIT_OBJECTS IMPLEMENTATION.
               is_item     = is_item
               iv_language = zif_abapgit_definitions=>c_english.
         ENDIF.
-      CATCH cx_sy_create_object_error.
+      CATCH cx_sy_create_object_error zcx_abapgit_exception.
```
mbtools, line 402: There's another exception below

larshp: fixed

PR #7489, opened 2026-01-02
flow: optimize

larshp:


src/cts/zcl_abapgit_cts_api.clas.abap
```diff
@@ -636,6 +636,7 @@ CLASS zcl_abapgit_cts_api IMPLEMENTATION.
 
     SELECT trkorr FROM e070 INTO TABLE lt_tasks
       WHERE strkorr = iv_request
+      AND trfunction = zif_abapgit_cts_api=>c_transport_type-wb_task
```
mbtools, line 639: Hmm. Objects can be included on requests directly without a task. Not sure if that makes a difference somewhere.

larshp: hmm, yea, this is mainly to exclude repairs

note: this method is currently only used on the flow page

mbtools: 👍

larshp:
only flow related, so merging

Issue #6281, opened 2023-05-14
SRVB diffs bind_type_data and responsible


fabianlupa:
![image](https://github.com/abapGit/abapGit/assets/5270359/6cca44a1-81f7-49ce-9b2c-a8a59e5b0669)

![image](https://github.com/abapGit/abapGit/assets/5270359/a869e74c-ac70-4845-ad09-e4c4b0b623e5)

- RESPONSIBLE should be cleared on serialize
- (MASTER_LANGUAGE should not be serialized and instead use main language of the repo?)
- The SKTD document reference seems to be an error on the remote as it does not seem to be part of the repo?
- BIND_TYPE_DATA and PUBLISHED / BINDING_CREATED: Not sure if these should be serialized at all? I. e. should pull publish automatically (then it doesn't seem to do it for me) or should this be manual and therefore not serialized?

mbtools:
Should all be fixed

PR #7495, opened 2026-01-07
SUSH: Fix "lead application does not exist"

### LINKED ISSUE #7247: SUSH: Deserialize and deletion issues
https://github.com/abapGit-tests/SUSH works, but there are other issues:

- Deserializing "lead application does not exist" 
  Examples: https://github.com/abapGit-tests/HTTP, https://github.com/abapGit-tests/SAJC_dep
- Incomplete deletion (#7128)

mbtools:
This serializer needed a good work-over. For TADIR or Service related SUSH objects (types `HT` and `HS`), the XML now contains the metadata for the lead application (`USOBHASH`) which is required to create SUSH object properly.

The change removes `USOBX_EXT` and `USOBT_EXT` from the XML which do not contain any relevant data, just texts and icons for the SU22 UI. It also removes several unnecessary nested TRY-CATCH statements.

The serializer has also been enhanced to cover the SU22 data for transactions (#4577, #7497).

Closes #7247


fabianlupa:
Amazing, I can do some tests this weekend if you like. I have some repos in mind where the SUSH objects "got lost" or were never able to be pulled.

mbtools:
Please test 🙏 

There are four types: transaction (TR), rfc function (RF), tadir object (HT), and service object (HS). We don't have any test case for HS.

fabianlupa:
I ran out of time but should be able to have a look in the next few days.

fabianlupa:
Hmm, I guess the migration from old to new format "version" will be tricky. Some notes:

I have SUSH objects in the repo with the old format. According to abapGit they exist in the remote but do not in the system. If I pull them though, exists() return true as an USOBHASH entry exists. Going further deserialize fails because of "Default values might not be sufficient". I do have other SUSH objects in that repo though that do exist (fully) in the system and show the same "error" message when opening them.

Currently checking with https://github.com/abap-kochbuch/abap-kochbuch/tree/english and this kind of state:

image

Note also the !USOBHASH in the last change by column.

mbtools:
I pulled the cookbook into a ABAP Cloud Dev Trial (7.58, no existing objects):

image

1. SRVB

image

Known diff due to published state (https://github.com/abapGit/abapGit/issues/6281). AG should probably publish it, too. The post-install steps are already documented in the cookbook.

2. SUSH

image

Pulling works with hash conflict. The diffs are due to the updated format.

image

image

One [SUSH object](https://github.com/abap-kochbuch/abap-kochbuch/blob/english/src/excel/f25aad387e955b9d8ef69230a5417cht.sush.xml) has an error but that's expected since it's missing the `USOBX` and `USOB_SM` entries. 

"Last changed by" was updated correctly to my username, too. The `!USOBHASH` comes into play if there's a conflict, I believe.

image

"Default values might not be sufficient" is a warning in my system. I did not get it as an error in AG.

From this point of view, it works well. It looks correct for 2 of the SUSH in your system (ie. just a diff). I don't know why your system has other hashes for 4 of them.

fabianlupa:
> "Default values might not be sufficient" is a warning in my system. I did not get it as an error in AG.

All warnings are converted to errors for me here:

image

Errors are then converted to exceptions here:

image

Which causes deserialize to end in the catch block skipping corr_insert.

Seems like note 3485891 patches this in (in 758 SP03).

fabianlupa:
This is my USOBHASH for ZACB*:

image

This is my TADIR for SUSH in the cookbook package(s).

image

(Ignore $COOKBOOK_SAPSUPPORT and ZACB_UI_LABELS2_O4 which do not belong to the repo)

I assume I have some leftover USOBHASH entries from some previous shenanigans? :D Did abapGit maybe not correctly delete these entries previously? Should exists() check both TADIR and USOBHASH? Or do we maybe just need to cleanup inconsistencies ourselves?

mbtools:
I applied the note and now there are indeed errors (instead of warnings). Since they show in SU22, and what I see in SU22 matches the repo, I don't think it's an AG problem:

image

image


My `usobhash` matches `tadir` and the repo:

image

image

I tried using `CL_SU22_ADT_OBJECT=>IF_SU22_ADT_OBJECT~CHECK_EXISTS` but ran into some issues. That's why I left `exists` with the select on `usobhash` which is the primary table for `sush`.

Since the repo is good and pulls ok, it seems like something did go wrong in your system. I would drop all related entries in `USOB*` tables and pull these objects from the repo. Then maintain them to remove config errors. Once green push to the repo (it should still match the hashes).

fabianlupa:
I'll clean up my system. `corr_insert` should still be called though I think if the exception is a "valid" outcome and the object was created/updated in the system.

mbtools:
Grrr, since applying the note, errors lead to a partial update of the object. It does not pull `usob_sm` for example. If a SUSH has errors, we probably should not serialize it. hmmm

mbtools:
It needs two notes to work properly:
- [3485891](https://me.sap.com/notes/3485891) - ADT: Check for APIS
- [3616245](https://me.sap.com/notes/3616245)- SU22 - incorrect check for authorization objects

I will update https://docs.abapgit.org/user-guide/reference/note-analyzer.html accordingly. 

The cookbook can be pulled now including SUSH (after https://github.com/abap-kochbuch/abap-kochbuch/pull/21). The only diffs are for the PUBLISHED status of SRVB.

fabianlupa:
I'll have a look this weekend. Thanks for all the fixes!

fabianlupa:
I cleaned up my USOBHASH, applied 3616245, and completely removed the repo. Then pulled mbtools:sush-defaults. No errors on pull. SU22 shows all the values.

Looks great to me!

PR #7496, opened 2026-01-07
TOBJ: Fix import of primary table flag

mbtools:
For "Individual Transaction Objects" (type = T), the primary table flag and some other fields were missing from the import.

An example can be found in https://github.com/abapGit-tests/SMBC_full. Here's the diff before the fix:

image


fabianlupa:
Tested with SMBC_full, works great 👍

fabianlupa (approved):

Issue #7492, opened 2026-01-07
Issue with "Data Config"


carrotcn:
Hello,

I encountered a problem when experimenting the table content transfer with "Data Config". There is no new TABU entry added after specifying the table in the menu.

I started with a new package that contains one table:

<img width="711" height="468" alt="Image" src="https://github.com/user-attachments/assets/ab534e1e-ee9f-4589-97e4-b35fc7ba2c58" />

I specified the table name in "Data Config" menu, with the WHERE clause. Then clicked the "Add" button.

<img width="1255" height="937" alt="Image" src="https://github.com/user-attachments/assets/44e52bc1-9dbd-47a6-bd00-a3ecaba998f8" />

And I verified there is existing data in the table:

<img width="1053" height="237" alt="Image" src="https://github.com/user-attachments/assets/aa675176-6f17-47a0-8663-f3db0ff08229" />

However, when I clicked the "back" button on top-right from the "Data Config" after finishing the configuration, there are still the same three objects in the repository. I did not see any TABU entries introduced.

Could you please let me know if I missed any steps here? Thanks!

mbtools:
It looks like you are configuring it correctly and you should see a TABU entry.

Can you create a public test repo so we can reproduce the issue, please?

carrotcn:
Sure! Here is the URL to the public test repo:
https://github.com/carrotcn/tab_test_repo.git

It does not contain table data (for obvious reason...) but feel free to insert something to the table after import if needed.
The test system I am using is S/4HANA on-premise 2023 SPS03 (S4CORE 108) if that helps.

Thank you.

mbtools:
Unfortunately, these are S4 data elements that don't exist in plain ABAP systems:

<img width="894" height="266" alt="Image" src="https://github.com/user-attachments/assets/ebf8e321-918b-4ead-ab35-086dfaab6296" />

<img width="687" height="495" alt="Image" src="https://github.com/user-attachments/assets/e5354d7a-b48b-4e3b-bfd7-0f1e1461dc02" />


I changed it to built-in types for my test:

<img width="684" height="246" alt="Image" src="https://github.com/user-attachments/assets/490acbb7-ab6d-46c4-ba7b-63cc6f81eba0" />

I added the same data config as you and it works correctly (v1.132.0):

<img width="1262" height="415" alt="Image" src="https://github.com/user-attachments/assets/240a1e8f-d09b-4244-a022-3fd4ed874c48" />

[...]

PR #7498, opened 2026-01-08
TABU: Prepare local storage of data config (Part 1)

mbtools:
Data config does not work if it's not pushed to remote or if repo is offline and without ZIP. This PR enhances the persistency classes to support local storage of the data config.

Repo and UI layers will be added separately.

Ref #7492

### DIFF
--- a/src/git/v2/zcl_abapgit_gitv2_porcelain.clas.abap
+++ b/src/git/v2/zcl_abapgit_gitv2_porcelain.clas.abap
@@ -234,7 +234,7 @@ CLASS zcl_abapgit_gitv2_porcelain IMPLEMENTATION.
       it_arguments = lt_arguments ).
 
     " add dummy packet so the v1 branch parsing can be reused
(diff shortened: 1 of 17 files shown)


src/data/zif_abapgit_data_config.intf.abap
```diff
@@ -53,4 +53,14 @@ INTERFACE zif_abapgit_data_config
...
+  METHODS save_config
+    IMPORTING
+      iv_repo_key TYPE zif_abapgit_persistence=>ty_repo-key
+    RAISING
+      zcx_abapgit_exception.
+  METHODS load_config
+    IMPORTING
+      iv_repo_key TYPE zif_abapgit_persistence=>ty_repo-key
+    RAISING
+      zcx_abapgit_exception.
```
larshp: feels like this class is getting too many methods that have names which can easily be misunderstood, perhaps its possible to extract these two methods to a separate interface?

mbtools:
Naming was confusing, indeed. I moved load/save to a separate interface.

PR #7500, opened 2026-01-09
Fix "identical keys" error when creating offline repo

### LINKED ISSUE #7481: Identical keys error when creating offline repo after previous error
1. New Offline
2. Fill out name and package
3. Create Offline Repo
4. Get an error message, because the folder logic cannot be prefix
5. Change the folder logic to Full
6. Create Offline Repo
7. Image

You then need to restart abapGit. I assume the repo in creation is added regardless of the error, causing the validation to fail on a second attempt.

Image

mbtools:
Makes sure that exceptions raised in checksum rebuild are caught and repo storage is reset.

Closes #7481


larshp:
nice

PR #7505, opened 2026-01-14
git get_length: refactor + additional unit tests + performance

larshp:
better performance by working on hex values instead of each bit (I dont have any measurements tho)

### DIFF
--- a/src/data/zcl_abapgit_data_serializer.clas.abap
+++ b/src/data/zcl_abapgit_data_serializer.clas.abap
@@ -51,12 +51,12 @@ CLASS zcl_abapgit_data_serializer IMPLEMENTATION.
         lo_ajson->keep_item_order( ).
         lo_ajson->set(
           iv_path = '/'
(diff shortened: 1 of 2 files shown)


mbtools:
It currently takes 1.5ms per call and I think it's one call per object. It would have to be a very large repo to see a 1 sec improvement. But the new code is nicer :-)

larshp:
> It currently takes 1.5ms per call and I think it's one call per object. It would have to be a very large repo to see a 1 sec improvement. But the new code is nicer :-)

Yea, it just seemed like a low hanging fruit, also reducing code, but not working yet

larshp:
okay, it does work, some numbers below. I had like 400k calls to this method in a trace

before:
image

after:
image

mbtools:
Factor 4x 👍 Lgtm

larshp:
its something on 400k calls, and cleans up the code 🏋️‍♂️

PR #7507, opened 2026-01-15
TABL: refactor GUI interaction to earlier when deleting with data

larshp:
when deleting a TABL with data, https://github.com/abapGit/abapGit/blob/main/src/objects/tabl/zcl_abapgit_object_tabl.clas.abap#L678-L702, it asks the user in SAP gui.
This PR moves the validation/decision to earlier in the process, along with the other deserialize checks/flags

### DIFF
--- a/src/objects/zcl_abapgit_object_srvb.clas.abap
+++ b/src/objects/zcl_abapgit_object_srvb.clas.abap
@@ -316,8 +316,12 @@ CLASS zcl_abapgit_object_srvb IMPLEMENTATION.
     FIELD-SYMBOLS <ls_create_info> TYPE any.
     FIELD-SYMBOLS <lv_description> TYPE any.
     FIELD-SYMBOLS <lv_field> TYPE any.
+    FIELD-SYMBOLS <ls_service_binding> TYPE any.
 
-    ASSIGN COMPONENT 'PUBLISHED' OF STRUCTURE mr_service_binding->* TO <lv_field>.
+    ASSIGN mr_service_binding->* TO <ls_service_binding>.
+    ASSERT sy-subrc = 0.
+
+    ASSIGN COMPONENT 'PUBLISHED' OF STRUCTURE <ls_service_binding> TO <lv_field>.
     IF sy-subrc <> 0 OR <lv_field> <> abap_true.
       RETURN.
     ENDIF.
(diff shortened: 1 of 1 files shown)


larshp:
test with https://github.com/abapGit-tests/TABL-delete-data and the program + branch there

image

src/objects/core/zcl_abapgit_objects_check.clas.abap
```diff
@@ -492,4 +521,104 @@ CLASS zcl_abapgit_objects_check IMPLEMENTATION.
...
+
+    lt_remote_files = ii_repo->get_files_remote( iv_ignore_files = abap_true ).
+
+    " Check for tables being deleted that contain data
+    LOOP AT it_results ASSIGNING <ls_result>
+      WHERE obj_type = 'TABL'
+        AND match IS INITIAL
+        AND packmove IS INITIAL ##PRIMKEY[SEC_KEY].
+
+      READ TABLE lt_remote_files TRANSPORTING NO FIELDS WITH KEY filename = <ls_result>-filename.
```
mbtools: Image

`READ TABLE lt_remote_files TRANSPORTING NO FIELDS WITH KEY file COMPONENTS filename = <ls_result>-filename.`

src/objects/core/zcl_abapgit_objects_check.clas.abap
```diff
@@ -492,4 +521,104 @@ CLASS zcl_abapgit_objects_check IMPLEMENTATION.
...
+      " Table contains data
+      READ TABLE rt_overwrite TRANSPORTING NO FIELDS WITH KEY object_type_and_name COMPONENTS
+        obj_type = <ls_result>-obj_type
+        obj_name = <ls_result>-obj_name.
+      IF sy-subrc <> 0.
+        CLEAR ls_overwrite.
+        MOVE-CORRESPONDING <ls_result> TO ls_overwrite.
+        ls_overwrite-devclass = <ls_result>-package.
+        ls_overwrite-action   = zif_abapgit_objects=>c_deserialize_action-delete_tabl_with_data.
+        ls_overwrite-icon     = icon_delete.
```
mbtools, line 616: The "remove table" and "data exists" popups look very similar:

Image

Image

I tried with `ICON_MESSAGE_ERROR_SMALL` which is more visible:

Image

mbtools (changes requested):
Works well. Just some tweaks

src/objects/core/zcl_abapgit_objects_check.clas.abap
larshp: fixed in https://github.com/abapGit/abapGit/pull/7507/commits/8a4e496038bb19abbb9db8ddb00d6de01f5f9ff5

src/objects/core/zcl_abapgit_objects_check.clas.abap
larshp: yea... I think we will be refactoring that popup sometime in the future

also so it works on web

mbtools: There are too many popups during pull anyway. It should be one page with list of objects and a single decision (checkbox) per object. Next to each object a list of things that happen and need to be confirmed (will be overwritten, contains data, etc). 

This is good to merge

larshp: agree

mbtools:
Original text still says "wip, not tested" ;-)

PR #7514, opened 2026-01-21
flow: refactor finding changes in git

larshp:
only changes to flow


larshp:
looked through copilot comments, think its all good

PR #7515, opened 2026-01-21
git protocol performance: utf8, dont check for byte order mark

larshp:
in the git protocol there are no byte order marks

this refactor improves performance, removes this call,
image
which in this case is almost a second


mbtools (approved):
👍

PR #7521, opened 2026-01-22
Feature/validate after push exit

gardian12:
Hello together,

I've added a new method `VALIDATE_AFTER_PUSH` to the `IF_ABAPGIT_EXIT` interface, because of I need to retreive the generated commit hash.

This is very valuable, while using this and other information like current branch to generate a version class for an application.
So the integration of GIT meta information into the code base will be possible.

A combination with the existing exit functionality was done. The lack of an event after push was the result of this investigation.

So I would be very thankful, if this feature will part of _abapGit_.

As an example the Google Antigravity's Editor.

image


larshp:
thanks, feature and implementation looks good to me

just pretty print the code and push again to fix errors from linting, and implement the method in the testclass referred in the linter findings

src/ui/routing/zcl_abapgit_services_git.clas.abap
```diff
@@ -94,6 +94,9 @@ CLASS zcl_abapgit_services_git IMPLEMENTATION.
 
     COMMIT WORK.
 
+    zcl_abapgit_exit=>get_instance( )->validate_after_push(
+      ii_repo_online ).
```
larshp:
```suggestion
    zcl_abapgit_exit=>get_instance( )->validate_after_push( ii_repo_online ).
```

gardian12:
See also PR [#286](https://github.com/abapGit/docs.abapgit.org/pull/286) for documentation.

larshp:
perfect, thanks

PR #7524, opened 2026-01-23
move files + run abapGit via SICF

larshp:
moving some code to https://github.com/abapGit/abapGit-web-classic

removed `FOR TESTING` in `zcl_abapgit_ui_core_injector` to be able to inject the new web "control", note all other changes are in `/test`

see web related issues at https://github.com/abapGit/abapGit-web-classic/issues


mbtools:
Rendering is working and looks great! Saving in dialog like Repo Settings does not seem to work but it will get there :-)

PR #7528, opened 2026-01-24
git delta: extract offset and length to new method

larshp:
testing performance, same logic but a lot fewer method calls

also looks better from a design/single responsibility perspective

runtime before:
image

runtime after:
image

not exactly the same number of calls with the same data, but its running 10+ seconds faster, so should be good


larshp:
additional unit tests added, preparing for https://github.com/abapGit/abapGit/pull/7536

PR #7532, opened 2026-01-26
move method `remove_ignored_files` to private

larshp:
poking around doing small changes, looking into the whole repo remote/online/offline/other refactoring, #2216


larshp:
image

PR #7536, opened 2026-01-27
git delta: extract offset and length to new method

mbtools:
@larshp I always wondered if the bit comparisons (`O` and `Z`) are useful and faster. Can you run this version past your perf test?


larshp:
suggest merging https://github.com/abapGit/abapGit/pull/7528 first, then its easier to review this PR IMHO

larshp:
I'll have a look sometime, but low on time this week

larshp:
still awaiting me

larshp:
bug in code changes, not transpiler

mbtools:
But is it faster?

larshp:
next step: @larshp to test

larshp:
before:
image

after:
image

so it looks like its a tiny bit faster, maybe not worth it as I find it more difficult to understand the Z and O operators

mbtools:
thanks for testing. agree, not worth it.

PR #7538, opened 2026-01-27
DESD: Add support for Dictionary External Schema Definition

JoergCMaier:
Support ABAP Dictionary CDS type DESD: https://github.com/SAP/abap-file-formats/tree/main/file-formats/desd. 
This seems to be the first ZABAPGIT AFF type that is Dictionary and AFF and thus does not allow to write active version directly (it is required to run DDIC activation to generate/publish an active version).


mbtools:
We know it's a bit frustrating to support all the way back to 7.02...

 - Use a field-symbol for the dereferencing: `ASSIGN lr_...->* TO <lv...>.`. Then use the field-symbol. 
 - `CONV #( )` is not supported but a direct assign of the value (`= 2`) should be fine.

mbtools:
Hi Joerg,

looks like your IDE does not respect `.editorconfig`. I fixed the eof for these:

image

On my 7.58 SP1, the test repo fails since the object type is not supported. 

image

I updated the common part to correctly check for the missing handler in lower releases. Now users get the correct message:

image

What is the min release for DESD?

JoergCMaier:
Hi @mbtools ,

DESDs are supported since 2025 (816). 

As ZABAPGIT uses DDIF_STATE_GET, I fixed it to support DESDs. The ABAP correction is required in 816 to support ABAPGIT (open source): Note 3709523 - ABAP CDS: DDIF_STATE_GET returns illegal_input for objects of type DESD.
Thanks for correcting the newlines/eof. 

BR, Jörg

larshp:
thanks @JoergCMaier

mbtools:

PR #7546, opened 2026-01-31
Fix detection of text/binary files

mbtools:
CR_LF was not removed correctly and could lead to false positive i.e. some text files were considered binary (see unit test).


christianguenter2 (approved):
LGTM 👍

PR #7555, opened 2026-02-03
ENHS: downport

larshp:


mbtools:
that class exists in 702 🤷

larshp:
yea, but not the `abap_language_version` method parameter

mbtools:
Ahh, i see, it's a method, not a function 👍 

i will startup my 702 after all the merges and double-check the standalone

PR #7561, opened 2026-02-03
DDLS: Add protocol for deletions

JoergCMaier:
Any deletions of objects with type DDLS get an empty ABAP Dictionary protocol due to explicit specification of "prid = -1". This is now fixed and allows better analysis of deletions using common ABAP Dictionary support tools (e.g. PROG RADPROTA or TRAN SDDLAR).

### DIFF
--- a/src/objects/zcl_abapgit_object_ddls.clas.abap
+++ b/src/objects/zcl_abapgit_object_ddls.clas.abap
@@ -125,6 +125,23 @@ CLASS zcl_abapgit_object_ddls IMPLEMENTATION.
   ENDMETHOD.
 
 
+  METHOD get_log_uuid.
+
+    DATA lv_tstmpl       TYPE timestampl.
+    DATA lv_tstmp_string TYPE string.
+
+    TRY.
+        cl_system_uuid=>convert_uuid_x16_static( EXPORTING uuid     = cl_system_uuid=>create_uuid_x16_static( )
+                                                 IMPORTING uuid_c32 = rv_log_uuid ).
+      CATCH cx_uuid_error.
+        GET TIME STAMP FIELD lv_tstmpl.
+        lv_tstmp_string = lv_tstmpl.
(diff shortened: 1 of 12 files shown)


mbtools:
Let's avoid the PERFORM in PROGRAM. Can we use the same as in object activation?

```
          logname        = lv_logname
          write_log      = abap_true
          prid           = 1
```

https://github.com/abapGit/abapGit/blob/23058f6130f975e37f65792ec81ecbf50e7d5e29/src/objects/core/zcl_abapgit_objects_activation.clas.abap#L189-L231

JoergCMaier:
Hi @mbtools ,

I understand the external perform is not nice. In current releases, ABAP Dictionary has CL_DD_LOG* to handle Dictionary protocols in OO coding. Not available until first support of DDLS with 7.40 though.

Unfortunately, there is not ii_log available in zif_abapgit_object~delete that could be used. I see ii_log is available in zcl_abapgit_objects->delete(). Should we extend the signature of the following methods and add ii_log?

- zcl_abapgit_objects->delete_object()
- zif_abapgit_object~delete()

BR, Joerg

PS: I must admit, I wonder why the architecture does not execute the deletion of all Dictionary objects in one single call of DD_MASS_ACT_C3? For activations it seems the Dictionary objects are activated by a common call of DD_MASS_ACT_C3 in ZCL_ABAPGIT_OBJECTS_ACTIVATION->ACTIVATE_DDIC().

Edit: typo: on -> one single

mbtools:
The log is useful for other object types, too. Yes, please add ii_log to these methods. 

There are 12 years of history here 😉 . The projects haven't been very large to need mass processing for deletions. There certainly have been more DDIC/CDS object types added over the years. So we might take a look handling it like the activation.

mbtools:
The log wasn't passed in `zcl_abapgit_objects=>delete` (because it was optional). I made the log parameter mandatory.

mbtools (approved):

PR #7565, opened 2026-02-07
Support "do not ignore" files in repo settings

mbtools:
You can now add negative ignore patterns by using `!` as a prefix to allow files that are otherwise excluded:

image

The behaviour is like `.gitignore`: Patterns are read top-to-bottom with later ones overriding earlier


src/repo/zcl_abapgit_dot_abapgit.clas.abap
```diff
@@ -338,6 +338,10 @@ CLASS zcl_abapgit_dot_abapgit IMPLEMENTATION.
         rv_ignored = abap_true.
         RETURN.
       ENDIF.
+      IF lv_ignore(1) = '!' AND lv_name CP lv_ignore+1(*).
```
larshp, line 351: if `strlen( lv_ignore ) = 0` then this will crash?

src/repo/zcl_abapgit_dot_abapgit.clas.testclasses.abap
```diff
@@ -104,6 +104,16 @@ CLASS ltcl_dot_abapgit IMPLEMENTATION.
       act = lv_ignored
       exp = abap_true ).
 
+    " Don't ignore file
+    lo_dot->add_ignore( iv_path = '!' && lc_root
+                        iv_filename = 'abaplint.json' ).
```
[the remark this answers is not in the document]
mbtools: it's a concat so ok

src/repo/zcl_abapgit_dot_abapgit.clas.abap
mbtools: fixed

PR #7569, opened 2026-02-09
Call deserialize post-processing exit at end of deserializing

mbtools:
The `deserialize_postprocess` exit was only called after deserializing objects. It's now called also at the very end of the `deserialize process` with the original remote and already updated local files.


larshp:
deserialize_postprocess is called in two places now?

mbtools:
I didn't remove the first call to stay compatible and it's also per deserialize phase (is_step). The additional call at the end lets you process any other files for example like /data or things in root like abaplint.json.

larshp:
this might break some stuff, so need this in the changelog

PR #7576, opened 2026-02-11
Fix: TADIR-MASTERLANG is lost

### LINKED ISSUE #7573: Local files - `ORIGLANG` language is not filled
I'd like to render a wall message if some objects are created with the wrong master language.
While implementing `zif_abapgit_exit~wall_message_repo` I noticed that `ORIGLANG` is not filled for local files. 

E.g. 

```
    FINAL(local_objects) = zcl_abapgit_repo_srv=>get_instance(
                                              )->get( is_repo_meta-key
                                              )->get_files_local( ).
```

Image

I'd have assumed that it's filled from `TADIR-MASTERLANG`.

Image

Is there a particular reason why that's not the case?

If not I'd propose to fill it here:
https://github.com/abapGit/abapGit/blob/0743b064354dba6c66de64d4b890603839b79a8f/src/objects/core/zabapgit_parallel.fugr.z_abapgit_serialize_parallel.abap#L30-L34
https://github.com/abapGit/abapGit/blob/0743b064354dba6c66de64d4b890603839b79a8f/src/objects/core/zcl_abapgit_serialize.clas.abap#L684-L687

christianguenter2:
fixes #7573

### DIFF
--- a/src/objects/core/zabapgit_parallel.fugr.xml
+++ b/src/objects/core/zabapgit_parallel.fugr.xml
@@ -85,20 +85,8 @@
      <SHORT_TEXT>abapGit - Serialize Object</SHORT_TEXT>
      <IMPORT>
       <RSIMP>
-       <PARAMETER>IV_OBJ_TYPE</PARAMETER>
-       <TYP>TADIR-OBJECT</TYP>
-      </RSIMP>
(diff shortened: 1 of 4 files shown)


mbtools (approved):
Thank you 🙏

PR #7579, opened 2026-02-12
SRVD: downport

larshp:
ref #7577


mbtools:
The exception didn't exist in 702, indeed

image

PR #7583, opened 2026-02-13
git: extra validations when deleting branch

larshp:
so it crashes early instead of returning error 500 from server


mbtools (approved):
👍

PR #7586, opened 2026-02-19
AIFC: Add ABAP Language Version support

mbtools:
This got lost somehow... final object. Removing experimental flag is next.

Ref #6689


larshp (approved):
thanks, looks good to me

PR #7595, opened 2026-02-20
Update lock message

paul-modderman:
Change message for locks to get some distinction between transport locks and editor/user locks.


paul-modderman:
oops

PR #7596, opened 2026-02-20
Update lock message

paul-modderman:
Change editor lock message to be more meaningful and distinct. AI agents have been confusing the editor lock for a transport lock.


paul-modderman:
oops again

mbtools:
The change to the lock message makes sense. Don't give up, Paul!

PR #7597, opened 2026-02-21
Change object lock message for clarity

indigosquirrelio:
Otherwise it's confusing whether this is a transport issue, or a regular editor lock.


mbtools (approved):
Great. Thanks 👍

PR #7599, opened 2026-02-23
Fix failing unit test

christianguenter2:
Regression from #7597
image


mbtools:
One of those tests excluded from abaplint. Thanks 🙏

PR #7600, opened 2026-02-26
Update exception message for locked objects

indigosquirrelio:
This is simpler and more direct, less AI confusion.


christianguenter2:
Please update unit test.
image

See #7599

larshp:
@indigosquirrelio update the unit test?

PR #7605, opened 2026-03-02
Customizing transport support for logical objects

habrahams:
Use Case:
In the SAP Payroll space logical objects exist that have to be transported via customizing transports. These logical objects also do not have a TADIR entries. 

The objects that I have identified and that are relevant to my use case are:
PCYC - HR: Transport customer PC rules
PFRM - Form for payroll remuneration statement
PMKC - HR: Transport customer features
PSCC - HR: Transport customer schemas

For background, I currently have a piecelist(Transportable transport list) which act as a bill of materials. In the zif_abapgit_exit~change_tadir exit I modify TADIR so that these objects are processed.

This PR's purpose is to get the customising transport number to the logical object and the generic processing class.
There are some changes that have to be made to the processing in zcl_abapgit_objects_generic but I will include it in a subsequent PR.

Regards,

### DIFF
--- a/src/objects/zcl_abapgit_objects.clas.abap
+++ b/src/objects/zcl_abapgit_objects.clas.abap
@@ -867,7 +867,7 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
         iv_text    = |Step { is_step-order } - { is_step-descr }:| &&
                      | { <ls_obj>-item-obj_type } { <ls_obj>-item-obj_name }| ).
 
-      IF zcl_abapgit_factory=>get_cts_api( )->is_object_type_customizing( <ls_obj>-item-obj_type  ) = abap_true.
+      IF zcl_abapgit_factory=>get_cts_api( )->is_object_type_customizing( <ls_obj>-item-obj_type ) = abap_true.
         lv_transport = is_checks-customizing-transport.
       ELSE.
         lv_transport = is_checks-transport-transport.


habrahams:
It would seem from the abaplint errors that I will have to include some additonal changes to zcl_abapgit_objects_generic so that the added optional parameters are picked up as used.

mbtools:
hmmm. I'm not opposed to this in general but I'm wondering if this is the right approach. 

1)

Is there a way for abapGit to determine upfront whether an object type needs a workbench or customizing request? If yes, then abapGit could pass the correct request to the object class in the existing `iv_transport` parameter.

Alternatively, we could add a step id = 'CUST'. If an object class returns this step id in `get_deserialize_steps`, then we pass the customizing request.

2)

`iv_transport` should be passed to the generic class anyway (to `zcl_abapgit_objects_generic->delete/deserialize`) but we could get away with a single parameter and keep it consistent as `iv_transport`. 

3) 

It's more work, but please move the `TR_...` functions to `zcl/if_abapgit_cts_api`. We try to keep them central and have the option to mock them.

PS: Changes to these core classes are better discussed first before implementing changes ;-)

habrahams:
Thank you for your feedback Marc, I was hoping the feedback would be along these lines due to my initial PR not being an optimal solution. The reason for it not being an optimal solution was to reduce the number of changes and to facilitate this discussion.

Let me refactor the changes taking into account your feedback.

habrahams:
@mbtools, regarding your second comment:

> iv_transport should be passed to the generic class anyway (to zcl_abapgit_objects_generic->delete/deserialize) but we could get away with a single parameter and keep it consistent as iv_transport.

Should I make the parameter optional for zcl_abapgit_objects_generic->delete/deserialize or can I go ahead and make it mandatory and  update where the generic class methods are called in the below classes?
image

mbtools:
pgmid -> tadir-pgmid + some empty methods where the enhanced interface is used.

looks good otherwise. I will run CI test when I'm back from a conference next week.

src/cts/zif_abapgit_cts_api.intf.abap
```diff
@@ -195,4 +195,33 @@ INTERFACE zif_abapgit_cts_api
...
+      system_customizing          TYPE trcateg VALUE 'CUSY',
+    END OF c_obj_transport_category.
+
+  "! Get an object's transport category
+  "! @parameter iv_pgmid | Program ID / R3TR and LIMU are relevant
+  "! @parameter iv_object | Object type
+  "! @parameter rv_category | Object category. See c_obj_transport_category constants
+  METHODS get_object_transport_category
+    IMPORTING
+      !iv_pgmid          TYPE pgmid DEFAULT 'R3TR'
```
larshp:
```suggestion
      !iv_pgmid          TYPE tadir-pgmid DEFAULT 'R3TR'
```

src/cts/zif_abapgit_cts_api.intf.abap
```diff
@@ -195,4 +195,33 @@ INTERFACE zif_abapgit_cts_api
...
+    RETURNING
+      VALUE(rv_category) TYPE trcateg.
+
+  "! Check if an object is a logical object
+  "! @parameter iv_pgmid | Program ID / R3TR and LIMU are relevant
+  "! @parameter iv_object | Object type
+  "! @parameter rv_is_logical_object | True if an object is a logical object
+  METHODS is_logical_object
+    IMPORTING
+      !iv_pgmid                   TYPE pgmid DEFAULT 'R3TR'
```
larshp:
```suggestion
      !iv_pgmid                   TYPE tadir-pgmid DEFAULT 'R3TR'
```

larshp:
@habrahams check errors from CI, https://github.com/abapGit/abapGit/pull/7605/checks?check_run_id=65763516999

habrahams:
Thanks Lars, busy with them now.

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -871,9 +872,14 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
 
       lv_obj_trkorr_cat = zcl_abapgit_factory=>get_cts_api( )->get_object_transport_category( <ls_obj>-item-obj_type ).
       IF lv_obj_trkorr_cat = zif_abapgit_cts_api=>c_obj_transport_category-client_specific_customizing.
-        lv_transport = is_checks-customizing-transport.
+        IF cs_checks-customizing-transport IS INITIAL.
+          ls_transport_type-request = zif_abapgit_cts_api=>c_transport_type-cust_request.
+          ls_transport_type-task    = zif_abapgit_cts_api=>c_transport_type-cust_task.
+          cs_checks-customizing-transport = zcl_abapgit_ui_factory=>get_popups( )->popup_transport_request( ls_transport_type ).
```
mbtools: Sorry, this is a no-no. There can't be any popups in this class. User interactions have to be handled before deserialize is started (i.e. in services layer).

habrahams: Will do, thanks for the feedback!

src/cts/zif_abapgit_cts_api.intf.abap
habrahams: This was fixed

src/cts/zif_abapgit_cts_api.intf.abap
habrahams: This was fixed

src/objects/zcl_abapgit_objects.clas.abap
habrahams: I have refactored the popup logic, it now follows the same logic as the workbench popup.

mbtools:
The code looks good. It will require some testing. I will get to it after the conference next week (https://taiefy.com/)

mbtools:
Your PR is good. Thanks for the contribution. 

I refactored the CTS part a bit to get a simple interface `is_object_type_customizing`.  It also avoids running the check for each object (which could be thousands) and does it only once per object type.

habrahams:
@mbtools , thank you for the assistance and reviews!

mbtools:
Let us know if this is sufficient for integrating the other object types.

PR #7606, opened 2026-03-03
[AFF] experimental: Add DOMA (Domain) serializer

albertmink:
empty description

PR #7611, opened 2026-03-06
memory conversion of XML to AFF

larshp:
in the future, abapGit will be able to read all old formats(XML) and serialize to only the latest([AFF](https://github.com/SAP/abap-file-formats/) json)

however, a lot of old repositories exists, and AFF support will be implemented one step at a time, and new AFF versions might be introduced. The idea is to provide a github action, which automagically will suggest PRs with format updates, this PR introduces a new interface meant for this purpose.

The new interface can be used to structure the updated object type handlers better, while still supporting a transpiled github action.

Suggest merging this PR, and then adjusting https://github.com/abapGit/abapGit/pull/7606 accordingly to see how it feels/works.


larshp:
> just the bom xml 👍

done, was in progress 😅

schneidermic0:
Maybe, I misunderstood your idea. Wouldn't it be sufficient to deserialize the xml (from the "old" xml format)? Serialization would then take place by the standard serialization method of the object, wouldn't it?

larshp:
> Maybe, I misunderstood your idea. Wouldn't it be sufficient to deserialize the xml (from the "old" xml format)? Serialization would then take place by the standard serialization method of the object, wouldn't it?

thanks, good catch

I've added a diagram to get an overview, it looks a bit more complicated than in my head, but should be simple as such,

image

mbtools:
Do we really need this automation and the "to/from memory" thing? 

Each repo owner will decide when to switch and what objects to switch from XML to AFF JSON format. It's the same as today with changes to the XML format: Update AG to latest version, open your repo, push the change. It's a one way from ABAP to repo.

PRs that magically appear in GH will cause more confusion or conflicts with pending local changes than being helpful. There could also be a difference between capabilities of the GH action and the AG the repo owner has running in his system. 

1. Deserialize

AG needs to be able to detect if an object is XML or JSON and call either the old class or AFF. That exists [already](https://github.com/abapGit/abapGit/blob/0b49a343879968fa174ed8b18bae26ce71039c55/src/objects/zcl_abapgit_objects.clas.abap#L725).

2. Serialize

Today, there's no choice and you always get the latest format. The question is whether we want to introduce an option:
- Choose AFF per repo 
- Choose AFF per object type
- Keep as is (current AFF switch = on: once an object type is AFF, you always get JSON)

schneidermic0:
Thanks for the diagram. In the diagram, I only see the deserialize method of the memory interface (`deserialize_to_memory`). The other `serialize` method is understand as the standard (existing) serialize method, isn't it?

schneidermic0:
Sorry, I just saw, you have already deleted the method for serialization.

larshp:
changed to draft, will add more details later

larshp:
closing

I think we can do it without, it will keep it simple. My toolbox has improved after first considering this feature

PR #7617, opened 2026-03-09
Add ADS check to SFPF Object Class

habrahams:
SFPF Objects depend on the ADS service (Adobe Document Service).
If this service is not available or configured,  SFPF objects should not be considered when importing an offline repository.

The check implemented in the constructor follows the same approach as the SAP standard test program `FP_PDF_TEST_00`.

### DIFF
--- a/src/objects/zcl_abapgit_object_sfpf.clas.abap
+++ b/src/objects/zcl_abapgit_object_sfpf.clas.abap
@@ -15,7 +15,7 @@ CLASS zcl_abapgit_object_sfpf DEFINITION
         !io_files       TYPE REF TO zcl_abapgit_objects_files OPTIONAL
         !io_i18n_params TYPE REF TO zcl_abapgit_i18n_params OPTIONAL
       RAISING
-        zcx_abapgit_exception.
+        zcx_abapgit_type_not_supported.
 
     CLASS-METHODS fix_oref
       IMPORTING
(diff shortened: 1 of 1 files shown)


src/objects/zcl_abapgit_object_sfpf.clas.abap
```diff
@@ -8,6 +8,15 @@ CLASS zcl_abapgit_object_sfpf DEFINITION
...
     INTERFACES zif_abapgit_object .
 
+    METHODS constructor
+      IMPORTING
+        !is_item        TYPE zif_abapgit_definitions=>ty_item
+        !iv_language    TYPE spras
+        !io_files       TYPE REF TO zcl_abapgit_objects_files OPTIONAL
+        !io_i18n_params TYPE REF TO zcl_abapgit_i18n_params OPTIONAL
+      RAISING
+        zcx_abapgit_exception.
```
christianguenter2: can you use `zcx_abapgit_type_not_supported` instead?

habrahams: Yes, of course, thanks Christian.
FYI I copied from the example @ [serializers.html](https://docs.abapgit.org/development-guide/serializers/serializers.html#constructor)
I'll push another commit shortly.

christianguenter2: Thx for pointing out. Documentation is outdated.
[#290](https://github.com/abapGit/docs.abapgit.org/pull/290)

larshp:
image

I got this popup after updating abapGit, when opening the abapGit repository

larshp:
@habrahams any thoughts on moving this check into the deserialize method? Looks like I can work on and show SFP objects when the connection is broken

mbtools:
@larshp looks like you have a partially configured `ADS` RFC destination. It shouldn't exist if ADS isn't used in the system or it should have the correct login.

larshp:
> @larshp looks like you have a partially configured `ADS` RFC destination. It shouldn't exist if ADS isn't used in the system or it should have the correct login.

yea, I guess the configuration is bad. But I cannot change any of it in the particular system

mbtools:
then indeed, the check should be at deserialize

larshp:
https://github.com/abapGit/abapGit/pull/7620 opened

habrahams:
> @habrahams any thoughts on moving this check into the deserialize method? Looks like I can work on and show SFP objects when the connection is broken

@larshp , when you say you can work on the object, do you mean via tcode SFP where you can maintain the context and properties? I believe this should be possible without a ADS connection, but you will not be able to activate the form or edit it via the layout tab in Adobe Livecycle designer.

I do not have an issue moving the check to the deserialize method.

PR #7627, opened 2026-03-16
suppress modification assistant popup

habrahams:
Use case and scenario:

Suppress the modification assistant popup for objects that are in a namespace other than Z or Y. The namespace in question is also only configured with a repair or recipient role.

I believe that this choice should be determined by setting a boolean value via a user exit, but I thought I would get some feedback first before adding an additional user exit method.

### DIFF
--- a/src/objects/oo/zcl_abapgit_oo_class.clas.abap
+++ b/src/objects/oo/zcl_abapgit_oo_class.clas.abap
@@ -500,7 +500,7 @@ CLASS zcl_abapgit_oo_class IMPLEMENTATION.
 
 * this will update the SEO* database tables
     TRY.
-        lo_update->revert_scan_result( ).
+        lo_update->revert_scan_result( abap_true ).
       CATCH cx_oo_source_save_failure INTO lx_error.
         zcx_abapgit_exception=>raise_with_text( lx_error ).
     ENDTRY.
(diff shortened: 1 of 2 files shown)


larshp:
if its configured correctly changes are allowed to the object, then there should be no assistant popup?

habrahams:
If the namespace is configured as Producer, then the popup is not displayed, but I intentionally do not want to provide the target system with the Producer Key, they will however always have the repair key configured.

larshp:
then I guess abapGit will have different behavior compared to editing the object manually in SE24/ADT ? ie. skipping the modification assistant popup

habrahams:
Yes, that's why I think it would be best to make it a manual override via a user exit. Similar to allowing SAP objects and SAP TABU entries.

larshp:
yea, a user exit is okay, we want to keep the default behavior close to what the developer usually experience

mbtools:
The code change by passes important auth checks. That's not the right way.

I have been using abapGit with namespaces for years but the Modification Assistant never shows up. 

When you import a repo with a namespace that has only a consumer key, objects are assigned source system = 'SAP' in TADIR. This means you can't modify them. This is intended and correct in my view.

You can change the TADIR entries in transaction SE03 to your own source system (sy-sysid). Or you use a repo setting to define how the source system should be set during import:

image

Can you try these options?

If this isn't sufficient or not working as described, we need a reproducible example.

habrahams:
Ok, let me look into this and get back to you....

mbtools:
I just updated the docs https://github.com/abapGit/docs.abapgit.org/pull/291

habrahams:
I'm going to close this PR for the moment. I will revisit this as soon as I am able.

Issue #6568, opened 2023-10-24
SAP GUI for HTML: Staging


mbtools:
After selecting files to stage, SAP GUI for HTML raises the following error message:

![image](https://github.com/abapGit/abapGit/assets/59966492/c63dfcfe-282b-4c79-8390-14ef43b48665)

```text
Received "[1006] at least one batch step has failed" on batch step #1: {"content":"rg_post_data.frame_name=itshtmlvwrfnC116&rg_post_data.sap_event=&/src/0242ac1100021eebb1bbc829fae4fe35.smim.abapgit_ci_result.html=?&/src/0242ac1100021eebb1bbc829fae4be35.smim.abapgit_ci_result.json=A","post":"action/401"}
```

SAP GUI for HTML:

![image](https://github.com/abapGit/abapGit/assets/59966492/605bfc3e-7da2-4ed1-bb3a-be8abd83dcea)

SAP GUI for Windows:

![image](https://github.com/abapGit/abapGit/assets/59966492/59038cd8-c2f2-419b-b0db-d6427e27e81c)

Cause:

A `a` element is used but `href` points to `#` and is handled by JS. This does not work for HTML GUI.

JS:

https://github.com/abapGit/abapGit/blob/d840aceefa0f0be6f060f8038d3fc252ed0ba08f/src/ui/zabapgit_js_common.w3mi.data.js#L780-L783

PR #7628, opened 2026-03-16
SAP GUI for HTML: Fix staging

mbtools:
That one was harder but I declare victory over SAP GUI for HTML! It's now a viable alternative :-)

Closes https://github.com/abapGit/abapGit/issues/6568


christianguenter2:
P.S. final boss `patch` is still around ;-)

mbtools:
Ah, you're right. I will take a look at patch

PR #7634, opened 2026-03-19
Repo Overview/Repo View: Enhanced repo settings

christianguenter2:
With this change all repo settings pages are directly accessible in repo overview and repo view. 

:white_check_mark: Allows faster navigation to Local/Remote etc., especially with the command palette.

:warning: Navigation to repository settings with mouse may slow down a bit, as one first needs to hoover and then click

Imho a tradeoff worthwhile.

Repo overview:
image

Repo view:
image


mbtools (approved):
We can also get the settings clicking the icon after the url. Good idea for the menu!

PR #7637, opened 2026-03-22
make the inactive icon non-selectable

larshp:
making it easier to copy pasta the object name


larshp:
hm, doesnt seem to work

PR #7649, opened 2026-03-30
ZCL_ABAPGIT_FACTORY: Harmonize BOUND check

christianguenter2:


mbtools (approved):
👍

larshp:
hmm, any guide/recommendations somewhere if/when to use INITIAL/BOUND ? I typically only use INITIAL

christianguenter2:
I thought there were but couldn't find any.
I prefer to use with `BOUND` when working with (object-)references. Which makes the latter a bit more obvious.
Main purpose of this pr is to harmonize the condition within the class, where `BOUND' is used majorily making it a bit easier to read. Should have made that a bit more obvious.

larshp:
PR is fine, just wondering about INITIAL/BOUND if I should add a rule to abaplint

christianguenter2:
> PR is fine, just wondering about INITIAL/BOUND if I should add a rule to abaplint

I'd like that. There has a been a discussion in the code-pal repo. There are cases where `IS INITIAL` and `IS NOT BOUND` are not equivalent. So `BOUND` makes it a bit less ambigious imho.
https://github.com/SAP/code-pal-for-abap/issues/472

Issue #7661, opened 2026-04-09
zlib: abapGit (the latest) fails to serialize [regression]


sbcgua:
At commit: 6a3699ae56c3071dcf4c04f865983d9883f5ec02

<img width="345" height="278" alt="Image" src="https://github.com/user-attachments/assets/fda3e9a5-ea81-48c9-97fa-05ad133b214c" />

larshp:
so this was the zlib change?

larshp:
reverting

each change was quite simple, but I guess there is some edge case

sbcgua:
Yeap, works after the reversal. And #7662 too ...

larshp:
hmm, okay, will do some additional checking and testing

sbcgua:
Closing, transformed and moved to #7668

PR #7664, opened 2026-04-10
zlib: try again

larshp:
original = https://github.com/abapGit/abapGit/pull/7657
reverted = https://github.com/abapGit/abapGit/pull/7663
issue = https://github.com/abapGit/abapGit/issues/7661

### DIFF
--- a/src/git/zlib/zcl_abapgit_zlib.clas.abap
+++ b/src/git/zlib/zcl_abapgit_zlib.clas.abap
@@ -80,28 +80,23 @@ CLASS zcl_abapgit_zlib IMPLEMENTATION.
 
   METHOD decode.
 
-    DATA: lv_bit   TYPE c LENGTH 1,
-          lv_len   TYPE i,
-          lv_count TYPE i,
+    DATA: lv_count TYPE i,
           lv_code  TYPE i,
           lv_index TYPE i,
           lv_first TYPE i.
 
 
     DO zcl_abapgit_zlib_huffman=>c_maxbits TIMES.
-      lv_len = sy-index.
+      lv_count = io_huffman->get_count( sy-index ).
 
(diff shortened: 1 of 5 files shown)


mbtools:
take_bit is probably not the same as take_bits(1). There's a clear mv_bits under certain conditions.

larshp:
> take_bit is probably not the same as take_bits(1). There's a clear mv_bits under certain conditions.

hmm, yea, perhaps the unit test should also check `mv_bits`

anyhow I've asked Opus, so lets see

UPDATE:
UPDATE UPDATE: it got below from above comment, so it didnt figure it out
image
but it doesnt seem like I'm able to reproduce

larshp:
hmm, it still insists,

image

larshp:
@sbcgua can you try outputting the endianess from your system? see #7666

and also pull this branch to see if it still has the problem

and run all unit tests in this branch to see the result of those

mbtools:
All tests pass on my box with (Charsize:  2, Endian:  L) but that's probably what you have.

mbtools:
Also tested with https://gitlab.com/sbu-absw/abaplint-example and https://gitlab.com/fernandofurtado/abap-markdown which call `zlib_decompress`. works fine 👍 

Codex says there's nothing endian-specific in the code and I agree. It's a bit of a mystery how it could fail but work without this change. 🤷

sbcgua:
Endianness: the system runs on Intel (BTW in 7666 the screenshot show big-endian??? is it another architecture?)
image

sbcgua:
OK, so this works (as well as probably the previous version), the bug redescribed in #7668 and seems to be unconnected to zlib

larshp:
gonna merge this, the code has already been reviewed in https://github.com/abapGit/abapGit/pull/7657

PR #7667, opened 2026-04-10
DTEL + TABL: fix exists() for identical names

larshp:
If a table in remote has the same name as a data element in the system, the exists() method returns true, but it doesnt exist, its a data element

small test program to verify the DTEL:

```abap
DATA ls_x030l TYPE x030l.

SELECT * FROM dd04l INTO TABLE @DATA(tab) UP TO 10000 ROWS.

LOOP AT tab INTO DATA(row).
  CALL FUNCTION 'DD_GET_NAMETAB_HEADER'
    EXPORTING
      tabname   = row-rollname
    IMPORTING
      x030l_wa  = ls_x030l
    EXCEPTIONS
      not_found = 1
      OTHERS    = 2.
  IF ls_x030l-tabtype <> 'E'.
    WRITE / ls_x030l-tabtype.
  ENDIF.
ENDLOOP.

WRITE / 'Done'.
```

### DIFF
--- a/src/objects/tabl/zcl_abapgit_object_tabl.clas.abap
+++ b/src/objects/tabl/zcl_abapgit_object_tabl.clas.abap
@@ -81,7 +81,7 @@ ENDCLASS.
 
 
 
-CLASS ZCL_ABAPGIT_OBJECT_TABL IMPLEMENTATION.
+CLASS zcl_abapgit_object_tabl IMPLEMENTATION.
 
 
   METHOD clear_dd03p_fields.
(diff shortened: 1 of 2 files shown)


mbtools:
tabtype has a bunch of other values <> 'E' that don't look like TABL:

image

Here's the mapping:

```abap
   data(lv_tabclass) = cl_dd_nametab_internal=>get_ddobjtype( iv_tabtype = ls_header-tabtype
                                                               iv_tabform = ls_header-tabform ).
```

Maybe use `DDIF_NAMETAB_GET` which returns `DDOBJTYPE` instead but it might be a bit slower.

larshp:
hmm, I'm a bit afraid to refactor it too much, to avoid errors and performance regressions

but well, might be good, guess there are more bugs in this code, but I dont want to spend much time testing it 😅

mbtools (approved):
Just rare name conflicts or rare object types. Merge 😁

PR #7674, opened 2026-04-23
new user exit: change username in password popup

larshp:
when the user specific value is not enough


mbtools:
I usually do this via create_http_client but looks ok to me.

mbtools (approved):
👍

PR #7676, opened 2026-04-23
flow: performance optimization through visit buffering

fdupret:
Trying to use flow, I'm facing terrible performance with a relatively small repository. 
I have 118 commits, 331 objects returned by list_no_blobs_multi, yet abapgit is decoding 4M commits in 1min30s. It's decoding the same commits over and over. I guess our git commit history did not follow git best practices at some point and commits are more intertwined than in most repos?

I tested this PR with abap2xlsx and went from an 11min timeout to... another timeout but further down the method (in the features loop). I guess there is still room for improvement?
I also tested this PR with our repos, and it went from 1min30s to 8s. 
However I'm still new to flow and not well-versed in the Git protocol itself, so please review this PR with care.


larshp (changes requested):
I'll have a look

larshp (approved):
thanks, looks good, I'll do a follow up PR to refactor the existing method a bit

PR #7682, opened 2026-04-30
SMTG: add AFF support

OliverBeck01:
Enable object type SMTG for the AFF format.
The new object handler can pull the old XML and new AFF format of objects with abap object type SMTG but will only push objects with format AFF.


larshp:
Hi,
As this is an existing object type, we'd like it to be backwards compatible, ie. reading the old XML files should work and then serialize to AFF json

OliverBeck01:
> Hi, As this is an existing object type, we'd like it to be backwards compatible, ie. reading the old XML files should work and then serialize to AFF json

Hello Larshp,
the new object handler is backwards compatible with XML format.

OliverBeck01:
> Hi, As this is an existing object type, we'd like it to be backwards compatible, ie. reading the old XML files should work and then serialize to AFF json

Hello Lars,
do you know by any chance were we can upload a new testobject for the AFF format?
thank you

larshp:
https://github.com/abapGit-tests/SMTG-aff created

src/objects/zcl_abapgit_object_smtg.clas.abap
```diff
@@ -343,105 +235,57 @@ CLASS zcl_abapgit_object_smtg IMPLEMENTATION.
...
-    RETURN.
-  ENDMETHOD.
-
-
-  METHOD zif_abapgit_object~get_deserialize_order.
-    RETURN.
-  ENDMETHOD.
-
-
-  METHOD zif_abapgit_object~get_deserialize_steps.
```
mbtools, line 373: xml deserialize depends on `get_deserialize_steps` and `get_metadata`.

mbtools (changes requested):
1. `SMTG` must be added to `ZCL_ABAPGIT_AFF_REGISTRY`

2. Trying to pull the XML version into a system that does not include the SMTG AFF class (yet), leads to an error: 

Image

The constructor must not fail in such case but fall back to the previous behavior.

Please test with https://github.com/abapGit-tests/SMTG

src/objects/zcl_abapgit_object_smtg.clas.abap
OliverBeck01: Method get_metadata( ) is called via zcl_abapgit_object_common_aff. I will check on method get_deserialize_steps()

mbtools:
Thanks. Pulling the old XML format works 👍 

Could you please add a test case to https://github.com/abapGit-tests/SMTG-aff for AFF?

OliverBeck01:
> Thanks. Pulling the old XML format works 👍
> 
> Could you please add a test case to https://github.com/abapGit-tests/SMTG-aff for AFF?

I added a test case, but the pull request need to be merged with the SMTG-AFF main branch.

OliverBeck01:
Added the method zif_abapgit_object~get_deserialize_steps, which will now provide deserialization steps for both AFF and XML format of email template obejcts according to https://docs.abapgit.org/development-guide/serializers/serializers.html#deserialize-process

mbtools:
Hi Oliver, 

I tested it again but it fails in systems that do not include the SMTG AFF handler class (in `zcl_abapgit_object_common_aff=>constructor`). 

We do not have any other productive case which supports both XML and AFF JSON formats. There are `DOMA` and `INTF` which have experimental status if you want to look at them. However, these classes use the experimental flag to distinguish between XML and JSON which won't work productively. Deserialize will always have to support both formats. 

We will have to discuss this separately until we can get this PR through. 

https://github.com/abapGit/abapGit/issues/7768

larshp:
@OliverBeck01 do you mind if I try opening a new PR with this feature? I suggest
1: read the XML into AFF intf
2: deserialize based on the AFF data
3: serialize to AFF

larshp:
@OliverBeck01 https://github.com/SAP/abap-file-formats/tree/main/file-formats/smtg doesnt seem to have a html, like you have added to https://github.com/abapGit-tests/SMTG-aff/tree/main/src ?

larshp:
@OliverBeck01 see https://github.com/SAP/abap-file-formats/issues/797

PR #7683, opened 2026-04-30
BDEF: check for inactive in exists()

larshp:
ref https://docs.abapgit.org/development-guide/serializers/serializers.html#interface


mbtools (approved):
good find 👍

YahorNovik:
Hi guys, here is the summury of the issue:
On newer SAP releases, CL_BDEF_OBJECT_PERSIST=>SAVE_METADATA only writes the inactive REPOSRC entry during CREATE, the active entry that older releases also wrote was intentionally removed by SAP. Our exists() method only checks the active version (p_version = 'A'), so after a failed activation leaves an inactive-only BDEF, subsequent syncs don't find it, take the CREATE path, and fail because the SAP operator correctly detects the existing inactive object and rejects the duplicate creation.

PR #7686, opened 2026-05-04
Avoid dumps during self-update of abapGit

mbtools:
Updating certain classes and interfaces of abapGit using the develop version of abapGit will lead to dumps. Such crashes can leave abapGit in an inconsistent state. This change introduces a check to prevent such situation for some common scenarios.


larshp (approved):
yea, well, okay

PR #7688, opened 2026-05-04
Add inactive version fallback to exists() for all affected object handlers

YahorNovik:
Extends the inactive version check from #7683 (BDEF only) to all other
object handlers that only check active version ('A').

On newer SAP releases, save_metadata only writes the inactive REPOSRC
entry during CREATE. Our exists() methods only checked p_version = 'A',
so after a failed activation leaves an inactive-only object, subsequent
syncs take the CREATE path and fail.

Affected handlers: AMSD, DRUL, DTDC, SAXX_SUPER, SKTD, SRFC
Already correct: SRVB (checks 'I' first), SRVD (checks 'A' then 'I')
SRFC also gets a null-reference guard for unbound persistence object.


mbtools (approved):
LGTM 👍

larshp:

PR #7689, opened 2026-05-04
SICF: Preserve original name

mbtools:
Closes #7652


mbtools:
Tested in https://github.com/abapGit/abapGit/issues/7652

PR #7695, opened 2026-05-07
fix(oo): Surface detailed syntax errors from CLAS/INTF check

YahorNovik:
## Problem

When a class or interface fails `syntax_check`, the exception only says:

> Class ZCL_FOO has syntax errors

The FM `SEO_CLASS_CHECK_CLASSPOOL` actually computes detailed errors (e.g. `Method "LV+" is unknown or PROTECTED or PRIVATE.` at line 14) but nobody reads them — the `wb_checklist` parameter is never passed.

Meanwhile, `zcx_abapgit_exception=>raise()` accepts `ii_log` and stores it in `mi_log`, but `add_exception` never drains it. 

## How to reproduce

1. Create a class with a syntax error (e.g. call a non-existent method)
2. Add it to a repo tracked by abapGit
3. Pull/deserialize — the log shows only "has syntax errors", not *what* the error is

### Part 1: `zcl_abapgit_log->add_exception`

When the caught exception is `zcx_abapgit_exception` with a bound `mi_log`, drain all inner log messages into the outer log before logging the exception text. Uses `TRY/?=/CATCH` for v702 compat.

### Part 2: `zcl_abapgit_oo_class` / `zcl_abapgit_oo_interface` — `syntax_check`

- Pass a `cl_wb_checklist` to the FM
- Extract detailed errors from the checklist
- Build a local log and raise with `ii_log` attached

## How to test

1. Add https://github.com/YahorNovik/abapgit-test-syntax-error as an online repo in abapGit 
2. Pull — the class `ZCL_TEST_SYNTAX_ERR` has an intentional syntax error (`me->nonexistent_method( )`)
3. After pull, check the log:
   - **Before fix:** log shows `Class ZCL_TEST_SYNTAX_ERR has syntax errors`
   - **After fix:** log also shows `Method "NONEXISTENT_METHOD" is unknown or PROTECTED or PRIVATE.`
4. Alternatively, call `syntax_check` directly:
   ```abap
   DATA lo_fnc TYPE REF TO zif_abapgit_oo_object_fnc.
   lo_fnc = zcl_abapgit_oo_factory=>get_by_name( 'ZCL_TEST_SYNTAX_ERR' ).
   lo_fnc->syntax_check( 'ZCL_TEST_SYNTAX_ERR' ).
   ```
5. Any catch site using `ii_log->add_exception( lx )` now surfaces the detail automatically

### DIFF
--- a/src/objects/oo/zcl_abapgit_oo_class.clas.abap
+++ b/src/objects/oo/zcl_abapgit_oo_class.clas.abap
@@ -978,26 +978,52 @@ CLASS zcl_abapgit_oo_class IMPLEMENTATION.
   METHOD zif_abapgit_oo_object_fnc~syntax_check.
     DATA:
       ls_clskey      TYPE seoclskey,
-      lv_syntaxerror TYPE abap_bool.
+      lv_syntaxerror TYPE abap_bool,
+      lo_checklist   TYPE REF TO cl_wb_checklist,
+      lt_errors      TYPE swbme_error_tab,
+      lo_local_log   TYPE REF TO zif_abapgit_log,
+      ls_item        TYPE zif_abapgit_definitions=>ty_item.
+
+    FIELD-SYMBOLS:
+      <ls_err>  TYPE swbme_error_entry,
+      <lv_line> TYPE string.
 
     ls_clskey-clsname = to_upper( iv_object_name ).
+
+    CREATE OBJECT lo_checklist.
 
     CALL FUNCTION 'SEO_CLASS_CHECK_CLASSPOOL'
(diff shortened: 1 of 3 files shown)


larshp (changes requested):
syntax_check method is only related to getting the version

Image

mbtools:
I still like the idea. But showing all syntax errors of a class isn't that helpful. You will need to edit the class where you can see and fix all errors, anyway.  There's a menu item to run syntax ATC. 

To keep it simple, just get the first error text and append it to the existing exception text or pass it as a long text in the raise.

PR #7696, opened 2026-05-07
DTSC support (AFF, DTSC properties file extension change)

krzykoba:
Dear Colleagues,

in order to support the migration scenario from PrivateCloud to PublicCloud (PUSH to Repo via ZABAGIT and PULL into ABAP Steampunk system), we decided to adopt the DTSC properties file and changed it from .xml to .json. The relevant implementation has been adopted and DTSC has been added in ZCL_ABAPGIT_AFF_REGISTRY. The ZABAPGIT is now also coherent with https://github.com/SAP/abap-file-formats
The repository with the test objects has been updated as well.

Best,
Krzysztof

PR #7698, opened 2026-05-08
DTSC: Switch to ABAP file format

krzykoba:
Now for up-to-date branch

in order to support the migration scenario from PrivateCloud to PublicCloud (PUSH to Repo via ZABAGIT and PULL into ABAP Steampunk system), we decided to adopt the DTSC properties file and changed it from .xml to .json. The relevant implementation has been adopted and DTSC has been added in ZCL_ABAPGIT_AFF_REGISTRY. The ZABAPGIT is now also coherent with https://github.com/SAP/abap-file-formats
The repository with the test objects has been updated as well.

Best,
Krzysztof

### DIFF
--- a/src/objects/aff/zcl_abapgit_aff_registry.clas.abap
+++ b/src/objects/aff/zcl_abapgit_aff_registry.clas.abap
@@ -73,6 +73,7 @@ CLASS ZCL_ABAPGIT_AFF_REGISTRY IMPLEMENTATION.
     register( 'UIAD' ).
     register( 'UIPG' ).
     register( 'UIST' ).
+    register( 'DTSC' ).
   ENDMETHOD.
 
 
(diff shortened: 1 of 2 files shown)


krzykoba:
this replaces PR https://github.com/abapGit/abapGit/pull/7696

mbtools:
Looks good. Do we have to be compatible with repos that already contain the xml version? Meaning still be able to read and deserialize the old object format?

krzykoba:
> Looks good. Do we have to be compatible with repos that already contain the xml version? Meaning still be able to read and deserialize the old object format?

No, we do not have to be backwards compatible. The first version of DTSC in ZABAPGIT is only 3-4 days old and we do not have any customer actively using the Dictionary Object DTSC. abapgit in Steampunk is not yet supported so we do not expect any repositories pushed from an abap Steampunk system either.

Have a nice weekend!

mbtools (approved):
Alright. I can't test it because I don't have access to a new enough release but lgtm 👍

krzykoba:
> Alright. I can't test it because I don't have access to a new enough release but lgtm 👍

Great! I've tested the scenarios in our Steampunk and Private Cloud landscape.

PR #7699, opened 2026-05-10
 js common: fix small bugs and dead branches

christianguenter2:
- StageHelper: STATUS.isValid was inverted; rename to isInvalid
  - CheckListWrapper.onClick: use the `e` parameter, not the global `event`
  - LinkHints: compare input.type to lowercase "hidden" so hidden inputs are actually skipped
  - KeyNavigation: Spacebar is " ", not ""; fixes dead Enter/Space path
  - CheckListWrapper: simplify `=== "X" ? false : true` to `!== "X"`
  - Diff filter "Only my changes": drop redundant ternaries inside `state === true` branches in onFilterOnlyMyChanges and applyOnlyMyChangesFilter


mbtools (approved):
nice catch on the space key and hidden issue. LGTM 👍

PR #7702, opened 2026-05-12
FUGR: bugfix serialization inactive function module

### LINKED ISSUE #7701: Serialization issue: inactive function modules
Steps to reproduce:

1: clone https://github.com/larshp/FUGR4
2: change ZFUGR4_FM1 to be inactive
3: serialize

Image

Image

in the files, `ZFUGR4_FM1` is now considered deleted, and include `lzfugr4u01` added, expect the function module include to be only changed

I'm having a look into fixing this right now

larshp:
closes #7701 <- see this for how to reproduce

regression from https://github.com/abapGit/abapGit/pull/7147

after the fix, which is the same behavior as PROG, the active version is serialized, and the object is marked as inactive,

image


marcellourbani (approved):
LGTM

samuely4ng (approved):
lgtm

mbtools (approved):
👍

PR #7704, opened 2026-05-14
fix: use RPY_INCLUDE_UPDATE to prevent empty active REPOSRC on failed activation

### LINKED ISSUE #7703: RPY_PROGRAM_UPDATE empties active REPOSRC when save_inactive='I' and activation fails
## Problem

`update_program` calls `RPY_PROGRAM_UPDATE` with `save_inactive='I'`. 
Internally, SAP writes an empty placeholder to the active REPOSRC row. 
Activation is supposed to promote I→A, but when activation fails, 
the active version stays empty — SE80/ADT shows blank source.

## Root Cause

`RPY_PROGRAM_UPDATE` with `save_inactive='I'` executes:

    "write always empty active version
    INSERT REPORT program_name FROM l_source_extended STATE srext_active.

where `l_source_extended` is uninitialized (empty).

## Proposed Fix

- Replace `RPY_PROGRAM_UPDATE` with `RPY_INCLUDE_UPDATE` in `update_program`
- Hardcode `save_inactive = c_state-inactive` (always write inactive)
- Remove the `iv_state` parameter (no caller needs active writes)
- Remove `iv_state = ''` from `deserialize_exit_include` callers

`RPY_INCLUDE_UPDATE` does not empty the active row when writing inactive.

PR to follow.

YahorNovik:
[… activation]

Fixes #7703


larshp:
tested ok, active version is preserved if clicking cancel in the activation popup

image

mbtools:
I have seen these symptoms before a few times. Thanks for tracking down the root cause of this issue. 

Does it work the same in 7.02?

larshp:
@mbtools is it possible for you to help testing in 702? you had a system running?

mbtools:
Will do tomorrow

mbtools:
Note: This is also used for deserializing includes of function groups. 

I tested a few scenarios and it's working ok.

PR #7708, opened 2026-05-19
Downport: fix 702 compatibility

mbtools:
- `assert_that` does not exist in 7.02
- missing table parameter in LXE get list


larshp (approved):
thanks

PR #7710, opened 2026-05-20
Fix LXE PO files for packages

mbtools:
- Avoid package name in comments of po file
- Add serialize/deserialize of local packages which are not supported by LXE (since there's no TADIR entry)

Tests:
https://github.com/abapGit-tests/PROG_lxe
https://github.com/abapGit-tests/FUGR_lxe


sbcgua (approved):
Looks good to me. Didn't test very deeply but our main product serializes correctly. Although I suppress comments there in real usage. But if I activate it back all seems correct (the PACKAGE in comments).

P.S. in general I think "suppress comments" behaviour should be adjusted. First that flag (supress LXE comments) should go to .abapgit.xml and stop being local - it is a rule that should travel with the repo (my design mistake). And also maybe it should be enabled by default. Anyway, it is a separate story

PR #7713, opened 2026-05-21
FUGS: fix "LXZAGF00 may not have an inactive version"

mbtools:
Error "LXZAGF00 may not have an inactive version" when importing exit function groups (`FUGS`). 

Regression #7704

Test: https://github.com/abapGit-tests/FUGS


larshp (approved):
thanks

PR #7716, opened 2026-05-22
SUSH: fix error when using transportable package

mbtools:
Error: "You cannot use request <space>" when trying to import SUSH object to transportable package

Fixed by passing transport request

Test: https://github.com/abapGit-tests/FUGR_rfc_with_scope with Z-package


larshp (approved):
lgtm

PR #7724, opened 2026-05-30
Page Stage: Fix add filtered and Commit

christianguenter2:
Fix: "Add Filtered and Commit" now stages all filtered files, not just local ones

Problem

  On the Stage page, applying a filter (e.g. developer) showed an accurate count in the action link — "Add Filtered and Commit (13)" — but clicking it only staged the 2 local
  files. The 11 visible rows under "Files to remove or non-code" were silently dropped, so the number of staged files never matched the count shown.

E.g.

image
image

  Cause

  StageHelper.markVisiblesAsAdded only marked visible local rows as add and reset everything else, including the visible remote rows that should be staged for removal.

  Fix

  markVisiblesAsAdded now stages every visible row according to its type:
  - visible local rows → add
  - visible remote rows ("to remove or non-code") → remove
  - non-visible rows → reset

  This mirrors the existing per-row behavior (clicking remove on a single remote row produces the same R status), so the bulk action and the displayed counter are now
  consistent.

E.g.
image

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -817,8 +817,9 @@ StageHelper.prototype.collectData = function() {
 StageHelper.prototype.markVisiblesAsAdded = function() {
   this.iterateStageTab(false, function(row) {
     // TODO refactor, unify updateRow logic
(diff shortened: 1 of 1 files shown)


mbtools (approved):
LGTM 👍

PR #7725, opened 2026-05-30
Overwrite Popup: Abort if nothing was selected

christianguenter2:
Before this change the whole deserialization process was executed although nothing was done. Including abapGit restart in case of abapGit repo.


mbtools (approved):
👍

PR #7726, opened 2026-05-30
test remote

hervianz:
test remote


mbtools:
I had a look at the comments in the code (maybe cut&paste that into the PR description). It looks like a useful tool but tbh I don't see this as a feature of abapGit. The main reason is the limited support for object types or objects that don't match the tadir level. RFC connections into production also are a bit problematic even if read-only. 

My suggestion would be to publish this as a standalone report instead.

hervianz:
Ehh i dont want to create PR 😂 my bad.. 

I'm still doing the testing, thanks for the input.. will do this in summer holiday 😁

PR #7727, opened 2026-05-31
COMMON.JS: Small fixes

christianguenter2:
1. Fix implicit global range in DiffColumnSelection.mousedownEventListener
2. Remove stale /* exported */ directives
3. Use strict equality in getSelectedText
4. change `const` to `var`
5. use `window.addEventListener` instead of overwrite
6. Fix memorizeScrollPosition
7. improve debugOuput


mbtools (approved):
lgtm 👍

PR #7730, opened 2026-06-02
LONGTEXTS: Clear DOKIL-DOKSTATE on serialize (#7729)

### LINKED ISSUE #7729: LONGTEXTS: DOKIL-DOKSTATE leaks into serialized XML for non-MSAG objects
### Summary
`DOKIL-DOKSTATE` is serialized into the XML for objects that use `zcl_abapgit_longtexts->read` (FUGR, PROG, etc.), even though abapGit intentionally treats this field as a non-portable runtime field. The deserializer correctly ignores it (`DOCU_UPDATE` is called with a hardcoded `state = 'A'`), which means the XML carries a value that is never honored — resulting in a permanent Diff between repo and target system that no pull can resolve.

For MSAG, this was already addressed by clearing the field on serialize (see PR #2605). The same pattern is missing from the shared longtexts class that other object types use.

### Steps to reproduce
1. Push an object whose longtext has `DOKIL-DOKSTATE` populated (e.g. `'R'`). The XML contains:
   ```xml
   <DOKIL>
     <LANGU>E</LANGU>
     ...
     <DOKSTATE>R</DOKSTATE>
   </DOKIL>
   ```

2. Pull the repo on any target system.

3. On the target, `DOKIL-DOKSTATE` is `'A'` (set by `DOCU_UPDATE`), while the repo XML carries the original value. Diff keeps showing the difference; subsequent pulls don't reconcile it.

### Root cause
`src/objects/texts/zcl_abapgit_longtexts.clas.abap`, method `read`. The `iv_clear_fields` block strips volatile fields from `HEAD` but leaves `DOKIL-DOKSTATE` populated, so it ends up in the XML. The deserializer in the same class then ignores it and writes `state = c_docu_state_active` — by design, because state is not portable.

The MSAG handler works around this by clearing `dokstate` in its own serialization path before calling into the longtexts helper. No other handler does, so the field leaks for FUGR, PROG, and any other consumer.

### Proposed fix
Clear `dokil-dokstate` inside the existing `iv_clear_fields` block in `zcl_abapgit_longtexts->read`, so all consumers behave consistently with the established pattern from #2605:
```abap
IF iv_clear_fields = abap_true.
  CLEAR: ls_longtext-dokil-dokstate,
         ls_longtext-head-tdfuser,
         ls_longtext-head-tdfreles,
         ls_longtext-head-tdfdate,
         ls_longtext-head-tdftime,
         ls_longtext-head-tdluser,
         ls_longtext-head-tdlreles,
         ls_longtext-head-tdldate,
         ls_longtext-head-tdltime.
ENDIF.
```
Tested locally — after re-pushing from the source system, the XML no longer contains `<DOKSTATE>`, and Diff is clean on the target.

### PR
I have a fix and will submit a PR shortly.

IGNVDRON:
Closes #7729.

### Change
`src/objects/texts/zcl_abapgit_longtexts.clas.abap`, method `read`. Add `dokil-dokstate` to the existing `iv_clear_fields` block:
```abap
IF iv_clear_fields = abap_true.
  CLEAR: ls_longtext-dokil-dokstate,    " added
         ls_longtext-head-tdfuser,
         ...
ENDIF.
```

### Why
`DOKIL-DOKSTATE` is a runtime field that abapGit intentionally does not round-trip — see #2605, which already stripped it for MSAG. The same clearing was missing in the shared longtexts class, so it leaked into XML for FUGR, PROG, and other consumers. The deserializer in this class already ignores the field (hardcoded `state = c_docu_state_active`), so XML carrying a value can never be honored — producing a permanent Diff between repo and target.

See #7729 for full analysis and reproduction.

### Test
Pushed a function group whose function-module longtext had `DOKIL-DOKSTATE = 'R'`. Before: XML carried `<DOKSTATE>R</DOKSTATE>`, target system had `DOKSTATE = 'A'` after pull, Diff stayed dirty. After: XML no longer contains `<DOKSTATE>`, Diff is clean on the target.

### DIFF
--- a/src/objects/texts/zcl_abapgit_longtexts.clas.abap
+++ b/src/objects/texts/zcl_abapgit_longtexts.clas.abap
@@ -218,7 +218,7 @@ CLASS zcl_abapgit_longtexts IMPLEMENTATION.
 
     LOOP AT lt_longtexts ASSIGNING <ls_longtext>.
 
-      lv_no_main_lang = boolc( iv_main_language <> <ls_longtext>-dokil-langu ).
+      lv_no_main_lang = boolc( <ls_longtext>-dokil-masterlang IS INITIAL ).
 
       CALL FUNCTION 'DOCU_UPDATE'
         EXPORTING
--- a/src/objects/texts/zcl_abapgit_longtexts.clas.abap
+++ b/src/objects/texts/zcl_abapgit_longtexts.clas.abap
@@ -121,7 +121,8 @@ CLASS zcl_abapgit_longtexts IMPLEMENTATION.
           line    = ls_longtext-lines.
 
       IF iv_clear_fields = abap_true.
-        CLEAR: ls_longtext-head-tdfuser,
+        CLEAR: ls_longtext-dokil-dokstate,
+               ls_longtext-head-tdfuser,
                ls_longtext-head-tdfreles,
                ls_longtext-head-tdfdate,
                ls_longtext-head-tdftime,


mbtools (approved):
LGTM. Thanks 👍

PR #7732, opened 2026-06-04
INTF: Fix attribute, methods, events docu deserialization

### LINKED ISSUE #7731: Interface member documentation missing after pull
### Summary
When pulling an interface, documentation on its members (attributes, methods, events) is missing on the target system, even though it is present in the repo's XML.

### Steps to reproduce
1. In SE24, open an interface that has no description on the interface itself, and add documentation to one of its methods (select method → F9).
2. Push the interface to abapGit.
3. Pull the repo on another system. Diff will show that <LONGTEXTS_IO> is missing.
4. In SE24 on the target system, open the method's documentation. It is empty.

Subsequent pulls do not restore it either.

IGNVDRON:
Closes #7731.
Method `zcl_abapgit_object_intf->deserialize_docu( )` returned early when the interface had no own documentation, skipping `deserialize_longtexts` for attributes, methods and events. 

Replaced early RETURN with IF/ELSE so that IA/IO/IE longtexts are still deserialized when the interface itself has no description, matching the existing CLAS pattern.

### DIFF
--- a/src/objects/zcl_abapgit_object_intf.clas.abap
+++ b/src/objects/zcl_abapgit_object_intf.clas.abap
@@ -184,23 +184,22 @@ CLASS zcl_abapgit_object_intf IMPLEMENTATION.
         iv_id          = c_longtext_id-interface
         iv_object_name = lv_object
         iv_language    = mv_language ).
-      RETURN.
-    ENDIF.
-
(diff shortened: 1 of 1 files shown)


mbtools (approved):
Good find. Thanks for the fix! 👍

Issue #4140, opened 2020-11-05
Error pulling SOAP Service-Provider (SPRX/INTF)


snapIBR:
When pulling a package containing a service-provider the deserialization of the service interface fails.
In zcl_abapgit_object_intf->deserialize_proxy an exception with the following message is raised:
`'' is not valid for 'operation pattern'. Expected values: NORMAL`
The problem may also be with the creation of the service provider itself, the pattern for the methods isn't set:
![grafik](https://user-images.githubusercontent.com/73654335/98248196-e4c95800-1f74-11eb-9261-3e16437a1a74.png)
Whereas it's set to `NORMAL` in the original system:
![grafik](https://user-images.githubusercontent.com/73654335/98248401-2a862080-1f75-11eb-8ef3-b2be8f78c453.png)
I've implemented a simple calculator SOAP API as a testcase in my repo: https://github.com/snapIBR/abapgit_serv_provider.git

christianguenter2:
I had similiar erros when dealing with other proxy objects. Couldn't detect the root cause though.
You can try patching it manually in the debugger. 

a) Operation pattern: Set a breakpoint in CL_PXNE_OPERATION IF_PXN_OPERATION_CONFIG~GET_OPERATION_PATTERN then set `pattern` to `normal`
![image](https://user-images.githubusercontent.com/17437789/98294860-eb75c080-1fb0-11eb-947c-d5ee2f1d26fb.png)


b) Interface pattern: Set a breakpoint in CL_PXNE_INTERFACE IF_PXN_CONFIGURATION~GET_INTERFACE_PATTERN then set `pattern` to `1`
![image](https://user-images.githubusercontent.com/17437789/98295603-16144900-1fb2-11eb-8ee0-1d779f4df363.png)

Maybe you can help finding the root cause?

snapIBR:
It seems the webservice profiles, which are used to derive the patterns, would be created via WEBI, but since the webservice was generated from a WSDL abapGit relies solely on SPRX-objects.
Unfortunately it seems like `cl_proxy_data=>db_save` doesn't offer any possibility to set the parameters.

mbtools:
The issue is related to the WEBI object which is not deserialized properly. I fixed a few things for WEBI in #5106. However, due to the following, the WEBI object is never created:

```abap
  METHOD zif_abapgit_object~deserialize.
...
    READ TABLE ls_webi-pvependpoint INDEX 1 INTO ls_endpoint.
[...]

PR #7737, opened 2026-06-08
Serialize and Deserialize WEBI Object Alongside Proxy Object XML when its automatically generated

igormorse:


mbtools:
Hi Igor

This looks like a solution for https://github.com/abapGit/abapGit/issues/4140 which is an old known bug. I tried to solve it many times but had no luck. 

Can you create a new PR just with the ABAP code, please?

PR #7740, opened 2026-06-11
Stage page: morph Patch button into "Patch Selected/Filtered (n)" like the commit button

christianguenter2:
The commit action on the stage page already adapts to the current context — "Add All and Commit", "Commit Selected (n)", or "Add Filtered and Commit (n)". The Patch button in the toolbar stayed a static "Patch" regardless. This change makes Patch behave the same way. Note that the sematic of the change buttons changed if something is filtered. Previously it was ignored, now it's considered.

| Context | Button label | Action on click / hotkey `p` |
|---|---|---|
| Rows manually selected | Patch **Selected** (n) | Patch the selected rows (unchanged) |
| Filter active, nothing selected | Patch **Filtered** (n) | Mark visible rows as added/removed, then patch them — same semantics as "Add Filtered and Commit" |
| Default | Patch | Unchanged |

image
image


mbtools (approved):
Nice 👍

PR #7742, opened 2026-06-11
Fix staging error when link hints are combined with the stage filter.

christianguenter2:
Staging or patching after typing a filter on the Stage page fails with:
`Unable to stage ZCL_ABAPGIT_GUI_PAGE_STAGE.CLAS.ABAP30. If the filename contains spaces, this is a known issue. Consider ignoring or staging the file at a later time.`

How to reproduce (before this fix)
1. Enable link hints (set a link hint key in the settings)
2. Open the Stage page of a repo with changes
3. Activate Link Hints
4. Type something into the filter box and press Enter
5. Commit or Patch (toolbar or command palette) => error 

image

Note the trailing 30 — it is the link hint code of the file's anchor leaking into the filename that gets posted to the backend.

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -607,7 +607,7 @@ StageHelper.prototype.onPageLoad = function() {
   var data = window.sessionStorage && JSON.parse(window.sessionStorage.getItem(this.pageSeed));
 
   this.iterateStageTab(true, function(row) {
-    var status = data && data[row.cells[this.colIndex["name"]].innerText];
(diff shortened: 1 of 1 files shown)


mbtools:
This makes me wonder if the commit link should be in the top menu as well. Might need a bit of styling to separate the multi-word menu items though.

I would name the menu "Commit: ..." and "Patch: ..."

christianguenter2:
> This makes me wonder if the commit link should be in the top menu as well. Might need a bit of styling to separate the multi-word menu items though.
> 
> I would name the menu "Commit: ..." and "Patch: ..."

Yeah, there's for sure more room for improvement. I try to come up with further fixes.

PR #7743, opened 2026-06-12
Stage page: add hotkey and command palette entry for "Changed by (me)" filter

christianguenter2:
The staging page already offers a "Changed by (me)" link in the table header to toggle filtering by the current user, but it was only reachable by mouse. This PR makes it keyboard-accessible:

New hotkey m ("Filter changed by me") registered in ZCL_ABAPGIT_GUI_PAGE_STAGE->ZIF_ABAPGIT_GUI_HOTKEYS~GET_HOTKEY_ACTIONS. It dispatches to the existing onFilterMe function in StageHelper (same mechanism as the p/submitPatch hotkey) and automatically appears in the hotkey help (?). The injected "(me)" link now gets a title and the command CSS class in StageHelper.injectFilterMe, so enumerateUiActions() picks it up and it appears in the command palette (F1) as "Filter changed by me". The class is only a marker for the palette enumerator; the link's appearance is unchanged.

No behavior change for the filter itself — pressing m toggles it on/off exactly like clicking the  link.


mbtools (approved):
👍

PR #7744, opened 2026-06-12
Fix "Activate Objects" for classes and interfaces

mbtools:
Fixes `Advanced > Activate Objects` for classes and interfaces which were previously ignored.


christianguenter2 (commented):
👍

PR #7745, opened 2026-06-12
Link hints: redeploy on activation instead of once

christianguenter2:
Before
image

After
image

<details>
<summary>Summary</summary>

Link hints were generated a single time, when the page scripts initialized. Any link created or rewritten after that point never got a hint, so on dynamic pages — most visibly the stage page — large parts of the UI were unreachable via link hints.

Two concrete failure modes on the stage page:

Late-created links never get hints. The per-row add / ignore / remove / reset commands are rendered by StageHelper at window load (updateRowCommand, initial run) — after LinkHints had already scanned the DOM. The whole command column therefore never had hints.
Existing hints get destroyed. Hint badges are injected as <span>s inside the target anchors, and the stage page rewrites those cells via innerHTML on every status click and every filter highlight — silently wiping the hints. The result was gaps in the hint numbering and links that visibly lost their hints during normal use.

</details>

<details>
<summary>Change</summary>

LinkHints no longer deploys hint containers in the constructor. Instead, each time the hint hotkey is pressed to show hints, it:

removes any hint containers still attached from the previous activation (removeHintContainers), and
runs deployHintContainers() against the current DOM.
Hiding hints, prefix filtering, yank mode, and dropdown handling are unchanged — they operate on the freshly built map exactly as they did on the static one.

</details>

<details>
<summary>Behavior notes</summary>

Hints now appear on all links present at the moment of activation, on every page (e.g. stage table commands, links revealed after filtering or toggling sections on the diff page).
Hint codes are renumbered on each activation — consistent with Vimium-style hinting, where codes are ephemeral by design.
Cost is one DOM scan per activation (on demand only); no work is done on pages where hints are never invoked, which previously paid the deployment cost unconditionally at load.

</details>

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -1409,7 +1409,7 @@ function LinkHints(linkHintHotKey) {
   this.linkHintHotKey    = linkHintHotKey;
   this.areHintsDisplayed = false;
   this.pendingPath       = ""; // already typed code prefix
-  this.hintsMap          = this.deployHintContainers();
+  this.hintsMap          = null; // deployed on each activation, the DOM changes after load
   this.activatedDropdown = null;
   this.yankModeActive    = false;
 }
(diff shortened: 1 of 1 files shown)


mbtools (approved):
👍

PR #7746, opened 2026-06-12
Stage page: unify Commit and Patch in the toolbar

christianguenter2:
<details>
<summary>Summary</summary>

The stage page previously offered Commit via three separate buttons below the
file list (`Commit All`, `Commit Selected`, `Commit Filtered`) while Patch
lived in the top toolbar — even though both act on the same staging selection.
This PR moves Commit up next to Patch and makes the two behave symmetrically.

**Toolbar before:** `Refresh | Diff | Patch | Back` + separate commit button row
**Toolbar after:** `Add All and Commit (n) | Patch All (n) ‖ Diff | Refresh | Back`

</details>

<details>
<summary>Changes</summary>

**Single dynamic Commit button.** The three commit anchors are replaced by one
toolbar button (`commitBtn`) whose label and action follow the current mode,
same as Patch already did:

| Mode | Commit | Patch |
|---|---|---|
| nothing selected | Add All and Commit (n) | Patch All (n) |
| files selected | Commit **Selected** (n) | Patch **Selected** (n) |
| filter active | Add **Filtered** and Commit (n) | Patch **Filtered** (n) |

**Patch default now matches Commit default.** Previously, Patch with no
selection opened the patch page for the *entire* repo diff — including
remote-only files that "Add All and Commit" would never touch — so the two
buttons silently operated on different file sets (e.g. 6 vs. 14). With no
selection, `get_page_patch` now falls back to the `stage_all` file set (all
local changes plus purely local deletions), and the button label shows that
count. When there is nothing to add or remove (n = 0), the label stays "Patch"
and the previous full-diff behavior is preserved.

**Visual grouping.** Toolbar separator items now also render at the top level
(previously only styled inside dropdowns) as a thin vertical divider, setting
the Commit/Patch pair apart from navigation actions.

**Command palette stays in sync.** Palette entries built from toolbar anchors
snapshotted their labels at page load and went stale once the buttons were
rewritten. Entries now re-read the anchor text each time the palette opens.

**Hotkeys:** new `c` → Commit, mirroring the existing `p` → Patch. Ctrl+Enter
continues to trigger Commit.

</details>

<details>
<summary>Implementation notes</summary>

- `ZCL_ABAPGIT_GUI_PAGE_STAGE`: `render_actions` reduced to the filter row;
  toolbar built in `get_menu` (separator items, server-rendered default labels
  with count from `count_default_files_to_commit`); `get_page_patch` falls
  back to `stage_all` when the selection yields no add/rm files;
  `render_scripts` passes `commitBtn` id and `stageAllAction`.
- `common.js` (`StageHelper`): new `submitCommit()` dispatches per mode
  (selected → `submit()`, filtered → `submitVisible()`, default → `stage_all`
  sapevent); `updateMenu` rewrites both button labels; default labels are
  captured from the server-rendered buttons at init. `CommandPalette` toolbar
  entries carry a `getTitle` closure refreshed in `toggleDisplay`.
- `common.css`: new top-level `.nav-container > ul > li.separator` rule
  (1px vertical divider, inherits text color, theme-safe).

</details>

<details>
<summary>Screenshots</summary>

All
image

Filtered
image

Selected
image

</details>

### DIFF
--- a/src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap
+++ b/src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap
@@ -788,11 +788,11 @@ CLASS zcl_abapgit_gui_page_stage IMPLEMENTATION.
         iv_txt = ``
         iv_typ = zif_abapgit_html=>c_action_type-separator
       )->add(
-        iv_txt = 'Refresh'
-        iv_act = |{ c_action-stage_refresh }|
-      )->add(
         iv_txt = |Diff|
         iv_act = |{ zif_abapgit_definitions=>c_action-go_repo_diff }?key={ mi_repo->get_key( ) }|
+      )->add(
+        iv_txt = 'Refresh'
+        iv_act = |{ c_action-stage_refresh }|
       )->add(
         iv_txt = |Back|
         iv_act = zif_abapgit_definitions=>c_action-go_back ).
--- a/src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap
+++ b/src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap
@@ -322,6 +322,13 @@ CLASS zcl_abapgit_gui_page_stage IMPLEMENTATION.
     DELETE lt_files WHERE method <> zif_abapgit_definitions=>c_method-add
                       AND method <> zif_abapgit_definitions=>c_method-rm.
 
+    IF lt_files IS INITIAL.
+      " No selection: patch the same file set as the default commit (stage_all)
+      lt_files = stage_all( )->get_all( ).
+      DELETE lt_files WHERE method <> zif_abapgit_definitions=>c_method-add
+                        AND method <> zif_abapgit_definitions=>c_method-rm.
+    ENDIF.
+
     ri_page  = zcl_abapgit_gui_page_patch=>create(
       iv_key        = lv_key
       it_files      = lt_files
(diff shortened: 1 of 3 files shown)


src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap
```diff
@@ -770,20 +756,43 @@ CLASS zcl_abapgit_gui_page_stage IMPLEMENTATION.
...
       )->add(
-        iv_txt = |Patch|
+        iv_txt = lv_patch_all_txt
         iv_typ = zif_abapgit_html=>c_action_type-onclick
         iv_id  = |patchBtn|
+        iv_opt = zif_abapgit_html=>c_html_opt-strong
+      )->add(
+        iv_txt = ``
+        iv_typ = zif_abapgit_html=>c_action_type-separator
+      )->add(
```
mbtools: Diff first, then refresh?

mbtools (commented):
Nice. I will test tomorrow

src/ui/pages/zcl_abapgit_gui_page_stage.clas.abap
christianguenter2: 👍

mbtools (approved):
It takes a little while to get used to but it's much more consistent with the rest.

My only follow-up would be to move the filter box to the left with a label and X button matching the repo list filter.

christianguenter2:
@mbtools what do you thinkg about this layout (not yet committed)?
image

mbtools:
The problem of two toolbars... there should be only one.... everywhere. Your last suggestion is to have only "Back" (or page navigation) at the top and the toolbar on the page (more like Fiori). I think that would be a separate change that should be applied consistently across *all* pages. For now my preference is to consolidation into one toolbar at the top (with "Back" as the last item).

christianguenter2:
Ok, merge as is then?

PR #7747, opened 2026-06-12
Fix binary file detection

mbtools:
If a file is longer than 1000 bytes, it can happen that the tail of the bytes has an incomplete utf8 character. The result was a failing utf8 conversion and detection as binary. The fix strips any imcomplete tail before the conversion.


larshp (approved):
thanks

PR #7757, opened 2026-07-04
user exit: adjusting/generating commit description + body

larshp:
so its possible to hook in robots, beep beep


larshp:
tested working

Issue #7760, opened 2026-07-06
Serialized programs do not contain system variants


HiWiSciFi:
In SAP you can create system variants which will be transported automatically along with the program they are for, thus they behave as part of the "metadata" of a program. (Information on System Variants: [here under "Notes"](https://help.sap.com/docs/SAP_NETWEAVER_701/6f43fe6c6c4b1014a88dfde574176a62/c0980389e58611d194cc00a0c94260a5.html))

How I would implement it:
Either
- Add a section to the `x.prog.xml` file (I am not very familiar with this yet, but this might be problematic if the file format is not allowed to change; if it is, I would just add a section after the Dynpro serialization)

or
- Add a seperate xml/json file for system variants (like `x.prog.vari.xml`)

I would be happy to implement this functionality myself if others deem it a useful feature.

mbtools:
You can extend the xml. Feel free to create a PR 👍

PR #7762, opened 2026-07-07
PROG/FUGR: Add support for system variants

HiWiSciFi:
Include serialization of system variants in the serialization of objects which can have selection screens attached.
Issue: #7760 

- [X] (De)Serialization of variant header
- [X] (De)Serialization of Parameter- & Select-Options-Values
- [x] Handle protected variants (protected variants can only be edited by the creator) (causes authorization error)
- [x] Abstract FM calls to methods to clean up processing logic
- [X] i18n for short texts
- [X] Fix i18n deserialization (non-LXE) => need to test and open seperate issue/pr
- [X] Test

> [!NOTE]
> Free selections did not interact with variants the way i thought they would. SAP does not natively support saving of free selections in variants and in standard transactions where it is possible, they misuse the variant fields themselves. The free selections that are returned by the variant-related FMs are not connected to 'FREE_SELECTIONS_INIT' in any way.


mbtools:
You're getting there 👍 

Please add a test case to https://github.com/abapGit-tests/PROG_sys_variant 🙏

larshp:
took a brief look, looks fine for me

HiWiSciFi:
Alright, thanks!
I will add a test report to the abapGit-tests/PROG_sys_variant repo once I have the serialization of the variant description texts implemented 🫡

src/objects/zcl_abapgit_object_fugr.clas.abap
```diff
@@ -1507,6 +1515,12 @@ CLASS zcl_abapgit_object_fugr IMPLEMENTATION.
       ls_cua = serialize_cua( lv_program_name ).
       io_xml->add( iv_name = 'CUA'
                    ig_data = ls_cua ).
+
+      lt_varis = serialize_varis( lv_program_name ).
+      IF lt_varis IS NOT INITIAL.
```
mbtools: no need for the if because xml->add checks for initial

src/objects/zcl_abapgit_objects_program.clas.abap
```diff
@@ -942,13 +1138,133 @@ CLASS zcl_abapgit_objects_program IMPLEMENTATION.
...
+
+      CALL FUNCTION 'RS_VARIANT_CONTENTS_255'
+        EXPORTING
+          report         = iv_program_name
+          variant        = <ls_catalog>-variant
+          execute_direct = abap_true
+        TABLES
+          valutab        = ls_vari-values
+          objects        = ls_vari-objects
+*          free_selections_desc =
```
mbtools: remove comments

mbtools (changes requested):
just two small changes 🙏

src/objects/zcl_abapgit_object_fugr.clas.abap
HiWiSciFi: Thanks, will do 👍🏽 
Edit: resolved in 233b1a9

mbtools:
Thx. It's still draft. Ready to merge?

HiWiSciFi:
Not yet, I would implement support for free selections, but as it turns out, the option to have those be a part of a variant is not available on older systems (going by paramters of the used FMs) (might be a change that came with S/4, not sure though).
What is the intended "abapGit-Way" of dealing with this?
If newer versions are able to store free selections in variants and older versions are not, should I implement this dynamically or not at all?
(this could be moved to a seperate issue/PR i guess)

I also have the serialization of the description text only implemented in the main language of the abapGit repo (`mv_language` in `zcl_abapgit_objects_program`). The way I understand it I should at least implement serialization of the additional languages (while respecting "only serialize main language" setting)

mbtools:
support in new versions, don't break old versions 😉 

here's a typical pattern for function calls that have new parameters in higher releases:
https://github.com/abapGit/abapGit/blob/58cde24a7eaf1c4e84893655648ddcbcec5aa886/src/objects/zcl_abapgit_objects_program.clas.abap#L637

Just add a todo for the additional languages

HiWiSciFi:
I am currently a bit stuck on a SAP feature that causes issues with the deserialization when chaning an existing variant:
When creating a system variant (or any variant for that matter) you have the option to protect it against changes. That way, only the user who last changed it (aka who set the flag to true) can make changes / delete the variant.
For abapGit this of course poses the issue that if the variant was created by another user than the one currently running the deserialization, it will run into errors because the variant cannot be edited or deleted.

SAP does provide a Report (`RSVARENT`) to remove protections from a variant as any user (provided they have sufficient authorization). This could in theory be used in abapGit to remove protection, update, and add the protection back in. SAP does not provide a FM for this (Understandable since it is kind of hacky).

This approach comes with a few caveats however.

- Let's say i'd use `SUBMIT` to call the report. That makes the deserialization functionality dependent on SAP not renaming the Parameters on the selection screen, since they do not use memory ids (has to be called with `SUBMIT rsvarent WITH ...`)
- If I were to reimplement what `rsvarent` does, i would have to do a hard write/update on the database, which is usually discouraged (depending on an official report definitely seems safer)
- Even if I implement it, doing the "remove protection, update, add protection" thing will change the last user who has edited the variant and therefore a user who previously had editing rights for that variant will lose them (this is somewhat expected behaviour but should probably at least be documented somewhere)

Any opinions/help would be appreciated ❤️

mbtools:
The serialized data must not contain any user-, system-, ot time-specific data. For example, all user names and timestamps must be cleared. Deserialize should set "last changed" to the current user (but it won't show in the XML). 

We use "remove protection, update, add protection" in some other place. It should be fine here, too. Just remove the protection with direct DB update.

HiWiSciFi:
Alright, thanks!
I wouldn't have included the User in the XML anyway, if anything I would have restored the previous owner, but since the "last changed" should be set to the current user I don't have to implement anything like that 👍

I will implement the protection updating using the direct update then, I'll implement it today in the afternoon 🫡

HiWiSciFi:
> [!NOTE]
> PO (LXE) translation entries are already automatically being serialized along with the report translations. They are also only serialized for system variants and not for "normal" variants, which means that no further adjustment is required in that department (with LXE enabled).

~Still have to check deserialization, but I will update this comment once I have.~ Deserialization works as well

Tested on A4H 2023

mbtools:
Let's get this merged. Any leftovers can go into a new issue/pr. Change from draft to ready?

HiWiSciFi:
Yes, sorry - i've had very little time the last few weeks.
I do know of a few minor things that do not really work correctly, but nothing that really breaks anything so i'll mark this as ready and move the remaining todos to a seperate issue/pr.

mbtools:
During deserialize, `get_vari_screens` was throwing if a variant didn't exist, for example, one first import of the test repo. This lead to an import error. The screen comparison wasn't necessary. I changed it to simply drop the variant and (re-)create it. 

There was a small diff on `mlangu` fields which I [removed from the test repo](https://github.com/abapGit-tests/PROG_sys_variant/commit/d79ebea895e3b58aa2a3b4db911f1484a4deb6dc). 

The rest was great!

PS: Sorry the reordering of methods which happens when you touch the code with `SE80`.

mbtools (approved):

mbtools:
Existing tests pass 👍 

image

mbtools:

PR #7763, opened 2026-07-08
refactor password popup

larshp:
delegates showing the login popup to the UI layer, so we can get rid of the SAP gui popup and do HTML instead


src/http/zcl_abapgit_http.clas.abap
```diff
@@ -84,39 +84,63 @@ CLASS zcl_abapgit_http IMPLEMENTATION.
...
+    " sending, so for Basic auth we only reach this point when there are no credentials
+    " yet or the cached ones were rejected with a 401.
+    lv_user = zcl_abapgit_login_manager=>get_username( iv_url ).
+    lv_pass = zcl_abapgit_login_manager=>get_password( iv_url ).
+
+    " Digest (e.g. https://www.gerritcodereview.com/) needs the cleartext password on
+    " every challenge and cannot be pre-applied as a static header, so reconstruct it
+    " from the cached credentials.
+    " https://en.wikipedia.org/wiki/Digest_access_authentication
+    IF rv_scheme = c_scheme-digest AND lv_user IS NOT INITIAL.
```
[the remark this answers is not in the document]
larshp: think this is okay, I dont think digest is used very much

src/http/zcl_abapgit_http.clas.abap
```diff
@@ -84,39 +84,63 @@ CLASS zcl_abapgit_http IMPLEMENTATION.
...
-    IF lv_user <> lv_default_user.
-      zcl_abapgit_persist_factory=>get_user( )->set_repo_login(
-        iv_url   = iv_url
-        iv_login = lv_user ).
-    ENDIF.
-
-    rv_scheme = ii_client->response->get_header_field( 'www-authenticate' ).
-    FIND REGEX '^(\w+)' IN rv_scheme SUBMATCHES rv_scheme ##REGEX_POSIX.
-
     CASE rv_scheme.
```
mbtools, line 142: Digest is already handled above so can be removed here

src/ui/core/zcl_abapgit_gui.clas.abap
```diff
@@ -333,6 +341,28 @@ CLASS zcl_abapgit_gui IMPLEMENTATION.
...
+        " and cache them in the login manager.
+        TRY.
+            IF request_credentials( lx_auth->mv_url ) = abap_true.
+              IF ls_handled-state IS INITIAL.
+                " Authentication interrupted the event handler, retry the action
+                handle_action(
+                  iv_action   = iv_action
+                  iv_getdata  = iv_getdata
+                  it_postdata = it_postdata ).
+              ELSE.
```
mbtools, line 355: Can this else case really happen? If handler is set, it was successful meaning not interrupted by this exception. Maybe just assert that state is initial?

src/http/zcl_abapgit_http.clas.abap
larshp: the first digest handling is for the cached scenario, the CASE is for after user entered details

src/ui/core/zcl_abapgit_gui.clas.abap
larshp: it shows an error on the next page if its not there

its probably not pretty, but overall, I'd like to move to html based popups and stuff, that will make it much easier to get agents to click around and test and fix stuff

mbtools:
I tested with a private repo i.e. login required. If you cancel the popup, the popup goes away but any subsequent try to open the repo results in `An exception was raised (sy530)`.

saplogon_SsYKTCQ3np

Here's the location of the raise:

image

image

It seems like this is the problem that copilot mentioned.

Otherwise works well.

larshp:
thanks, fixed

mbtools:
If the user or password is empty and you hit enter, it just cancels the popup. It should try to authenticate (and error) or give a message that user/password is obligatory.

larshp:
> If the user or password is empty and you hit enter, it just cancels the popup. It should try to authenticate (and error) or give a message that user/password is obligatory.

fixed,
image

mbtools (approved):
Ship 🛳️ 👍

Issue #7765, opened 2026-07-08
Trying to commit in an abapgit Project with LXE enabled fails with CX_SY_DYN_CALL_ILLEGAL_TYPE


DerGuteWolf:
Trying to commit in an abapgit Project with LXE enabled fails with CX_SY_DYN_CALL_ILLEGAL_TYPE, see 

<img width="457" height="192" alt="Image" src="https://github.com/user-attachments/assets/4a563ee0-995f-448e-bc42-f91a6f2fda4c" /> and

<img width="433" height="226" alt="Image" src="https://github.com/user-attachments/assets/fc915ef3-0f55-4101-8bc1-c08e926768c0" />

Reason is in https://github.com/abapGit/abapGit/blob/b20d89386b9585c19927d6dfb81d7004bc241986/src/objects/texts/zif_abapgit_lxe_texts.intf.abap#L10 and https://github.com/abapGit/abapGit/blob/b20d89386b9585c19927d6dfb81d7004bc241986/src/objects/texts/zif_abapgit_lxe_texts.intf.abap#L11

lxe_pcx_s1 has uppcase as lxeuppcase which is char(1) and has texttype as lxetexttype which is char(4)

PR #7767, opened 2026-07-08
Bugfix type `LXE_PCX_S1`

DerGuteWolf:
lxe_pcx_s1 has uppcase as lxeuppcase which is char(1) and has texttype as lxetexttype which is char(4)


larshp (approved):
thanks, looks good to me

mbtools:
Thanks 👍 

Closes #7765

PR #7770, opened 2026-07-09
Select source branch for new branch

### LINKED ISSUE #7769: Allow selecting the source branch when creating a new branch
## Description
When creating a new branch, abapGit always bases it on the currently checked-out branch, and the source branch isn't shown. There's no way to branch from a different source. Since abapGit switches to the new branch after each create, the checked-out branch changes over time, making it easy to branch off the wrong source by mistake.

## Request
- Let the user pick the source branch when creating a new branch.
- Add an optional per-repo default source branch in Local Settings (used as the default when set). If not set, the currently checked-out branch is used as the default, as today.

## Benefit
It is always clear what the source of a new branch will be, and you can pick the right one. This avoids branching from the wrong source by mistake. Optional: no change when left empty.

IGNVDRON:
Closes #7769

Adds the ability to choose the source branch when creating a new branch, so it's always clear what a new branch will be based on.

**Changes**
1. Commit screen: new read-only "Create New Branch From" field next to "New Branch Name", with a branch picker.
2. Create-branch popup: now asks for the source branch first, with a clear title.
3. Local Settings: optional per-repo "Default Source Branch for New Branches". Used as the default value in 1 and 2 when set; otherwise the checked-out branch is used, as today.

All optional — no change in behavior when the setting is left empty.

**Screenshots**

Local Settings – new default source branch field:
image

Commit screen – new "Create New Branch From" field:
image

Create-branch popup – source branch selection:
image


mbtools:
- Do we really need the default branch setting? How many times are you creating branches so that the setting will save you a 2 seconds to pick the right source branch? 
- Instead of changing "Branch > Create" to a two step process, I would prefer a new menu item "Branch > Create From"
- Changing the source branch for the commit does not look like a wise thing to do. You are pushing a diff that is based on the current branch. So if you use a different branch as a source, what exactly do you end up with?

IGNVDRON:
Thanks for your comments!

* The main idea is to avoid a mistake, not to save 2 seconds. abapGit switches to the new branch right after creating it, so the checked-out branch keeps drifting and you have to remember to switch back to the base branch first. 
My real pain is on the commit screen: I stage the necessary objects, fill all the commit info, and only then notice the checked-out branch is wrong - at which point I have to go back to the first abapGit screen, choose correct branch and redo everything. That's why the idea to change the new branch source on the commit screen came up to me. But good catch - I hadn't thought it through: changing the source after the differences are already calculated would produce incorrect results.
* On the separate menu item - agree. That would leave current branch creation process unchanged.

So, should we keep the separate menu for new branch source and setting for default source of new branches? Or maybe I just have to learn my lesson, remember the steps, and stop trying to fix it with a feature 😃

mbtools:
imho, a setting should be useful for a large part of the users. I don't see that being the case here. 

"Branch > Create From" certainly makes sense and I would use it too 😀

IGNVDRON:
Reworked based on the review.
### Changes
- Option `Branch > Create From` dropdown item added in repository view.
- Option `+ create new from ...` added to the switch-branch popup on main screen.

### Screenshots
Branch dropdown – new "Create From" entry:
image

Switch-branch popup – new "+ create new from ..." entry:
image

Source branch selection pop-up:
image

mbtools:
It works well! Thanks you 🙏

PR #7773, opened 2026-07-10
refactor branch picker to html popup

larshp:
phase 2 from https://github.com/abapGit/abapGit/pull/7764


larshp:
@copilot resolve the merge conflicts in this pull request

larshp:
lets move forward 🚢 it will probably need a few more fixes tho

mbtools (approved):
Tested. A popup like in "Repo Settings > Remote" would be nicer but this works. 👍

larshp:
> Tested. A popup like in "Repo Settings > Remote" would be nicer but this works. 👍

we can probably have the LLM fix that after it lands

PR #7774, opened 2026-07-10
Refactor `xml->add` initial check

mbtools:
`xml->add` already contains a check for initial values. This removes the unnecessary ifs in the serializer classes.


larshp (approved):
thanks

PR #7781, opened 2026-07-22
repo overview: mass export of selected offline repos

### LINKED ISSUE #7780: Repository list: mass export of selected offline repositories
## Problem

On the repository list (Repositories overview) only a single repository can be selected at a time, and the offline **Export** action exports one repository per invocation. Exporting several offline repositories one by one is tedious.

## Proposal

Add a checkbox to offline repositories in the repository list (plus a "select all" checkbox in the header). Selecting one or more and pressing **Export** writes one re-importable ZIP per repository into a folder chosen once by the user — a timestamped sub-folder that is opened afterwards — following the existing **Transport Mass Downloader** pattern.

To keep the selection model consistent, the single-repository, selection-based actions (Pull, Stage, Diff, Import, …) show an error when more than one repository is selected.

## Notes

- Scope is limited to **offline** repositories (only they get a checkbox).
- Implementation reuses `zcl_abapgit_zip=>encode_files` / `zcl_abapgit_zip=>save_binstring_to_localfile` and the frontend service `directory_browse`, and adds a new action `zip_export_selected`. Each generated ZIP is byte-identical to the current single Export, so it remains individually re-importable.

I have a working implementation ready to submit as a PR. Opening this issue first to discuss the approach/UX, as suggested in `CONTRIBUTING.md`.

lrcherubini:
Closes #7780

Offline repositories now have a checkbox in the repository list (plus a "select all" checkbox in the header). Selecting one or more and pressing Export writes one re-importable ZIP per repository into a chosen folder (timestamped sub-folder, which is opened afterwards), mirroring the Transport Mass Downloader pattern.

Single-repo, selection-based actions (Pull, Stage, Diff, Import, …) now show an error when more than one repository is selected.

Reuses zcl_abapgit_zip=>encode_files / save_binstring_to_localfile and the frontend directory_browse. Adds action zip_export_selected.


mbtools:
I'm not opposed to this feature but I'm wondering how useful it really is. How many users would use it and how many times? Also why limit to offline repos?

It sounds more like a standalone tool using the classes provided in the abapGit docs (API).

larshp:
> I'm not opposed to this feature but I'm wondering how useful it really is. How many users would use it and how many times? Also why limit to offline repos?
> 
> It sounds more like a standalone tool using the classes provided in the abapGit docs (API).

hmm, perhaps try asking an LLM to write the standalone tool, to see how that would look

lrcherubini:
> > I'm not opposed to this feature but I'm wondering how useful it really is. How many users would use it and how many times? Also why limit to offline repos?
> > It sounds more like a standalone tool using the classes provided in the abapGit docs (API).
> 
> hmm, perhaps try asking an LLM to write the standalone tool, to see how that would look

To make sure I build the right thing — by "standalone tool" do you mean (a) a separate report/page shipped inside the abapGit repo itself, similar to the existing Transport Mass Downloader (not integrated into the repo overview screen, but still part of this codebase), or (b) a fully separate report living outside the abapGit repo, only consuming the public API classes?
Because an external tool has no compile-time link to abapGit's internals, so an internal refactor (e.g. to zcl_abapgit_zip or zcl_abapgit_repo_srv) could silently break it (offline import/export process).

mbtools:
Option b. We mitigate side-effects of abapGit changes by running CI that checks several repos that depend on abapGit. We can add a new tool.

lrcherubini:
hum. Could be something like this? With all logic wrapped in a class "zcl_abapgit_repo_mass_export" for example?
image
image

mbtools:
Thank you for your contribution, Leandro. I just don't see this being used much. 

Based on your idea, I added [`ZABAPGIT_REPO_EXPORT`](https://github.com/Marc-Bernard-Tools/ABAP-Tools-for-abapGit/pull/24) to the collection of abapGit-related community tools at https://github.com/Marc-Bernard-Tools/ABAP-Tools-for-abapGit. Feel free to submit enhancements there.

image

larshp:
hi, did you figure this out?

lrcherubini:
Hi Marc and Lars. Sorry for late I was too busy. Thanks for the `ZABAPGIT_REPO_EXPORT`.

PR #7782, opened 2026-07-23
Enable browser back navigation in SAPGUI + WebGui

### LINKED ISSUE #7779: GUI Back-Navigation breaks event handlers
UPDATE:
Issue has been reproduced in a slightly different way. See [here](https://github.com/abapGit/abapGit/issues/7779#issuecomment-5059752941).

Orignal Issue:
When using the SAP-GUI back-button keybind (F3) to navigate back from a repo view to the overview, the event handlers in zcl_abapgit_gui are wrong.
The internal table still contains the event handlers from the repo view (ZCL_ABAPGIT_GUI_PAGE_REPO_VIEW) instead of the overview (ZCL_ABAPGIT_GUI_PAGE_REPO_OVER).

This causes incorrect event handling

> [!NOTE]
> This only happens when using the F3-Key to navigate back! Pressing the button on the screen works

Found on on-prem with SAP_BASIS 750 SP 0007
Reproduced on A4H 2023 (SAP_BASIS 758 SP 0002)

Tested with multiple Themes and occurs on every one, so the issue is most likely in the abapGit handling.

Results from debugging:
- `zcl_abapgit_gui->back` is called when clicking on the back-button but not when using F3
- `AT SELECTION-SCREEN ON EXIT-COMMAND` is not called on F3
- Guess: The issue might be with the GUI status

christianguenter2:
Browser Back did not trigger abapGit's back. Intercept it via a History  API sentinel + popstate and route it to the go_back sapevent.

  - Click the real rendered Back element so the control's own handler runs. Match via el.hrefsav on WebGUI (ITS strips the action from href), href/formaction on the desktop control. Clicking preserves ITS session context that a rebuilt form submit would lose.
  - Fall back to submitting go_back on pages with no Back element (works in the desktop control; does not route on WebGUI).
  - Ignore self-induced popstate via gSapeventNavPending so only genuine Back presses trigger go_back.
  - Wire up with redirectBrowserBackToSapEvent() on page load.
  -  Add back button to repo view

fixes #7779

### DIFF
--- a/src/ui/lib/zcl_abapgit_gui_page.clas.abap
+++ b/src/ui/lib/zcl_abapgit_gui_page.clas.abap
@@ -397,6 +397,7 @@ CLASS zcl_abapgit_gui_page IMPLEMENTATION.
     render_command_palettes( ri_html ).
     ri_html->add( |toggleBrowserControlWarning();| ).
     ri_html->add( |displayBrowserControlFooter();| ).
+    ri_html->add( |redirectBrowserBackToSapEvent();| ).
 
   ENDMETHOD.
 
--- a/src/ui/pages/zcl_abapgit_gui_page_repo_view.clas.abap
+++ b/src/ui/pages/zcl_abapgit_gui_page_repo_view.clas.abap
@@ -1247,7 +1247,11 @@ CLASS zcl_abapgit_gui_page_repo_view IMPLEMENTATION.
       iv_act = zif_abapgit_definitions=>c_action-abapgit_home
     )->add(
       iv_txt = zcl_abapgit_gui_buttons=>help( )
-      io_sub = zcl_abapgit_gui_menus=>help( ) ).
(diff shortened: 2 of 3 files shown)


HiWiSciFi:
Altight, i've tried every way I could think of to break it and it seems to work.
- Rightclick -> Back
- Binding `back` to a key and pressing it
- Alt+Left
each of those does navigate back and it's being handled correctly by SAP now.
I've also tried forward navigation (Alt+Right) but it doesn't do anything (which is alright imo) which also means that it doesn't break this.

Just one quick question about the code (more of a general abapGit question, not directly related to this PR): i've the rest of the js file and have realized that your change is consistent with it, but I do wonder why `var` is used everywhere instead of `let`? Is it due to browser version compatibility?

mbtools:
`let` (and `const`) didn't exist when abapGit was started and IE10/11 never fully supported it 😉

Issue #7636, opened 2026-03-19
SAP GUI for HTML: Patching does not work


mbtools:
Patch page does not work with SAP GUI for HTML.

Ref #7625, #7628

mbtools:
Closed by #7782, #7783, #7814, #7816

Issue #6829, opened 2024-02-27
SAP GUI for Java - Minor issues: Hotkeys, vimium features and command palettes not working


christianguenter2:
follow up on #5082

PR #7783, opened 2026-07-24
Make JS-triggered sapevents work on WebGUI via a global form and data-sapevent marker

christianguenter2:
**TL;DR fixes broken hotkeys, command palette, stage/patch buttons, browser back, link hints and more on WEBGUI + simplifications and cleanups**

Need thorough testing on:
- [x] SAPGUI for Windows Edge
- [x] SAPGUI for Windows IE
- [x] WebGUI
- [x] SAPGUI for Java

On SAP GUI for HTML (WebGUI), ITS only routes sapevents through forms and
anchors it wired up while server-rendering the page. Forms built by
JavaScript at submit time are not wired, so the raw "sapevent:" scheme is
rejected — breaking every JS-initiated action on WebGUI: hotkeys, the
command palette, and the browser-back fallback. ITS also rewrites sapevent
hrefs (e.g. "sapevent:go_back" -> "#sapevent25"), so JS cannot reliably
find an element by its action or rebuild the URL from a rewritten href.

Two server-rendered hooks fix both problems:

* zcl_abapgit_gui_page: render a single hidden
  <form id="global_sapevent_form"> in the page scaffold, so every page has
  one ITS-wired form. submitSapeventForm reuses it when the caller passes
  no form, instead of creating an unwired one (createElement remains as
  fallback for pages outside the standard scaffold). Injected hidden
  fields are tagged with data-sapevent-field and cleared on the next
  submit, since the form is shared and non-navigating actions (filter,
  clipboard yank) would otherwise accumulate stale values. On the desktop
  browser control the form's action is simply overwritten, so it is
  harmless there.

* zcl_abapgit_html->a(): emit a data-sapevent="<action>" attribute
  (HTML-escaped) on sapevent anchors. Unlike the href, this marker
  survives ITS rewriting, giving JS a stable way to locate the real wired
  element for a given action.

On top of these, the JS callers now trigger the server-rendered element
instead of reconstructing the navigation:

* findSapEventElements/findSapEventElement (new, shared): match elements
  by data-sapevent first, falling back to hrefsav/href/formaction on the
  desktop controls. Whole-word regex with metacharacters escaped, so an
  action like "jump?key=1" cannot throw or mismatch. Anchors are preferred
  over submit inputs.

* clickSapEvent (new): click a sapevent element after setting
  gSapeventNavPending, so the browser-back trap ignores the popstate the
  control emits during the navigation. Non-sapevent elements (onclick,
  plain links) do not arm the flag — it would never be consumed and the
  next genuine Back press would be swallowed.

* Hotkeys: replace the href-scraping resolution chain (getSapEventHref /
  getSapEventInputAction / getSapEventForm /
  getAllSapEventsForSapEventName / eliminateSapEventFalsePositives) with a
  single findSapEventElement + clickSapEvent. Clicking the element also
  preserves any getdata it carries (e.g. "...?key=<repo>"). Hotkey
  tooltips use the same finder.

* Command palette (enumerateUiActions): always click the wired anchor via
  clickSapEvent instead of extracting the action from the href, which is
  rewritten on WebGUI anyway.

* Browser back (triggerSapEventBack): the no-Back-element fallback now
  routes through the global form, so browser Back also works on WebGUI on
  pages that render no Back element (e.g. repo list).

The per-page hidden-form workaround is retired: the stage page's
render_deferred_hidden_events is removed along with chunk_lib's
render_event_as_form and its ty_event_signature type, and
submitSapeventForm no longer looks for form_<action> stubs.

### DIFF
--- a/src/ui/core/zcl_abapgit_html.clas.abap
+++ b/src/ui/core/zcl_abapgit_html.clas.abap
@@ -347,13 +347,14 @@ CLASS zcl_abapgit_html IMPLEMENTATION.
 
   METHOD zif_abapgit_html~a.
 
-    DATA: lv_class TYPE string,
-          lv_href  TYPE string,
-          lv_click TYPE string,
-          lv_id    TYPE string,
-          lv_act   TYPE string,
-          lv_style TYPE string,
-          lv_title TYPE string.
+    DATA: lv_class         TYPE string,
+          lv_href          TYPE string,
+          lv_click         TYPE string,
+          lv_id            TYPE string,
+          lv_act           TYPE string,
+          lv_style         TYPE string,
+          lv_title         TYPE string,
+          lv_data_sapevent TYPE string.
 
     lv_class = iv_class.
 
@@ -385,7 +386,12 @@ CLASS zcl_abapgit_html IMPLEMENTATION.
           IF iv_query IS NOT INITIAL.
             lv_act = lv_act && `?` && iv_query.
           ENDIF.
-          lv_href  = | href="sapevent:{ lv_act }"|.
+          lv_href          = | href="sapevent:{ lv_act }"|.
+          " Stable action marker that survives ITS href rewriting on WebGUI,
+          " so JS (e.g. hotkeys) can find and click the element by its sapevent.
+          lv_data_sapevent = | data-sapevent="{ escape(
+            val    = lv_act
+            format = cl_abap_format=>e_html_attr ) }"|.
         WHEN zif_abapgit_html=>c_action_type-onclick.
           lv_href  = ' href="#"'.
           lv_click = | onclick="{ iv_act }"|.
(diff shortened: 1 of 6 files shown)


christianguenter2:
What I testet successfully
- [x] SAPGUI for Windows 8.10 - Edge
- [x] SAPGUI for Windows 8.10 - IE
- [x] WEBGUI on SAP ABAP 2023
- [x] JAVAGUI for Windows 8.10 rev 8

mbtools:
I tried win and web gui. AFAICT, it works great!  Thanks 🙏 

Looks like we can close #7636

Does https://github.com/abapGit/abapGit/issues/6829 work now, too? If yes, we might call it officially supported

christianguenter2:
Yes, this definitively fixes #7636 and #6829. Thx for the heads up.
My main workflow (keyboard centric and patching) works great on javagui with this fix. So imho we can call it supported. Last pieces are removing this warning (not sure whether it made ever any sense in javagui context)
image
and then we can remove this 
image

Btw. `get_gui_version` looks a bit strage for javagui:
image
image

also interesting, browser dev tools can be opened in javagui directly
image

mbtools:
Nice. Let's do that in subsequent pr. I will update the docs after. 

The Java gui version looks shifted from the version-id (typical sap bug because they don't have a test for it 🤷 ).

mbtools (approved):
Lgtm 👍

PR #7786, opened 2026-07-27
additional hotkey ctrl+p for command palette action selection

christianguenter2:
In javagui, after toggling command selector with F1, it loses the focus to the OK-Code field. Command palette has to be refocussed to select a command. Therefore Id like to change the hotkey from F1 to ctrl+p, similar to other tools.


mbtools:
It would also open up F1 for triggering help (as it typically does).  👍

christianguenter2:
adjust hotkey overview 

Before
image
After
image

sbcgua:
F1 actually works in vs code as well (by default). And actually it is quite convenient (better than ctrl+p, if one doesn't need help ... which is rather rare). I'd suggest keeping both (by the way, like vs code also does).

christianguenter2:
> F1 actually works in vs code as well (by default). 

ah didn't  know that

> And actually it is quite convenient 

the jury is still out on that one ;-)

> I'd suggest keeping both (by the way, like vs code also does).

sure, ok for me. I'll adjust

christianguenter2:
Now we have both
image

mbtools:
in my win gui (8.00.15), ctrl+p opens the command palette and the print dialog. i think it needs to prevent the keyboard event from going further (`e.preventDefault();`). F1 works fine, though.

image

christianguenter2:
Interesting, with edge control? Works for me with 8.10 and edge control. AFAIS `event.preventDefault()` is called.
Turns out I have the same issue when using the IE control.
image

Hm...

mbtools:
IE control only. Edge works. IE is dead. Just leave as is

mbtools (approved):

sbcgua:
works for me as well. Thanks!

PR #7789, opened 2026-07-27
SHMA: bugfix GUI popup

larshp:
if deserializing only a SHMA, it will show a popup when trying to generate the classes, test with https://github.com/abapGit-tests/SHMA/tree/hvam/onlytheshma

it will now show an error that the root class doesn't exist. For on-stack development it will generate it along with the area, for off-stack the class must be created at the same time


mbtools:
The generated class might be different in different releases. (It should probably not be serialized). I'm not sure if there's a better solution to avoid the popup 🤷

PR #7793, opened 2026-07-30
Repo View: Back leaves abapGit if end of nav stack is reached

### LINKED ISSUE #7792: Repo View: "Back" not working correctly
The repo view now has "Back" in toolbar (since https://github.com/abapGit/abapGit/pull/7782). If you launch abapGit directly to the repo view ("Show Last Opened Repository" = on), then clicking "Back" does nothing. I would expect to exit abapGit (since it was the first nav step).

christianguenter2:
fixes #7792


mbtools (approved):
thanks 👍

PR #7796, opened 2026-07-30
update AFF files + add SMTG AFF

larshp:
files manually copied from https://github.com/abapGit/aff_mirror/

`DOMA` logic adjusted


larshp:
opened https://github.com/SAP/abap-file-formats/pull/820 for fixing the SMTG typos

mbtools:
In aff_mirror, the interfaces have no comments (abapdoc). Where did they come from? Do we need them in AG?

larshp:
> In aff_mirror, the interfaces have no comments (abapdoc). Where did they come from? Do we need them in AG?

will update, sec, did wonder why the comments got in

mbtools (approved):
👍

PR #7797, opened 2026-07-30
bugfix diff3

larshp:
beep beep 🤖


larshp:
oh, wrong way around, closing

larshp:
nah, perhaps not

perhaps I'm tired 😴

larshp:
now works on and off stack,

image

mbtools (approved):
Nice! 

No bugs in my diff3 class?

larshp:
> Nice!
> 
> No bugs in my diff3 class?

I think the diff3 was okay, just the SAP flavor worked a bit different, so this adjusts it to how the SAP part works, so we hopefully get similar/identical functionality

PR #7798, opened 2026-07-30
Add "F3" to hotkey panel

mbtools:
- Add "F3 - Go back" to hotkey infopanel
- Make the keys wider to avoid wrapping 
- Put the F-keys in order

PS: contains updated where-used of JS functions in ABAP

After:

image


mbtools:
@copilot add the eslint hints back but keep the where used list

christianguenter2:
@copilot add the eslint hints back but keep the where used list

christianguenter2 (commented):
LGTM 👌

mbtools:
It removed errorstub and another function. I can't check from remote but don't we still need them?

christianguenter2:
They are not used anymore, so from my understanding we can delete them.
Last `errorStub`-call was removed with #7746
`getIndocStyleSheet` never had a caller. Even when it was introduced with #2825.

Btw. AI is great for such research
image

mbtools:
Thank you!

PR #7800, opened 2026-08-03
feat(objects): add RAP Behavior Definition (BDEF) object serializer

ashvinupadhyay1132:
Adds object serializer for RAP Behavior Definitions (BDEF).

- Implements ZCL_ABAPGIT_OBJECT_BDEF class inheriting from ZCL_ABAPGIT_OBJECTS_SUPER.
- Includes unit tests.


mbtools:
Please add a description so we understand why this serializer class should be changed

mbtools:
Hmm. BDEF is already supported. Here's an example https://github.com/abapGit-tests/BDEF. Why do we need a change?

PR #7801, opened 2026-08-04
feat(objects): add OData V4 Service Model (IWOM) object serializer

ashvinupadhyay1132:
.


larshp:
@ashvinupadhyay1132 can you help explaining a bit? at first sight this looks like spam

mbtools:
IWOM is already supported

PR #7805, opened 2026-08-06
SMTG: wip

larshp:
ref https://github.com/abapGit/abapGit/pull/7682


larshp:
open issues/questions:
https://github.com/abapGit/abapGit/pull/7682#issuecomment-5202518147
https://github.com/SAP/abap-file-formats/issues/797

PR #7808, opened 2026-08-06
DTEL: AFF handling

larshp:
https://github.com/abapGit-tests/DTEL-AFF-examples created

image


larshp:
doing https://github.com/abapGit/abapGit/pull/7809 first

mbtools (approved):
tested and works well. 👍

PR #7809, opened 2026-08-06
DOMA: add unit tests + bugfixes

larshp:


mbtools:
I fixed a couple things in https://github.com/abapGit-tests/DOMA-AFF-examples and used it for a test. 

There are some warnings after pull and one diff:

image

image

PS: When experimental AFF is OFF, pulling the test repo dumps (expected an error):

image

image

larshp:
image

larshp:
image

image

there is a diff for output length, which I think is okay, the SE11 editor adds it automatically,

image

src/objects/zcl_abapgit_objects.clas.abap
```diff
@@ -1236,6 +1245,26 @@ CLASS zcl_abapgit_objects IMPLEMENTATION.
...
 
 
+  METHOD read_metadata.
+
+    CLEAR: eo_xml, es_metadata.
+
+    IF io_files->is_json_metadata( ) = abap_true.
+      " There's no XML and metadata for JSON (AFF) format
+      IF zcl_abapgit_aff_factory=>get_registry( )->is_supported_object_type( is_item-obj_type ) = abap_false.
+        zcx_abapgit_exception=>raise( |Object is serialized in AFF format. Enable experimental | &&
```
mbtools: add another case to check if object type is actually supported with experimental flag. so we have two message: "not supported at all" and "supported with experimental flag"

larshp: fixed in https://github.com/abapGit/abapGit/pull/7809/commits/43b5af563083e1ff5e45ba09e74932f018b63d8d

mbtools (approved):
Tested. Alright, this is good to go now! 👍

Issue #7715, opened 2026-05-22
XML serializer: CX_SY_CONVERSION_NO_NUMBER when NUMC/P fields contain leading blanks


skosysparency:
### Environment
- abapGit version: latest `main` (verified against build as of 2026-05-22)
- SAP NetWeaver / S/4HANA: n/a
- Affected class: `ZCL_ABAPGIT_XML_OUTPUT`, method `ZIF_ABAPGIT_XML_OUTPUT~ADD`

### Steps to Reproduce
1. Have any object whose serializer passes an elementary data object of
   ABAP type `P` (packed) or `N` (NUMC) to `io_xml->add( )` — directly
   or indirectly via a structure component — where the value contains
   a leading blank (e.g. `' 00000'` instead of `'000000'`).
2. Such invalid content can arise from:
   - Customer Z-tables with a domain change in the past where existing
     rows were never migrated.
   - Legacy data imported via SQL/LSMW that did not honour the NUMC
     conversion exit.
   - Buffered system tables where a `SPACE` slipped into a numeric
     column at insert time.
3. Run serialization (stage / push / pull) on that object.

### Expected behavior
The serializer either tolerates the invalid content (e.g. by treating
the value as a string and writing it as-is into the XML element) or
raises a `zcx_abapgit_exception` with a clear object reference, so the
calling framework can skip the object and log the issue instead of
terminating the whole run.

### Actual behavior
Runtime error `CX_SY_CONVERSION_NO_NUMBER` from inside
`CALL TRANSFORMATION id ... SOURCE (lt_stab) RESULT XML li_doc.` in
`zif_abapgit_xml_output~add`. The exception is not caught by abapGit —
the program terminates with an ST22 dump and the whole serialization
run is aborted.

### Root cause
Current implementation (file `zcl_abapgit_xml_output.clas.abap`,
method `zif_abapgit_xml_output~add`, lines verified on current `main`):

```abap
APPEND INITIAL LINE TO lt_stab ASSIGNING <ls_stab>.
<ls_stab>-name = iv_name.
GET REFERENCE OF ig_data INTO <ls_stab>-value.

CALL TRANSFORMATION id
  OPTIONS initial_components = 'suppress'
  value_handling = 'move'
  SOURCE (lt_stab)
  RESULT XML li_doc.
```

`ig_data` is passed by reference straight into `CALL TRANSFORMATION id`.
For elementary types `P` and `N`, the identity transformation performs
strict numeric validation of the source value during serialization to
[...]

PR #7812, opened 2026-08-06
Catch transformation errors

mbtools:
Avoids dumps when serializing inconsistent data like fields for type n with leading spaces (instead of leading zeros).

Mirrors the behavior of deserializing (in xml_input). 

Ref #7715

https://help.sap.com/doc/abapdocu_latest_index_htm/latest/en-US/ABAPCALL_TRANSFORMATION_OPTIONS.html


larshp:

mbtools:
Thats what you get when you try to edit on a phone. Thank you for fixing it 🙏

larshp:
ship it 😀

PR #7814, opened 2026-08-07
WebGUI: Fix dialog buttons

### LINKED ISSUE #7811: WebGUI: Commit Button doesn't work
Pressing the commit button in WebGUI doesn't transfer fields and result in errors. Pressing hotkey 'enter' works though.
Image

Image

christianguenter2:
Clicking a button of a dialog on the HTML GUI did not transfer the fields of the form and ran into validation errors, while pressing enter worked. Several problems, all around the way ITS wires sapevent routing into a page:

1. ITS wires up the action of a form, but not the formaction attribute of an input, so commands are rendered as links. A link navigates without the payload of the form, hence the empty fields. Render them as links that submit the form instead.

2. ITS uses two carriers for that routing: an empty form like global_sapevent_form gets it encoded into the action, a form holding submit elements gets a dummy action plus hidden ~control / ~event / PARAMS fields. submitSapeventForm only knew the first variant, so on a dialog it left the baked in PARAMS untouched and every command raised the event of the form - "Back" on the commit page committed. Set PARAMS instead when ITS carries the routing in fields, for all of them, since a form can hold more than one pair.

3. ITS replaces submit() and collects the fields of the form itself, reading the value of every entry of form.elements. A fieldset is part of that collection but has no value, which made the collection fail on every dialog using field groups (e.g. local and remote settings). Hand out an empty value for those.

4. A generated form id holds a time stamp, whose decimal separator is not valid in a selector. Look the form up by id instead, and keep the generated ids free of the separator.

5. The side actions of a field and the radio buttons raised their event by hand, either through formaction or by writing a raw sapevent url into the action of the form. Both bypass the routing of ITS, and a raw sapevent url inside a handler even gets rewritten by ITS, which breaks the JS around it. Route them through submitSapeventForm too.

Fixes #7811


mbtools:
I installed a UR patch which broke my Webgui 😠. Sorry can't test this but looks good 👍

mbtools (approved):

Issue #3860, opened 2020-09-09
ACGR: Support for PFCG roles (authorizations) 


gardian12:
Hello together, I need to resume the initial feature request #827 

**How to handle roles?**

Here are the facts about roles:

- Roles can't be handled with BC Sets
- Roles are customizing and not linked to a development package
- Roles are essential part of an application and should be delivered as well
- Roles can be easily handled as files, see transaction PFCG (upload / download function)
- Therefore it would be highly recommended also to cover PFCG roles by abapGit
- Without disturbing the core of abapGit it would be possible.

As an example: Customizing files, like roles, could be stored with a separated folder /**_cust_**/, parallel to /**_src_**/.
Manual selection of role to be added to the repo would be needed any way.

Hope you can reopen, the feature request with this insights.

jrodriguez-rc:
We can build initial object support based on Download/Upload roles functionality
![image](https://user-images.githubusercontent.com/56556686/92698006-3629e200-f34c-11ea-98ef-4475da10c361.png)

As a workaround, while is thinking how to assign customizing to the repository, roles can be added via abapGit exit `zif_abapgit_exit->change_tadir`

gardian12:
Well at first we can keep it simple, like a simple upload/download option at abapGit UI to keep or retrieve the files.
In background there should be a separated folder created and handled by abapGit to store the files.
This folder should be different to the _src_ folder not to harm the development objects underneath.

jrodriguez-rc:
Some additional changes need to be done, everything outside of `/` and `/src/` folders is ignored
![image](https://user-images.githubusercontent.com/56556686/92709372-24017100-f357-11ea-935a-9e0cda778c04.png)

jrodriguez-rc:
Role support is being discussed in this issue #3474

gardian12:
When everything outside of `/` and `/src/` folders is ignored, we do not have a problem o create a folder beside `/src/` (e.g. `/cust/`) to store additional files. The function module `PRGN_DOWNLOAD_AGRS` and `PRGN_EXECUTE_UPLOAD` can do the job of writing and receiving a role as file. Details as discussed in #3474 are covered by these.

mbtools:
Since there are no `TADIR` entries for roles i.e. no association with packages, I would treat them the same way as `TABU` data. An extension of the current data serializer seems feasible resulting in role data being persisted in `/data/`.

mbtools:
[...]

PR #7815, opened 2026-08-07
ACGR: Add object handler for PFCG roles (#3860)

IGNVDRON:
## Add ACGR object handler for PFCG roles (#3860)
Adds `zcl_abapgit_object_acgr`, an object handler that lets abapGit serialize PFCG security roles (object type **ACGR**).&#x20;

Relates to #3860 and #3474.

### What it does
Implements the standard `zif_abapgit_object` methods:

- **serialize / deserialize** - full round trip of the role's R3TR ACGR transport tables, keyed by `AGR_NAME`. `AGR_USERS` / `AGR_USERT` are excluded.
- **delete** - via the standard `PRGN_ACTIVITY_GROUP_DELETE`.
- **is\_locked**, **jump**, **changed\_by** and the remaining minor methods.

### How to use it for now
Role-to-package assignment can be stored in table `AGR_ATTS` with `ATT_TYPE = 'DEVCLASS'`.

1. Append the domain value `DEVCLASS` to the domain `MENU_ATT3`.
2. Assign each role to its package via `SM30`->` VAGRATTS`: 
   `agr_name` = role name
   `att_type` = `DEVCLASS`,
   `att_value` = package name.
3. Inject the roles into staging by implementing `zif_abapgit_exit~change_tadir`: select the roles whose `AGR_ATTS`-`DEVCLASS` lies within the repository's package tree and append them as `R3TR ACGR` rows, each carrying its folder path.

### Open question
Roles are not in TADIR, so abapGit cannot discover them via the standard object list (which is built from `zcl_abapgit_tadir->read`). A handler like `zcl_abapgit_object_acgr` can serialize a role, but there is no discovery step that feeds it into the repo. This can be solved via `change_tadir` exit, but every abapGit user should implement it manually.

A cleaner option would be an optional interface `zif_abapgit_object_lister` with `list_objects_in_package( iv_package )` returning TADIR-shaped rows. After reading TADIR, `zcl_abapgit_tadir` could call all implementations of 'list_objects_in_package( )' and merge results. `zcl_abapgit_object_acgr` would then provide object list based on package assignment (using AGR_ATTS or other), removing the need of exit.

What is your opinion on adding this hook to abapGit core?

### DIFF
--- a/src/objects/zcl_abapgit_object_acgr.clas.abap
+++ b/src/objects/zcl_abapgit_object_acgr.clas.abap
@@ -1,0 +1,244 @@
+CLASS zcl_abapgit_object_acgr DEFINITION
+  PUBLIC
+  INHERITING FROM zcl_abapgit_objects_super
+  CREATE PUBLIC.
+
+  PUBLIC SECTION.
+    INTERFACES zif_abapgit_object.
+
+  PROTECTED SECTION.
+    METHODS get_generic
+      RETURNING
+        VALUE(ro_generic) TYPE REF TO zcl_abapgit_objects_generic
+      RAISING
+        zcx_abapgit_exception.
+
+  PRIVATE SECTION.
+ENDCLASS.
+
+
+
+CLASS zcl_abapgit_object_acgr IMPLEMENTATION.
+  METHOD get_generic.
+    CREATE OBJECT ro_generic
+      EXPORTING
+        is_item     = ms_item
+        iv_language = mv_language.
+  ENDMETHOD.
+
+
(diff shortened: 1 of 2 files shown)


mbtools:
It's a creative solution but requires a change to a standard domain, additional maintenance of roles, and an exit. That is not suitable for standard abapGit. 

Please discuss changes like this in an issue before coding and submitting a PR.

mbtools:
The "before export" function does a lot of things for roles. We can't skip this. Sorry, this is not going to work.

PR #7816, opened 2026-08-07
Fix links to namespaced objects on the HTML GUI

christianguenter2:
A namespace ends up as # in the file name of an object (/emi/tax -> #emi#tax) and # gets escaped when the name is put into the url of a sapevent. The desktop browser control unescapes the url before handing it over, the HTML GUI does not, which is why unescape has to do it - and it covered every character escape_url produces except %23.

Clicking a diff link or an object link on a namespaced object therefore ended in "File /src/%23emi%23... not found".
image


mbtools (approved):
👍

PR #7813, opened 2026-08-07
Fix SAPGUI for Windows determination

christianguenter2:
fixes #7810 
| Check | SAP GUI for Windows | SAP GUI for Java | WebGUI |
|-------|:-------------------:|:----------------:|:-------:|
| **zcl_abapgit_frontend_services** ||||
| `is_sapgui_for_windows` | ✅ | ❌ | ❌ |
| `is_webgui` | ❌ | ❌ | ✅ |
| `is_sapgui_for_java` | ❌ | ✅ | ❌ |
| **Function Modules** ||||
| `GUI_IS_AVAILABLE` | ✅ | ✅ | ✅ |
| `GUI_HAS_ACTIVEX` | ✅ | ❌ | ✅ |
| `GUI_HAS_JAVABEANS` | ❌ | ✅ | ❌ |
| `GUI_IS_ITS` | ❌ | ❌ | ✅ |
| `GUI_IS_WEBGUI` | ❌ | ❌ | ✅ |
| **cl_gui_object** ||||
| `gui_is_running` | ✅ | ✅ | ✅ |
| `activex` | ✅ | ❌ | ✅ |
| `javabean` | ❌ | ✅ | ❌ |
| `www_active` | ❌ | ❌ | ✅ |

<details>
<summary>For the records check report</summary>

```
REPORT ztest_gui.

CLASS report DEFINITION.

  PUBLIC SECTION.
    CLASS-METHODS run.

  PRIVATE SECTION.
    CLASS-METHODS gui_is_available
      RETURNING VALUE(result) TYPE abap_bool.

    CLASS-METHODS gui_has_activex
      RETURNING VALUE(result) TYPE abap_bool.

    CLASS-METHODS gui_has_javabeans
      RETURNING VALUE(result) TYPE abap_bool.

    CLASS-METHODS gui_is_its
      RETURNING VALUE(result) TYPE abap_bool.

    CLASS-METHODS gui_is_webgui
      RETURNING VALUE(result) TYPE abap_bool.

ENDCLASS.


CLASS report IMPLEMENTATION.
  METHOD run.
    WRITE / |===========-zcl_abapgit_frontend_services-===========|.
    WRITE / |is_sapgui_for_windows: { zcl_abapgit_ui_factory=>get_frontend_services( )->is_sapgui_for_windows( ) }|.
    WRITE / |is_webgui: { zcl_abapgit_ui_factory=>get_frontend_services( )->is_webgui( ) }|.
[...]

PR #7817, opened 2026-08-07
Do not report a browser control on the HTML GUI

christianguenter2:
Needs #7813 as prerequisite.

The footer guessed the type of the embedded browser control from the user agent, so on the HTML GUI, which runs in the browser of the user and embeds nothing, it claimed "IE" for every browser except Edge.

Render the placeholder only for SAP GUI for Windows, the one GUI that does embed a control, mirroring is_edge_control_warning_needed. The JS side now checks whether the placeholder exists at all.

Before:
image

After:
image


mbtools (approved):
👍

PR #7830, opened 2026-08-19
fix(ui): arm back-trap flag in submitFormById

### LINKED ISSUE #7826: Database Utility: Save does not work with Edge Control
Clicking "Save" when editing a config entry in Data Utilities returns to the previous page without saving when using SAP GUI for Windows with Edge Control.  It works fine with IE.

Is this just my machine?

Image

christianguenter2:
The browser-back trap ignores popstate events that our own sapevent navigations cause, by way of the gSapeventNavPending flag. Every sapevent entry point set it except submitFormById, which submits a server-rendered sapevent form directly.

On the Edge control the popstate emitted while handling such a submit was therefore read as a genuine Back press and fired go_back, which superseded the pending request. Saving a config entry in the Database Utility returned to the previous page without persisting; the three Apply buttons on the merge resolution page were broken the same way. The IE control is unaffected because the trap never arms there (no pushState support).

Closes #7826

### DIFF
--- a/src/ui/zabapgit_js_common.w3mi.data.js
+++ b/src/ui/zabapgit_js_common.w3mi.data.js
@@ -319,8 +319,17 @@ function setInitialFocusWithQuerySelector(sSelector, bFocusParent) {
   }
 }
 
-// Submit an existing form
+// Submit an existing, server-rendered sapevent form (its action carries the
+// event, so nothing has to be rewritten here).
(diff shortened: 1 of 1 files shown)


mbtools:
Test issue fixed in #7829

mbtools (approved):
Thank you 😃

Issue #6215, opened 2023-04-14
SLDD: Generic Application Access Rules (Definition)


ed-holland:
This object is used for Access Rule Definitions defined in transaction SLDW.

These rules are used to circumvent SQL Injections. Abap coding needs to be adjusted to use these rules.
The Rule definitions needs to be transported via the AddOn and GIT support will help with change management.
More information can be found in the SAP Guide "Securing_Remote_Function_Calls_RFC".

The details of this object seem to be located in the following tables:

SLDW_HEADER
SLDW_HEADERT (by Language)
SLDW_ELEMENTS

mbtools:
To implement this object type, create a PR or reopen this issue for further discussion (#5912).

PR #7832, opened 2026-08-21
SLDD: Add support for Generic Application Access Rules

IGNVDRON:
**SLDD** (Generic Whitelist for RFC-enabled function modules) serializer (#6215).

### DIFF
--- a/src/objects/zcl_abapgit_object_sldd.clas.abap
+++ b/src/objects/zcl_abapgit_object_sldd.clas.abap
@@ -71,7 +71,7 @@ CLASS zcl_abapgit_object_sldd IMPLEMENTATION.
       WHERE name = ms_item-obj_name
       ORDER BY moddate DESCENDING modtime DESCENDING.
     ENDSELECT.
-    IF  ls_element_change-changed_by IS NOT INITIAL
+    IF ls_element_change-changed_by IS NOT INITIAL
     AND ls_element_change            >= ls_header_change.
       rv_user = ls_element_change-changed_by.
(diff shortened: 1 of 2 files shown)


mbtools:
Please add a test case to https://github.com/abapGit-tests/SLDD

IGNVDRON:
Created a testcase.
https://github.com/abapGit-tests/SLDD/pull/1

src/objects/zcl_abapgit_object_sldd.clas.abap
```diff
@@ -0,0 +1,124 @@
...
+    CREATE OBJECT ro_generic
+      EXPORTING
+        is_item        = ms_item
+        iv_language    = mv_language
+        io_field_rules = lo_field_rules.
+  ENDMETHOD.
+
+
+  METHOD zif_abapgit_object~changed_by.
+    rv_user = c_user_unknown. " not stored by SAP
```
mbtools: Isn't this stored in `SLDW_HEADER-MODIFIER`? Should be a dynamic select so it doesn't cause syntax error in lower releases

mbtools (changes requested):

src/objects/zcl_abapgit_object_sldd.clas.abap
IGNVDRON: Yeah, it is stored in both - SLDW_HEADER and SLDW_ELEMENTS. I have added logic to find last modifier from both tables.

mbtools (approved):
Works. Thanks for the addition! 👍

PR #7835, opened 2026-08-23
update npm dependencies + enable rule xml_bom

larshp:
fyi @mbtools


mbtools (approved):
Thanks! This rule will help off stack to get it right more often 😁

PR #7837, opened 2026-08-24
feat(git): allow an explicit commit timestamp

larshp:
Add ZCL_ABAPGIT_GIT_TIME=>GET_UNIX_FROM_LOCAL, the inverse of GET_UTC, which converts a date and time in the time zone of the application server, eg E070-AS4DATE and E070-AS4TIME, into unix time. The value is converted to UTC, so the timezone indicator stays '+0000' like in GET_UNIX. SY-ZONLO is not used, it is the personal time zone of the user and can be empty in background jobs.

Add an optional TIME field to ZIF_ABAPGIT_GIT_DEFINITIONS=>TY_COMMENT, so the timestamp flows through all existing push callers without signature changes. TY_UNIXTIME moved to the interface, which keeps it self contained, ZCL_ABAPGIT_GIT_TIME references it, so existing usages keep working.

RECEIVE_PACK_PUSH uses the supplied timestamp for both the committer and the author, ie history replay semantics instead of cherry pick semantics, so a tool that pushes released transports can date each commit when the change actually happened. If TIME is initial the current time is used, exactly as before, no existing call site needs changing.

### DIFF
--- a/src/git/zcl_abapgit_git_time.clas.abap
+++ b/src/git/zcl_abapgit_git_time.clas.abap
@@ -6,7 +6,7 @@ CLASS zcl_abapgit_git_time DEFINITION
   PUBLIC SECTION.
 
     TYPES:
-      ty_unixtime TYPE c LENGTH 16 .
+      ty_unixtime TYPE zif_abapgit_git_definitions=>ty_unixtime .
 
     CLASS-METHODS get_unix
       RETURNING
@@ -22,6 +22,15 @@ CLASS zcl_abapgit_git_time DEFINITION
       RAISING
         zcx_abapgit_exception .
 
+    CLASS-METHODS get_unix_from_local
+      IMPORTING
+        !iv_date       TYPE d
+        !iv_time       TYPE t
+      RETURNING
+        VALUE(rv_time) TYPE ty_unixtime
+      RAISING
+        zcx_abapgit_exception .
+
     CLASS-METHODS get_utc
       IMPORTING
         !iv_unix TYPE ty_unixtime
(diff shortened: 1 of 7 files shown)


larshp:
to be used in https://github.com/abapGit/git-historical-extract

mbtools (approved):
👍

src/git/zcl_abapgit_git_time.clas.abap
```diff
@@ -77,6 +86,46 @@ CLASS zcl_abapgit_git_time IMPLEMENTATION.
...
+
+    DATA lv_seconds   TYPE i.
+    DATA lv_timestamp TYPE timestamp.
+    DATA lv_timezone  TYPE timezone.
+
+    IF iv_date IS INITIAL.
+      zcx_abapgit_exception=>raise( 'Cannot determine unix time, date is initial' ).
+    ENDIF.
+
+    lv_timezone = cl_abap_tstmp=>get_system_timezone( ).
```
mbtools: `get_system_timezone` does not exist in 702

```abap
select single tzonesys from ttzcu into lv_timezone.
```

src/git/zcl_abapgit_git_time.clas.testclasses.abap
```diff
@@ -22,6 +24,82 @@ CLASS ltcl_time_test IMPLEMENTATION.
...
+      EXPORTING
+        iv_unix = lv_seconds
+      IMPORTING
+        ev_date = lv_date
+        ev_time = lv_time ).
+
+    CONVERT DATE lv_date TIME lv_time
+      INTO TIME STAMP lv_act TIME ZONE lc_utc.
+
+    lv_timezone = cl_abap_tstmp=>get_system_timezone( ).
```
mbtools: see above

mbtools (changes requested):

src/git/zcl_abapgit_git_time.clas.abap
larshp: thanks, fixed

src/git/zcl_abapgit_git_time.clas.testclasses.abap
larshp: thanks, fixed

mbtools (approved):
👍

PR #7839, opened 2026-08-26
WDYN: Fix CX_SY_REGEX_TOO_COMPLEX in method name regex

cilku:
The current regex captures with (.*) followed by \s*, so both can match the same trailing whitespace. This makes the submatch boundaries ambiguous and the kernel raises CX_SY_REGEX_TOO_COMPLEX when resolving leftmost-longest. Restricting the capture to [\w~/]+ makes the split unambiguous. Anchoring with ^ additionally prevents matching METHOD inside identifiers such as the following line which caused a dump on my system: 
`    PO_METHOD          = l_po_method`

image


mbtools (approved):

mbtools:

PR #7840, opened 2026-08-26
SICF: Fix dump when node name contains special characters

mbtools:
Fixed `STRING_OFFSET_LENGTH_TOO_LARGE` if an SICF service contains for example a `.` in the name like `.well-known`.

Changed `name_escape/unescape` public in `zcl_abapgit_filename_logic`.

Note: `.` is the only special character allowed in service names that will be escaped.


larshp:
is it possible to add a unit test for the scenario?

mbtools:
Test repo
https://github.com/abapGit-tests/SICF_name_with_dot

larshp (approved):
I've added unit tests

mbtools:
Thanks for the tests 😃

PR #7841, opened 2026-08-27
Seed the GUI environment from ABAP instead of guessing it in JS

christianguenter2:
common.js had no single notion of which GUI it was rendered into. Each
probe re-derived a piece of it on every call, in its own vocabulary:
getSapeventPrefix ran two document-wide querySelector calls on every
sapevent submit to rediscover something that cannot change after page
load, and two separate functions open-coded
navigator.userAgent.includes("Edg").

Reading the user agent also gets the answer wrong outside SAP GUI for
Windows, where it describes the browser of the user rather than a
browser control - the footer once reported "IE" for a user on Chrome.
That was worked around by rendering the footer element only on Windows,
which left the JS silently relying on the backend's choice of markup to
stay correct.

The backend knows the GUI from SAP's own APIs, so let it say so:
render_environment emits setEnvironment(...) as the first statement of
every page's script block. It passes booleans rather than a type code,
so no shared vocabulary has to cross the language boundary.

On the JS side gEnv holds the seeded facts, getSapeventPrefix moves to
module scope and memoizes, and a named isEdgeControl() replaces the two
user agent reads. The ITS probes in submitSapeventForm now run only on
the HTML GUI, where ITS wiring can exist at all, and the branch handling
the second ITS variant reuses isItsForm instead of re-testing the action.

No behaviour change is intended on any GUI.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ list.txt
 /playwright/.cache/
 .DS_Store
 .claude
+.frontend
--- a/src/ui/lib/zcl_abapgit_gui_page.clas.abap
+++ b/src/ui/lib/zcl_abapgit_gui_page.clas.abap
@@ -92,7 +92,7 @@ CLASS zcl_abapgit_gui_page DEFINITION PUBLIC ABSTRACT
         !ii_html TYPE REF TO zif_abapgit_html
       RAISING
         zcx_abapgit_exception .
-    METHODS render_environment
+    CLASS-METHODS render_environment
       IMPORTING
         !ii_html TYPE REF TO zif_abapgit_html
       RAISING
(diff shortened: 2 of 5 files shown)


mbtools:
Good improvement! Thx

Side note: your agents add a lot of comments to the code which is good doc but it makes the standalone report larger and larger. Maybe some of it should be a design.md instead.

christianguenter2:
> Side note: your agents add a lot of comments to the code which is good doc but it makes the standalone report larger and larger. 

Valid point, or we strip the comments in the build process?

mbtools:
It's already 5x past the recommended load size. Maybe it's more fun to see how far we can push it  😀 

As you wrote, we can always strip the comments in https://github.com/larshp/abapmerge (src/pragma.ts).

image

mbtools (approved):
LGTM for Win IE + Edge

PR #7842, opened 2026-08-28
Refactor: Remove dependencies on char and numc data elements

mbtools:
Disallow generic data elements like `CHAR...`, `NUMC...`, or `SYCHAR...`


larshp (approved):
I think most of these are released, but anyhow, good

PR #7845, opened 2026-08-30
Return created package to new repository page

### LINKED ISSUE #7643: UI: Optimize "Create Package" flow
The flow to create a new package is suboptimal 

New Offline/Online, leave package empty > Create Package > Create

image

package remains empty

image

We should either return package name or ensure that package is not initial before `Create Package` and make package read-only on create package page. I think the latter is a bit more straightforward and easier to implement.

_Originally posted by @christianguenter2 in https://github.com/abapGit/abapGit/issues/7631#issuecomment-4098654470_

christianguenter2:
Closes #7643

Creating a package from the New Offline/Online Repository page left the Package field empty after returning, so the user had to type or pick the package again.

ZCL_ABAPGIT_GUI_PAGE_CPACKAGE=>CREATE now optionally accepts the calling page's form data and the field to fill. On successful creation the new package name is written back before navigating back, so the caller re-renders with the package already filled in.

Both parameters are optional, existing callers are unaffected.


mbtools (approved):
nice 👍

PR #7846, opened 2026-08-30
Return created repository url to new online repository page

christianguenter2:
Follow-up to #7643 for the "Create Repo" flow.

Creating a GitHub repository from the New Online Repository page left the Url field unchanged after returning, so the user had to type the url of the repository they had just created.

ZCL_ABAPGIT_GUI_PAGE_CR_REPO=>CREATE now optionally accepts the calling page's form data and the field to fill. On successful creation the url of the new repository is written back before navigating back, so the caller re-renders with the url already filled in.

Both parameters are optional, existing callers are unaffected.


mbtools (approved):
works 👍

PR #7852, opened 2026-09-02
TOBJ: Fix deserializing ABAP language version

### LINKED ISSUE #7851: TOBJ: ABAP language version lost on deserialize and rejected on serialize — CL_TOBJ_OBJECT_DATA carries it
Follow-up to #7833 as requested. SRVD is fixed by #7850, thanks. TOBJ fails for a different reason — and, unlike SRVD, the object is left permanently damaged.

## What happens

`ZCL_ABAPGIT_OBJECT_TOBJ` handles transport objects through `CTO_OBJECT_GET` and `OBJ_GENERATE`. Neither carries the ABAP language version:

* **deserialize** — `OBJ_GENERATE` has no language version parameter:

  ```abap
  CALL FUNCTION 'OBJ_GENERATE'
    EXPORTING
      iv_objectname = ...  iv_objecttype = ...  iv_objecttext = ...
      iv_objcateg   = ...  iv_objtransp  = ...  iv_devclass   = ...
  ```

  The created object therefore has no language version, whatever the file contains.

* **serialize** — `CTO_OBJECT_GET` returns `OBJH` with the `ABAP_LANGUAGE_VERSION` component present but empty. `clear_abap_language_version` reads that blank as *Standard ABAP* and raises:

  ```
  Object TOBJ ZEXAMPLET has ABAP language version "Standard ABAP"
  but repository is set to ABAP language version "ABAP for Cloud Development"
  ```

So a repository that pins `<ABAP_LANGUAGE_VERSION>` can never contain a TOBJ that abapGit itself created, and no `pull` repairs it.

## The damage is permanent

Once the object exists without a language version, there is **no supported way to set it**:

* `SOBJ` shows the *Language Version* field blank and read-only.
* In ADT the object opens as AFF JSON, but the `header` block — where `abapLanguageVersion` lives — is read-only.
* ADT offers no standalone creation of a Transport Object Definition, so it cannot be recreated by hand.

## ADT does set it, through an API abapGit already uses elsewhere

A TOBJ generated by the ADT *Business Configuration Maintenance Object* wizard carries the attribute. Same object type, same release, two exports:

```jsonc
// generated by ADT
"header": {
  "description": "...",
  "originalLanguage": "en",
  "abapLanguageVersion": "cloudDevelopment"
}

// created by abapGit deserialization
"header": {
  "description": "...",
  "originalLanguage": "en"
}
```

The path ADT uses is `CL_TOBJ_OBJECT_DATA` (inheriting `CL_BLUE_GENERIC_OBJECT_DATA`), whose `IF_WB_OBJECT_DATA_MODEL~SET_HEADER` does exactly this:

```abap
METHOD if_wb_object_data_model~set_header.
  ms_data-content-schema = 'https://json-schema.org/draft/2020-12/schema'.
  ms_data-content-header-abap_language_version = p_abap_langu_version.
  ms_data-content-header-description           = p_description.
  ms_data-content-header-original_language     = p_original_langu.
ENDMETHOD.
```

with the structures typed by `IF_AFF_TOBJ_V1` (*AFF Json Schema TOBJ*).

This is not a new API to adopt: **abapGit already uses this pattern for SRVD** — `get_wb_object_operator( )`, then `IF_WB_OBJECT_OPERATOR~READ` with `eo_object_data`, in `ZCL_ABAPGIT_OBJECT_SRVD`. Only the TOBJ handler still uses the pre-AFF pair. Moving it over would fix both directions and keep the language version check meaningful, and it fits the direction of #3888.

## Reproduction, independent of release

1. Deserialize a TOBJ into a package whose ABAP language version is *ABAP for Cloud Development*.
2. Set `<ABAP_LANGUAGE_VERSION>cloudDevelopment</ABAP_LANGUAGE_VERSION>` on the repository.
3. Serialize — the mismatch above is reported.

`OBJ_GENERATE` has no language version parameter on 2025 either, so step 1 produces a damaged object on any release.

## Meanwhile

We work around it by not pinning `ABAP_LANGUAGE_VERSION` on repositories that contain a TOBJ; abaplint enforces the ABAP for Cloud restrictions on our side. Reported so the case is on record, not as a blocker.

## Environment

* abapGit 1.134.0
* SAP_BASIS 758 (S/4HANA 2023), cross-checked against S/4HANA 2025
* repository language version: `cloudDevelopment`
* object: TOBJ, category `CUST`, *Individual Transaction Object*

mbtools:
Closes #7851

### DIFF
--- a/src/objects/zcl_abapgit_object_tobj.clas.abap
+++ b/src/objects/zcl_abapgit_object_tobj.clas.abap
@@ -146,11 +146,6 @@ CLASS zcl_abapgit_object_tobj IMPLEMENTATION.
     io_xml->read( EXPORTING iv_name = 'OBJM'
                   CHANGING  cg_data = lt_objm ).
 
-    ASSIGN COMPONENT 'ABAP_LANGUAGE_VERSION' OF STRUCTURE ls_objh TO <lv_abap_language_version>.
-    IF sy-subrc = 0.
-      set_abap_language_version( CHANGING cv_abap_language_version = <lv_abap_language_version> ).
-    ENDIF.
-
     CALL FUNCTION 'OBJ_GENERATE'
       EXPORTING
         iv_korrnum            = iv_transport
@@ -198,6 +193,17 @@ CLASS zcl_abapgit_object_tobj IMPLEMENTATION.
     UPDATE objh SET objtransp = ls_objh-objtransp
       WHERE objectname = ls_objh-objectname
       AND objecttype = ls_objh-objecttype.
+
+* fm OBJ_GENERATE does not respect ABAP language version so we set it here directly
+* update must be dynamic since field does not exist in lower releases
+    ASSIGN COMPONENT 'ABAP_LANGUAGE_VERSION' OF STRUCTURE ls_objh TO <lv_abap_language_version>.
+    IF sy-subrc = 0.
+      set_abap_language_version( CHANGING cv_abap_language_version = <lv_abap_language_version> ).
+
+      UPDATE ('OBJH') SET abap_language_version = <lv_abap_language_version>
+        WHERE objectname = ls_objh-objectname
+        AND objecttype = ls_objh-objecttype.
+    ENDIF.
 
 * fm OBJ_GENERATE ignores several fields like primary table flag
 * for Individual Transaction Objects


jmoutreux-twm:
Tested on SAP_BASIS 758 — it works.

Procedure, for the record:

1. Deleted the existing TOBJ. ADT refuses with *"Transport Object Definition ... is already transported"*, so I used `OBJ_GENERATE` with `iv_maint_mode = 'D'` — the same call abapGit itself makes.
2. Added `<ABAP_LANGUAGE_VERSION>5</ABAP_LANGUAGE_VERSION>` to `<OBJH>` in the serialized file. The object had been created by an earlier deserialization, so the element was missing.
3. Pulled with the patch applied.

Result: `SOBJ` now shows Language Version `5`, the AFF representation in ADT shows `"abapLanguageVersion": "cloudDevelopment"`, and the object properties show *ABAP for Cloud Development*.

Two notes that may deserve a line somewhere:

* The value written comes from the serialized file, not from the repository setting. A TOBJ already serialized without the element needs it added by hand before a pull can repair it.
* There is no other repair route. A transport object cannot be edited at all: `SOBJ` shows the field read-only, the AFF header in ADT is read-only, and `CL_WB_OBJECT_OPERATOR` refuses with *"Editing Transport Object Definitions is not supported"*. Setting the value at creation, as this PR does, is the only possibility.

---

While re-testing with `<ABAP_LANGUAGE_VERSION>` pinned, a third object type shows the same symptom for yet another reason: **SUSH**.

`ZCL_ABAPGIT_OBJECT_SUSH` does handle the language version in both directions, and its `clear_metadata` does not touch the field — it only moves `modifier`, `moddate`, `modtime`, `srcsystem`, `author` and `devclass`. So the blank value comes straight from `IF_SU22_ADT_OBJECT~SELECT`.

The picture on SAP_BASIS 758, for a package generated by the ADT *Business Configuration Maintenance Object* wizard:

* **SRVD** — field blanked by abapGit before the check (fixed by #7850)
* **TOBJ** — attribute `OBJ_GENERATE` cannot write (this PR)
* **SUSH** — value never returned by the SU22 API

Not asking for anything on the third one; reporting it so the pattern is on record. We keep `ABAP_LANGUAGE_VERSION` unset on that repository in the meantime.

Thanks for the quick turnaround.

jmoutreux-twm:
Correction to my previous comment — the SUSH part was wrong, sorry for the noise.

The blank value is not a limitation of `IF_SU22_ADT_OBJECT~SELECT`. The object simply had no language version yet: a SUSH is created together with the service binding, but the attribute is set when the object is **first maintained in SU22**. I had generated that BCO the day before and had not completed its SU22 configuration. Once I did, the object carried the language version and abapGit serialized it without complaint.

So there is no third case. On SAP_BASIS 758, with `<ABAP_LANGUAGE_VERSION>cloudDevelopment</ABAP_LANGUAGE_VERSION>` pinned, only two were real:

* **SRVD** — fixed by #7850. Still failing on my side only because my installation predates that merge.
* **TOBJ** — this PR, tested and working.

Thanks again for the quick turnaround on both.

PR #7855, opened 2026-09-05
MSAG: Add missing message translations

### LINKED ISSUE #7853: MSAG: Message class translations not serialized
When serializing an object of type MSAG (Message Class), translations for the different languages are not included in the XML.

If the translations are added to the XML using VS Code or another tool and then pulled into the SAP system using abapGit, the translations are correctly imported into the system.

However, when the object is serialized again, abapGit removes the translations from the XML. 

We have configured the translation languages in the repository settings. This configuration works correctly for other object types, but translations are not preserved for MSAG (Message Class) objects.

mbtools:
Add missing translations, if the message class does *not* have any translations but some individual messages are translated.

Closes #7853


christianguenter2:
Thx, for fixing this. Bogged me also a few times.

didierbueno:
thanks, it's working now

Issue #7768, opened 2026-07-08
Dual support for classic (XML) and AFF (JSON) during transition


mbtools:
If an object type is transitioning from XML to JSON format, we need to support both formats. For serializing, there will be a clear cutover: once AFF is shipped for an object type, abapGit will only serialize as JSON (unless marked as "experimental"). For deserialize, it depends on the format found in the repo to be backward compatible.

Having one class to support XML and JSON leads to several problems (see Ref https://github.com/abapGit/abapGit/pull/7682) that our current design doesn't support. One class for both will quickly become a maintenance nightmare (see `DOMA` and `INTF` ).

### Suggestion

If both cases need to be supported, there should be two classes: the old `ZCL_ABAPGIT_OBJECT_xxxx` (XML) and the new `ZCL_ABAPGIT_OBJECT_AFF_xxxx` (JSON). 

In `ZCL_ABAPGIT_OBJECTS=>DESERIALIZE`, we decide based on the file extension (`is_json_metadata( )`) which class to instantiate. 

In `ZCL_ABAPGIT_OBJECTS=>SERIALIZE` and `ZCL_ABAPGIT_OBJECTS=>DELETE`, the class should depend on `zcl_abapgit_aff_factory=>get_registry( )->is_supported_object_type( xxxx ).` 

### Todo

- [ ] Adjust `ZCL_ABAPGIT_OBJECTS` as described above
- [ ] Rename existing AFF classes `ZCL_ABAPGIT_OBJECT_xxxx` to `ZCL_ABAPGIT_OBJECT_AFF_xxxx`
- [ ] Split `ZCL_ABAPGIT_OBJECT_DOMA` and `ZCL_ABAPGIT_OBJECT_INTF` into two classes
- [ ] Split `ZCL_ABAPGIT_OBJECT_SMTG` into two classes (https://github.com/abapGit/abapGit/pull/7682)
- [ ] Update [docs](https://docs.abapgit.org/development-guide/serializers/serializers.html)

@larshp @albertmink

larshp:
hmm, but `ZCL_ABAPGIT_OBJECT_xxxx` and `ZCL_ABAPGIT_OBJECT_AFF_xxxx` will use the same code to deserialize, where should that code reside? I havent looked at SMTG yet

mbtools:
That duplication in DOMA and INTF is just temporary. Both should have a class derived from
COMMON_AFF which points to the SAP-delivered implementation (like the other AFF objects).

[...]

PR #7861, opened 2026-09-07
DTDC: Guard deserialize against unbound io_xml (#7860)

### LINKED ISSUE #7860: Unhandled CX_SY_REF_IS_INITIAL dump in ZCL_ABAPGIT_OBJECT_DTDC~DESERIALIZE when repo has JSON metadata for a DTDC object
### Summary
Deserializing (pulling) a repository containing a `DTDC` (Dynamic Cache) object whose metadata is stored as JSON causes an unhandled ABAP short dump (`OBJECTS_OBJREF_NOT_ASSIGNED_NO` / `CX_SY_REF_IS_INITIAL`) instead of a normal abapGit error/log entry.

### Environment
- abapGit standalone, backend version `1.134.0` (`ZIF_ABAPGIT_VERSION=>C_ABAP_VERSION`)
- SAP_BASIS 757 SP03, kernel 789, on-prem HANA

### Steps to reproduce
1. Have a repo containing a `DTDC` object whose metadata is JSON (i.e. `ZCL_ABAPGIT_OBJECTS_FILES=>IS_JSON_METADATA` returns `abap_true` for it — a `*.dtdc.json` file present).
2. Deserialize/pull that repo (step `DDIC`).

### Actual result
Short dump, exception `CX_SY_REF_IS_INITIAL`, uncaught in `ZIF_ABAPGIT_OBJECT~DESERIALIZE` (class `ZCL_ABAPGIT_OBJECT_DTDC`).

Failing statement:
```abap
    ASSIGN mr_dynamic_cache->* TO <ls_dynamic_cache>.
    ASSERT sy-subrc = 0.

>>> io_xml->read(
      EXPORTING
        iv_name = 'DTDC'
      CHANGING
        cg_data = <ls_dynamic_cache> ).
```
`io_xml` is `{O:initial}` (unbound) at the point of the call.

Call stack (top frames):
```
#0 ZCL_ABAPGIT_OBJECT_DTDC=>ZIF_ABAPGIT_OBJECT~DESERIALIZE
#1 ZCL_ABAPGIT_OBJECTS=>DESERIALIZE_STEP (step = DDIC)
```

### Root cause
`DESERIALIZE_STEP` clears `io_xml` whenever `IS_JSON_METADATA()` is true for the object. `ZCL_ABAPGIT_OBJECT_DTDC~DESERIALIZE` never checks `io_xml IS BOUND` before calling `io_xml->read(...)`, so it dereferences null instead of raising a controlled `zcx_abapgit_exception`. Related to the open XML/JSON dual-support gap in #7768.

### Suggested fix
Guard the `io_xml->read(...)` call in `ZCL_ABAPGIT_OBJECT_DTDC` with an `IS BOUND` check and raise `zcx_abapgit_exception` (e.g. "DTDC: JSON metadata not supported for deserialize") instead of letting the runtime terminate the session.

ThePlenkov:
## Summary
- `ZCL_ABAPGIT_OBJECT_DTDC~DESERIALIZE` called `io_xml->read(...)` without checking that `io_xml` was bound. When a DTDC object's metadata is stored as JSON, `DESERIALIZE_STEP` clears `io_xml`, causing an unhandled `CX_SY_REF_IS_INITIAL` short dump (`OBJECTS_OBJREF_NOT_ASSIGNED_NO`) instead of a normal abapGit error/log entry.
- Added an `IS BOUND` guard that raises a controlled `zcx_abapgit_exception` ("DTDC: JSON metadata not supported for deserialize") so the failure is reported via the log instead of crashing the session.

Fixes #7860. Related to the XML/JSON dual-support gap in #7768.

#### Test plan
- [x] `abaplint` passes (0 issues, 1418 files analyzed)
- [ ] Manual: deserialize a repo containing a `DTDC` object with `*.dtdc.json` metadata and confirm a controlled error is logged instead of a short dump

Generated with [Devin](https://devin.ai)

### DIFF
--- a/src/objects/zcl_abapgit_object_dtdc.clas.abap
+++ b/src/objects/zcl_abapgit_object_dtdc.clas.abap
@@ -316,6 +316,10 @@ CLASS zcl_abapgit_object_dtdc IMPLEMENTATION.
     ASSIGN mr_dynamic_cache->* TO <ls_dynamic_cache>.
     ASSERT sy-subrc = 0.
 
+    IF io_xml IS NOT BOUND.
+      zcx_abapgit_exception=>raise( |DTDC: JSON metadata not supported for deserialize| ).
+    ENDIF.
+
     io_xml->read(
       EXPORTING
         iv_name = 'DTDC'


larshp:
will https://github.com/abapGit/abapGit/pull/7809/commits/16e2098bf4663f3ba6e99462ff54cee77ddf6e05 also fix this issue?

ThePlenkov:
@larshp Good question. I checked — `DTDC` is **not** registered in `zcl_abapgit_aff_registry` (only `DTSC`, `DTIX`, plus `DOMA`/`INTF` experimental are listed in `initialize_registry_table`).

So yes, PR #7809's new `read_metadata` would also prevent this specific dump: for a DTDC object with JSON metadata, `is_json_metadata()` is true and `is_supported_object_type('DTDC')` returns `abap_false`, so it raises `"Object is serialized in AFF format. Enable experimental feature..."` **before** `deserialize` is ever called — the unbound `io_xml` dereference is never reached.

Two caveats that still make this guard worthwhile:
1. The #7809 message is a bit misleading for DTDC, since DTDC isn't an AFF type at all (it's not in the registry) — there's nothing to "enable". The message here names the actual object/limitation.
2. If DTDC were ever added to the AFF registry in the future, `read_metadata` would return with cleared `eo_xml` and the dump would recur, since `deserialize` itself still has no `IS BOUND` check. The guard in this PR is defensive at the point of use.

Happy to close this in favor of #7809 if you'd prefer to handle it centrally there — just let me know.

mbtools:
sorry, how did you end up with JSON for DTDC?

ThePlenkov:
> sorry, how did you end up with JSON for DTDC?

i like this question @mbtools 

actually real reason is - https://github.com/abapify/abapgit-examples/tree/main/src/cds-rap my own repo which i generated with rap examples. And i guess source of this was AFF. Apart of that - i don't mind if some AFF formats are not supported yet. What I want to say is that short dump is not the best thing to ctach - i'd rather raised not supported exception and skipped some files..

mbtools:
So the repo was not created by abapGit... Anyway, I agree that it should not crash and provide an error message. 

The issue can happens with any object. The fix should be implemented by the framework (in `zcl_abapgit_objects`) and not in all of the individual object types. 

Let's close this one and make sure that #7809 `read_metadata` returns the right message.

ThePlenkov:
@mbtools yes I agree - it should be a general fix

PR #7865, opened 2026-09-08
Fix WebGUI keyboard actions and guard palette submissions

christianguenter2:
Fix WebGUI dialog hotkeys, picker keyboard actions, and palette submission navigation.

The recent WebGUI improvements fixed form submission, but some keyboard paths did not recognize the updated controls. Command-palette submissions also bypassed the navigation guard added in #7830.

  This PR fixes three related issues:

  - Dialog hotkeys: Preserve data-sapevent on WebGUI dialog commands so hotkey lookup can find the action and annotate its tooltip.
  - Picker buttons: Include WebGUI side-action buttons, such as package and branch pickers, in the command palette. Selecting their link hint now activates the picker instead of only focusing it.
  - Submission navigation: Route palette submissions through a shared navigation guard. Guard link-hint activation as well, including submit inputs that inherit their action from the form. This prevents a submission-induced popstate
    from incorrectly triggering Back.

  Adds eight JavaScript regression tests and two ABAP rendering tests, with JavaScript tests included in PR CI.

Successfully Tested with latest versions of:
- [x] SAP GUI for HTML
- [x] SAP GUI for Windows
- [x] SAP GUI for Java

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -16,6 +16,7 @@ jobs:
     - run: |
         npm install
         npm run eslint
+        npm run test:ui
         npm run merge
         npm run merge.ci
     - name: Upload build artifact
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "license": "MIT",
   "private": true,
   "scripts": {
-    "test": "npm run eslint && npm run abaplint",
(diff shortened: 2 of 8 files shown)


mbtools (approved):
lgtm 👍

PR #7866, opened 2026-09-09
Expand UI tests and fix command palette filtering

christianguenter2:
Adds JavaScript tests for form submission, staging, and command palette behavior using the existing Node/VM test setup.

  The palette tests exposed two bugs, fixed in this PR:

  - Pressing Enter after filtering to zero matches executed the previously selected, hidden command. The selection is now cleared.
  - Reopening the palette cleared the search field but retained the old filter. Both now reset together.

  Coverage includes WebGUI routing and parameter encoding, repeated form submissions, staging selection/filter precedence, keyboard navigation, and fuzzy matching. A shared VM loader keeps the test setup reusable without adding
  dependencies.

Adds test/ui/README.md documenting how to run and extend the tests, covered behaviors, regression expectations, and limitations of the DOM fixtures.


mbtools (approved):
👍

PR #7868, opened 2026-09-09
Fix HTML truncation with non-ASCII content in SAPGUI for Java

### LINKED ISSUE #7867: SAPGUI for Java: HTML-assets are truncated
When using latest SAPGUI for Java (8.10 rev 13), HTML-assets are truncated when there are umlauts. If there are too many - 17 seems to be the magic threshold - JS breaks.

Truncation with <=16-Umlauts - no error because only html end tags are removed
Image

Truncation with >17-Umlauts - JS Init error
Image

Image

This is only a problem in SAPGUI for Java, other GUIs are not affected.

While we can fix this in abapGit, it seems something for SAP to fix, too. I try to draw some attention on that.

christianguenter2:
HTML pages containing non-ASCII characters can be truncated in SAP GUI for Java because the transferred UTF-8 byte length exceeds the character-based size. This can cut off the final JavaScript block and prevent page initialization.

For example, a document containing 961,441 characters encoded as 961,542 UTF-8 bytes returned only 961,441 bytes, losing the final 101 bytes.

This change loads HTML as UTF-8 binary data for SAP GUI for  Java, using the encoded byte length. Windows and WebGUI retain the existing character-based loading path. WebGUI requires that path for SAP event-link conversion; applying binary loading there broke navigation.

fixes #7867

### DIFF
--- a/src/ui/core/zcl_abapgit_gui.clas.abap
+++ b/src/ui/core/zcl_abapgit_gui.clas.abap
@@ -202,10 +202,19 @@ CLASS zcl_abapgit_gui IMPLEMENTATION.
 
   METHOD cache_html.
 
-    rv_url = zif_abapgit_gui_services~cache_asset(
-      iv_text    = iv_text
-      iv_type    = 'text'
-      iv_subtype = 'html' ).
+    IF zcl_abapgit_ui_factory=>get_frontend_services( )->is_sapgui_for_java( ) = abap_true.
+      "Java GUI needs the UTF-8 payload length in bytes for non-ASCII HTML.
+      rv_url = zif_abapgit_gui_services~cache_asset(
+        iv_xdata   = zcl_abapgit_convert=>string_to_xstring_utf8( iv_text )
+        iv_type    = 'text'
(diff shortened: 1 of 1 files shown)


mbtools:
The size comes from the conversions method. Did you check that the bug isn't there?

```abap
zcl_abapgit_gui
METHOD zif_abapgit_gui_services~cache_asset.
...
zcl_abapgit_convert=>string_to_tab(
         EXPORTING
           iv_str  = iv_text
         IMPORTING
           ev_size = lv_size
           et_tab  = lt_html ).
```

christianguenter2:
Yes. The character/byte mismatch originates in its returned size, but that doesn’t necessarily make string_to_tab itself wrong.

In string_to_tab `ev_size = strlen( iv_str )` counts characters, matching the character table it produces. The mismatch occurs when that size is passed to Java GUI’s loader for the UTF-8 payload. Changing only ev_size to a byte count would mix two different representations.

mbtools (approved):
Thanks 🙏

PR #7869, opened 2026-09-10
Fix palette title escaping and repository selection guards

christianguenter2:
Command palette filtering now escapes title text before adding match highlights, so characters such as <, > and & remain literal text.

Repository navigation now safely handles empty lists, out-of-range indexes, Enter without a selection, and restoring a saved selection for a removed repository.


mbtools (approved):
👍

PR #7871, opened 2026-09-12
Keep staged files when returning to the stage page in WebGUI

christianguenter2:
The stage table stores its state in sessionStorage on the way out and restores it on the way back in, but the handler was only registered for beforeunload.

SAP GUI for HTML renders abapGit in an ITS-managed iframe and replaces that iframe on every navigation instead of navigating it. A frame torn down that way gets pagehide, never beforeunload, so nothing was ever written and returning from the commit page reset every row to "?".

Register the handler for pagehide as well. The embedded controls of the desktop GUIs predate pagehide, so beforeunload stays; storing twice is harmless because it writes the same state under the same key.

Verified on SAP GUI for HTML: before the change the seed key was absent from sessionStorage after leaving the page, and probes showed pagehide and unload firing while beforeunload did not.


mbtools (approved):
👍

PR #7872, opened 2026-09-12
Prevent diff filter clicks from triggering browser Back

christianguenter2:
Diff filter links use href="#" but did not cancel the default navigation.
The resulting popstate was interpreted as browser Back and triggered go_back, returning the user to the previous page.

This regression was introduced on 23 July 2026 by
a51687c22ebadc0101d1c9beb6d78c789c1c43ad
(Enable browser back navigation in SAPGUI + WebGui, #7782). The missing cancellation predates that commit, but its browser-back trap turned the fragment navigation into an unintended SAP back action.

Cancel default navigation for checklist filter clicks. Add regression coverage for link and icon clicks, legacy srcElement events, Only my changes, and preservation of genuine Back navigation.


mbtools (approved):
👍

PR #7877, opened 2026-09-14
Fix UI filtering, patch selection, palette clicks and saved state

christianguenter2:
- repo overview: tolerate absent/stale localStorage, and look up the selected row by dataset key instead of an interpolated selector
- stage: fall back to a literal search on an incomplete regex, and escape marked text before writing it back into innerHTML
- stage: ignore saved statuses that are not one of A/R/I/?
- diff: derive filter visibility from the full set of unchecked options so categories intersect instead of clobbering each other
- diff: match jump-list entries and jump targets on the whole path
- patch: bound the checkbox id prefix with "_" so section 1 no longer selects the lines of sections 10 and 11
- palette: walk up to the LI so clicks on <mark> reach the command
- scroll position: never let a failing sessionStorage write abort the refresh action it wraps


mbtools (approved):
it didn't break anything for me. lgtm 👍

PR #7879, opened 2026-09-15
Fix Slack badge link in README

larshp:
Updated Slack badge link to new inviter URL.


larshp:
we had another join the slack channel, so guess this PR works

mbtools (approved):
Nice

PR #7881, opened 2026-09-16
Guard keyboard shortcuts and diff copying; expand JavaScript tests

christianguenter2:
Plain hotkeys now ignore Ctrl/Alt/Meta combinations, preventing shortcuts such as Ctrl+S from triggering Stage. Typing in editable controls no longer changes link-hint yank mode.

Diff copying now ignores table-background clicks, empty selections, and selections outside the active diff table. Clipboard failures leave native copying available.

Adds regression tests for these cases, source-viewer loading and error handling, and enables coverage reporting for the shipped JavaScript.

Current coverage of `common.js`
image


mbtools (approved):
👍

mbtools:
There's an issue with copying a selection in the split diff view (unified works fine) but I think that bug was there before already 🤷 

When selecting multiple lines, only the first is copied to clipboard:

image

christianguenter2:
Hm, works for me in all GUIs as it did before 🤔

mbtools:
copy-to-clipboard issue is unrelated to this change. merging

PR #7883, opened 2026-09-21
Respect Exclude Paths when reading remote .abapgit.xml

pokrakam:
`zcl_abapgit_repo->find_remote_dot_abapgit` reads `/.abapgit.xml` from the raw
`mt_remote` table, so it ignores `local_settings-exclude_remote_paths`. Every
other consumer of the remote file list goes through `get_files_remote( )`,
which applies the exclusions - this method is the one inconsistency.

This change makes it read the filtered `get_files_remote( )` instead.

Behavior is unchanged for repos that don't exclude `/.abapgit.xml`. For repos
that do, the locally persisted .abapgit.xml (starting folder, folder logic) is
no longer overwritten from the remote on every pull. That makes it possible to
bind several repos on different systems to different subfolders of one
monorepo, each with its own locally configured starting folder, while the root
.abapgit.xml keeps serving the default binding.

Covered by a new unit test (`dot_ignored_when_excluded`) alongside the two
existing `find_remote_dot_abapgit` tests.


mbtools (approved):
Subtle difference between mt_remote and lt_remote 😉 . Thanks for the fix!

mbtools:
apack is pretty much dead but we should apply the same fix to `find_remote_dot_apack` 🙏

pokrakam:
Didn't even notice. Good call, done.

mbtools (approved):
👍

PR #7884, opened 2026-09-22
Fix link hint cancellation and disabled controls; guard menu shortcuts

christianguenter2:
Cancelling link hints now clears copy mode, preventing the next hint selection from unexpectedly copying instead of activating a link.

Link hints skip disabled controls, including controls inside disabled fieldsets, and recheck their state before activation. Menu keyboard navigation now preserves Ctrl/Alt/Meta combinations.


mbtools (approved):
👍

PR #7885, opened 2026-09-22
Fix repository stats Back navigation in WebGUI

christianguenter2:
The Back button on Repository Stats submits a form without a main action. WebGUI therefore doesn’t initialize its SAP event routing, causing navigation to fail.

Render Back as a navigation link instead, since the read-only Stats page has no form data to submit.


mbtools (approved):
👍

PR #7886, opened 2026-09-22
Fix split diff block copying for incoming remote changes

christianguenter2:
Copying multiple lines from a split diff with incoming remote changes copies only the first selected line.

The renderer swaps the old/new columns for incoming changes, but the copy handler assumes fixed column positions. Subsequent lines are read from the opposite side and filtered out.

Use the clicked cell’s actual column index, which also handles the optional patch column.

cc @mbtools , long time bug.
Fixes https://github.com/abapGit/abapGit/pull/7881#issuecomment-5767583960


mbtools:
It's perfect with Edge control.  Thank you 🙏 🙏 🙏 

On IE, the copy result includes the other columns:

```txt
    FIELD-SYMBOLS: <ls_remote> LIKE LINE OF mt_remote. 276 ~     DATA lt_remote TYPE zif_abapgit_git_definitions=>ty_files_tt. 
277   277   
278 ~     get_files_remote( ). 278 ~     FIELD-SYMBOLS: <ls_remote> LIKE LINE OF lt_remote. 
279   279   
280 ~     READ TABLE mt_remote ASSIGNING <ls_remote> 
```

image

christianguenter2:
@marcfbe regarding IE, can you test #7887?

mbtools:
The combo of https://github.com/abapGit/abapGit/pull/7886 and #7887 works great! Thx 👍

I will let you do the merge so they don't overwrite things.

mbtools (approved):

PR #7887, opened 2026-09-22
Fix diff column copying in the IE browser control

christianguenter2:
IE's Node.contains() accepts only elements, so the selection range check added in #7881 rejected every text-node range and fell back to native copying, which includes all table columns.


mbtools:
The combo of #7886 and this one works great! Thx 👍

mbtools (approved):

